PR #311, opened 2015-01-21
Add GlideModule for lazily initialization.

### LINKED ISSUE #299: Reduce Glide setup time
Currently it seems like it can take up to 30ms. We should consider ways to reduce that time or at least allow ModelLoaders to be registered lazily. Branching from @TWiStErRob in #295:

```
Normal startup (5 samples):
TIMING﹕ Glide.get: 15.87, Glide.register: 0.09, new Client: 0.70, new Factory: 0.12
TIMING﹕ Glide.get: 16.60, Glide.register: 0.09, new Client: 0.76, new Factory: 0.15
TIMING﹕ Glide.get: 20.78, Glide.register: 0.12, new Client: 1.10, new Factory: 0.21
TIMING﹕ Glide.get: 23.38, Glide.register: 0.12, new Client: 1.07, new Factory: 0.21
TIMING﹕ Glide.get: 31.86, Glide.register: 0.18, new Client: 5.16, new Factory: 0.34

with Start Method Tracing:
TIMING﹕ Glide.get: 64.97, Glide.register: 0.82, new Client: 2.47, new Factory: 0.40
```

sjudd:
Fixes #299.

### DIFF
--- a/integration/okhttp/src/main/AndroidManifest.xml
+++ b/integration/okhttp/src/main/AndroidManifest.xml
@@ -1,5 +1,8 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.bumptech.glide.integration.okhttp">
-    <application />
-
+    <application>
+        <meta-data
+            android:name="com.bumptech.glide.integration.okhttp.OkHttpGlideModule"
+            android:value="GlideModule" />
+    </application>
 </manifest>
--- a/integration/okhttp/src/main/java/com/bumptech/glide/integration/okhttp/OkHttpGlideModule.java
+++ b/integration/okhttp/src/main/java/com/bumptech/glide/integration/okhttp/OkHttpGlideModule.java
@@ -1,0 +1,27 @@
+package com.bumptech.glide.integration.okhttp;
+
+import android.content.Context;
+
+import com.bumptech.glide.Glide;
+import com.bumptech.glide.load.model.GlideUrl;
+import com.bumptech.glide.module.GlideModule;
+
+import java.io.InputStream;
+
+/**
+ * A {@link com.bumptech.glide.module.GlideModule} implementation to replace Glide's default
+ * {@link java.net.HttpURLConnection} based {@link com.bumptech.glide.load.model.ModelLoader} with an OkHttp based
+ * {@link com.bumptech.glide.load.model.ModelLoader}.
+ *
+ * <p>
+ *     If you're using gradle, you can include this module simply by depending on the aar, the module will be merged
+ *     in by manifest merger. For other build systems or for more more information, see
+ *     {@link com.bumptech.glide.module.GlideModule}.
+ * </p>
+ */
+public class OkHttpGlideModule implements GlideModule {
+    @Override
+    public void initialize(Context context, Glide glide) {
+        glide.register(GlideUrl.class, InputStream.class, new OkHttpUrlLoader.Factory());
+    }
+}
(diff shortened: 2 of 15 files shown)


TWiStErRob:
Concern: metadata without a value looks strange.
Wouldn't it be better if you did `name="com.bumptech.glide.module.flickr" value="full.class.name"`, less parsing (no need for substring, but you can keep the startswith check) and better readability, compliance.
The doc for metadata.name says to prefix your name with package: http://developer.android.com/guide/topics/manifest/meta-data-element.html

sjudd:
@TWiStErRob The reason why the fully qualified class name has to be in the key is that there can be only one metadata tag for a given key. If we use anything other than the fully qualified class name, it's possible a user would accidentally try to add two different implementations with the same key (two different classes in the same package for example). If they did so, we'd have no way to detect the error because the manifest merger/parser would silently merge the two.

I agree the syntax is funky, but unless there's some way around the fully qualified class name issue, I'm not sure we have anything useful to put in the value.

TWiStErRob:
Note that I still left flicker in the name, which is just a random name. That name after that is not used anywhere just helps readability, users can organize modules nicer. E.g. for okhttp it would be different, the point is to start it with `...glidemodule.`.
I read the design doc lately and as I remember the merger will conflict if two modules have the same name after glidemodule, but the values are different.
The best would be a multivalued metadata which is not supported sadly.

TWiStErRob:
Also in case of placeholders (e.g. I hate duplicating the package name so I have a placeholder for it like `${packageName}.subpackage.Class`) it's cleaner to start with that instead of having it in the middle.

sjudd:
We've actually tried something like this with user specifiable names, and found that people would still occasionally choose the same name. Using the class name just completely eliminates the possibility, unless they add the same class twice, in which case ignoring the second copy should be fine.

We could drop ``glidemodule:`` altogether, but then we wouldn't be able to assert as much (malformed class name and failing to implement GlideModule would have to be ok). 

We could append :glidemodule instead if you think that's cleaner?

sjudd:
Another option would be to leave the fully qualified class name as the key, but set glidemodule as the value. Maybe that's cleaner? It would at least avoid the string parsing.

TWiStErRob:
Hmm that's interesting... just realized: all of these are hacks, so I guess the last one is the best of them you're taking onto consideration.

Rant follows :) Your response to people occasionally choosing the same names is the approach I saw at my university in the classes following my year, a lot of students were failing the most basic programming tests so they made them easier every semester. My view/approach is always: use stuff as designed, so keys as keys and value as values. End of rant.

sjudd:
I hear you. It's worth keeping in mind though that it's not even necessarily possible for people to do the right thing. If you have a number of independent library projects, each owned by separate teams, and multiple of those teams define their own modules, we don't want to force people to go inspect all of the other team's modules before they come up with a module name (or keep a list of all used names somewhere...). 

I agree it's the best of bad options, I'll move module into the value and remove the prefix.

sjudd:
Made the change, PTAL, thanks for the input!

sjudd:
I added documentation about manifest merging, hopefully I got it right (having not done this in gradle before...)

kxs109:
I have a question, I want to load some imgs include gif from the net and list in the listview, it is ok and amazing to show, i like the performance,  but it's bad when run the app the first time after install it, it will take a long time to show these imgs. i don't know why...

TWiStErRob:
Please open a new question to follow up, also see https://github.com/bumptech/glide/blob/master/CONTRIBUTING.md#filing-issues

PR #318, opened 2015-01-28
Enable to build even if using Java 8

rejasupotaro:
I tried to run test. But it failed.

```java
$ ./gradlew test
...
:library:checkstyleTestDebug UP-TO-DATE
:library:findbugs FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':library:findbugs'.
> The version of FindBugs (2.0.3) inferred from FindBugs classpath is too low to work with currently used Java version (1.8). Please use higher version of FindBugs. Inspected FindBugs classpath: [findbugs-2.0.3.jar, jcip-annotations-1.0.jar, jsr305-2.0.1.jar, bcel-2.0.1.jar, jFormatString-2.0.1.jar, dom4j-1.6.1.jar, asm-3.3.jar, asm-commons-3.3.jar, commons-lang-2.6.jar, AppleJavaExtensions-1.4.jar, jaxen-1.1.6.jar, xml-apis-1.0.b2.jar, asm-tree-3.3.jar]

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED

Total time: 30.706 secs
```

It seems that this project cannot build with Java 8

```shell
$ java -version
java version "1.8.0_20"
Java(TM) SE Runtime Environment (build 1.8.0_20-b26)
Java HotSpot(TM) 64-Bit Server VM (build 25.20-b23, mixed mode)
```

because of Find Bugs v2.0.3 doesn't support Java 8.

FindBugs™ - Find Bugs in Java Programs http://findbugs.sourceforge.net/

So I updated Find Bugs and related build settings.

Related issue: https://github.com/sjudd/DiskLruCache/pull/1


rejasupotaro:
@sjudd Could you check whether if this branch can build on your PC? Maybe you should re-import this project.

sjudd:
I'll check.. We want to continue to build the source with java 6 for the time being, it might be ok to build tests with java 7 though.

Either way, please revert the .iml/.idea changes, we'd like to keep the ide setup in the repo.

sjudd:
Argh yeah 2.0.3 doesn't support java 8, but 3.x doesn't support java 6, so we can't accept this pull request yet. If you want to build, you'll have to install java 6 (7 may also work?). Glide 4.0 will build with java 7 which should alleviate this problem.

rejasupotaro:
Oh, I see... Okay. I close this pull-req. Thank you.

TWiStErRob:
@sjudd Remember that we are using gradle, a scripting build tool. It is possible to set the `findbugs.toolVersion` in library/build.gradle like so (untested, hacked together on phone):
```
findbugs {
    if (org.apache.commons.lang.SystemUtils.isJavaVersionAtLeast(1.7f)) {
        toolVersion = "3.0.0"
    } else {
        toolVersion = "2.0.3"
    }
}
```
See http://stackoverflow.com/a/23706899/253468, Apache commons should be available since I kind of remember gradle depending on it, if not, use the method of the OP (java.version.startsWith).

sjudd:
@TWiStErRob that's a reasonable suggestion, particularly since the next major release will hopefully be Glide 4.0 with support for java 7. As long as the checks don't change radically between the two versions I'd be ok with this if it makes it more convenient for people. 

@rejasupotaro Do you want to try again with that change?

sjudd:
Glide 4.0a will build with java 7 and 8. I added corresponding build flavors to travis so this will be more reliable in the future: https://github.com/bumptech/glide/commit/d34b710a70d3d2591d443d89277a8275779f1641

PR #336, opened 2015-02-08
Add a Size + Config Strategy.

### LINKED ISSUE #335: BitmapPool hit rate decreased in Glide 3.5
To work around #301, we banned Bitmaps with configs other than ARGB_8888 from being placed into the BitmapPool and defaulted to always requesting ARGB_8888 Bitmaps.

Unfortunately the framework doesn't always obey inPreferredConfig. GIFs in particular almost always produce Bitmaps with null (hidden) configs. 

Since those Bitmaps are banned from the pool, apps displaying a mix of single frame GIFs and other content will end up with a net drain on the BitmapPool, causing jank. 

From some limited testing it appears safe to re-use ARGB_8888 Bitmaps to decode single frame GIFs, and to re-use Bitmaps with null configs to decode most image types with inPreferredConfig == ARGB_8888.

sjudd:
Fixes #335.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/load/engine/bitmap_recycle/PrettyPrintTreeMap.java
+++ b/library/src/main/java/com/bumptech/glide/load/engine/bitmap_recycle/PrettyPrintTreeMap.java
@@ -1,0 +1,21 @@
+package com.bumptech.glide.load.engine.bitmap_recycle;
+
+import java.util.TreeMap;
+
+class PrettyPrintTreeMap<K, V> extends TreeMap<K, V> {
+    @Override
+    public String toString() {
+        StringBuilder sb = new StringBuilder();
+        sb.append("( ");
+        for (Entry<K, V> entry : entrySet()) {
+            sb.append('{').append(entry.getKey()).append(':').append(entry.getValue()).append("}, ");
+        }
+        final String result;
+        if (!isEmpty()) {
+            result = sb.substring(0, sb.length() - 2);
+        } else {
+            result = sb.toString();
+        }
+        return result + " )";
+    }
+}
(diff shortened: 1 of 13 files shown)


library/src/main/java/com/bumptech/glide/load/engine/bitmap_recycle/PrettyPrintTreeMap.java
```diff
@@ -0,0 +1,21 @@
...
+        for (Entry<K, V> entry : entrySet()) {
+            sb.append('{').append(entry.getKey()).append(':').append(entry.getValue()).append("}, ");
+        }
+        final String result;
+        if (!isEmpty()) {
+            result = sb.substring(0, sb.length() - 2);
+        } else {
+            result = sb.toString();
+        }
+        return result + " )";
```
TWiStErRob: Using stringbuilder and then a + op is pretentious ;)

library/src/main/java/com/bumptech/glide/load/engine/bitmap_recycle/SizeConfigStrategy.java
```diff
@@ -0,0 +1,234 @@
...
+    private static final Bitmap.Config[] ARGB_4444_IN_CONFIGS = new Bitmap.Config[] {
+            Bitmap.Config.ARGB_4444
+    };
+    private static final Bitmap.Config[] ALPHA_8_IN_CONFIGS = new Bitmap.Config[] {
+            Bitmap.Config.ALPHA_8
+    };
+
+    private final KeyPool keyPool = new KeyPool();
+    private final GroupedLinkedMap<Key, Bitmap> groupedMap = new GroupedLinkedMap<Key, Bitmap>();
+    private final Map<Bitmap.Config, TreeMap<Integer, Integer>> sortedSizes =
```
TWiStErRob: I'd use SortedMap (iface) instead of class in this class.

library/src/main/java/com/bumptech/glide/load/engine/bitmap_recycle/PrettyPrintTreeMap.java
sjudd: Hah, done.

library/src/main/java/com/bumptech/glide/load/engine/bitmap_recycle/SizeConfigStrategy.java
sjudd: I had to use NavigableMap since I use ceilingKey, but done.

sjudd:
Rebased in here: https://github.com/bumptech/glide/commit/431ccaf04b5c3db3f87e2ac2c1ca151f5d63744f

Issue #202, opened 2014-10-19
GifDrawable may try to draw with recycled bitmaps


sjudd:
This may happen if decoding a frame fails but we recycle the previous frame anyway.

```
E/AndroidRuntime( 3051): FATAL EXCEPTION: main
E/AndroidRuntime( 3051): Process: com.bumptech.glide.samples.giphy, PID: 3051
E/AndroidRuntime( 3051): java.lang.RuntimeException: Canvas: trying to use a recycled bitmap android.graphics.Bitmap@528c492c
E/AndroidRuntime( 3051):    at android.graphics.Canvas.throwIfCannotDraw(Canvas.java:1084)
E/AndroidRuntime( 3051):    at android.view.GLES20Canvas.drawBitmap(GLES20Canvas.java:811)
E/AndroidRuntime( 3051):    at com.bumptech.glide.load.resource.gif.GifDrawable.draw(GifDrawable.java:185)
E/AndroidRuntime( 3051):    at android.widget.ImageView.onDraw(ImageView.java:1037)
E/AndroidRuntime( 3051):    at android.view.View.draw(View.java:14465)
E/AndroidRuntime( 3051):    at android.view.View.getDisplayList(View.java:13362)
E/AndroidRuntime( 3051):    at android.view.View.getDisplayList(View.java:13404)
E/AndroidRuntime( 3051):    at android.view.View.draw(View.java:14182)
E/AndroidRuntime( 3051):    at android.view.ViewGroup.drawChild(ViewGroup.java:3103)
E/AndroidRuntime( 3051):    at android.view.ViewGroup.dispatchDraw(ViewGroup.java:2940)
E/AndroidRuntime( 3051):    at android.view.View.getDisplayList(View.java:13357)
E/AndroidRuntime( 3051):    at android.view.View.getDisplayList(View.java:13404)
E/AndroidRuntime( 3051):    at android.view.View.draw(View.java:14182)
E/AndroidRuntime( 3051):    at android.view.ViewGroup.drawChild(ViewGroup.java:3103)
E/AndroidRuntime( 3051):    at android.view.ViewGroup.dispatchDraw(ViewGroup.java:2940)
E/AndroidRuntime( 3051):    at android.view.View.getDisplayList(View.java:13357)
E/AndroidRuntime( 3051):    at android.view.View.getDisplayList(View.java:13404)
E/AndroidRuntime( 3051):    at android.view.View.draw(View.java:14182)
[...]

PR #341, opened 2015-02-14
Before draw bitmap, check bitmap was recycled.

ZeroBrain:
sometimes glide occured exception as java.lang.IllegalArgumentException

exception message

```
GlideBitmapDrawable.java line 101 in GlideBitmapDrawable.draw()
Cannot draw recycled bitmaps
```

so add checking bitmap was recycled.

I reviewed GifDrawable.java. it was checked recycle state.
so add same one to GlideBitmapDrawable.java

### DIFF
--- a/library/src/main/java/com/bumptech/glide/load/resource/bitmap/GlideBitmapDrawable.java
+++ b/library/src/main/java/com/bumptech/glide/load/resource/bitmap/GlideBitmapDrawable.java
@@ -98,7 +98,9 @@ public class GlideBitmapDrawable extends GlideDrawable {
             Gravity.apply(BitmapState.GRAVITY, width, height, getBounds(), destRect);
             applyGravity = false;
         }
-        canvas.drawBitmap(state.bitmap, null, destRect, state.paint);
+        if (!state.bitmap.isRecycled()) {
+            canvas.drawBitmap(state.bitmap, null, destRect, state.paint);
+        }
     }
 
     @Override


TWiStErRob:
Hmm, isn't this just hiding another bug? (covering the symptom)

I thought `draw` is called on the UI thread, and so does `recycle` so they should be ordered and the target be cleared as soon as (or even before) recycle happens, i.e. draw cannot be called with an invalid bitmap.

What makes this invariant not true?

ZeroBrain:
maybe it's hiding another bug.
but sometimes glide' lrubitmappool remove drawing bitmap.
especially low memory device was occurred frequency.
my crash was occured almost by Glide' `Cannot draw recycled bitmaps`.

GifDrawable is checked recycled or not.
but GlideBitmapDrawable isn't checked.

I thought it's more critical than hiding bug.

sjudd:
Yes this is hiding another bug. We should not be recycling resources that are still in use. The lifecycle is roughly:

1. Decode bitmap in background
2. Pass Bitmap to Target on main thread and draw it.
3. Clear Target (or view) by either calling Glide.clear() or starting a new load into the same Target (or view). 
4. Return Bitmap to pool (and go back to #1).

If you are starting multiple loads into the same view or otherwise retaining a Bitmap after starting a new load or clearing an old load, you will see this exception. If you want to load multiple images, you need to make sure they're either loaded into separate Targets (or views), or that you avoiding using any loaded resources after a new load is started or the Target is cleared.

Probably worth [filing a new issue][1] with whatever issue you're seeing so we can look in to it.

[1]: https://github.com/bumptech/glide/issues/new?body=**Glide%20Version/Integration%20library%20(if%20any)**:%0A**Device/Android%20Version**:%0A**Issue%20details/Repro%20steps**:%0A%0A**Glide%20load%20line**:%0A%60%60%60%0AGlide.with(context)...%3B%0A%60%60%60%0A%0A**Stack%20trace**:%0A%60%60%60%0Apaste%20stack%20trace%20here%0A%60%60%60

ZeroBrain:
@sjudd 
ah~ I understand. I guess multiple loading point in my app.

It need to add wiki or some info page.
someone may suffer about it, too

TWiStErRob:
@sjudd do you know why the gif check is there? Just to see the whole picture here.

@ZeroBrain you can click the link at the end of Sam's comment and describe your circumstances, so we can help figure out if it's a bug in Glide or you're using it incorrectly.

sjudd:
@TWiStErRob Git blame says: https://github.com/bumptech/glide/issues/202. GifDrawable switches back and forth between two targets on each frame but doesn't really handle when a single frame fails to load.

A better fix would be to re-use the same Target twice in a row if a single frame fails to load, so the GIF would drop a frame rather than briefly stop drawing at all.

sjudd:
@ZeroBrain I added an article to the wiki, hope it helps: https://github.com/bumptech/glide/wiki/Resource-re-use-in-Glide

mufumbo:
How you guys are able to fix this? I have a super simple ListView and call glide for each reused row. 

From the wiki article, it seems that I have to be responsible to not call glide again if the user is scrolling super fast on the ListView and a row is being overwritten by another load?

mufumbo:
Also, do you guys know why this exception doesn't happen on lollipop? It happens always on 4.1.

PR #352, opened 2015-02-24
New Glide class static methods that allow disk cache & all cache types cleaning.

nemphys:


nemphys:
PR updated, please check

library/src/main/java/com/bumptech/glide/load/engine/Engine.java
```diff
@@ -293,6 +293,10 @@ public void onResourceReleased(Key cacheKey, EngineResource resource) {
         }
     }
 
+    public void clearDiskCache() {
+        diskCacheProvider.getDiskCache().clear();
```
sjudd: Should this happen asynchronously?

library/src/main/java/com/bumptech/glide/load/engine/cache/DiskLruCacheWrapper.java
```diff
@@ -114,4 +114,16 @@ public void delete(Key key) {
             }
         }
     }
+
+    @Override
+    public void clear() {
+        try {
+            getDiskCache().delete();
+            diskLruCache = null;
```
sjudd: I'd guess that findbugs (the reason for the build breakage) is complaining about the unsynchronized access to diskLruCache here.

library/src/main/java/com/bumptech/glide/load/engine/Engine.java
TWiStErRob: The name implies blocking work. Thing about how would you use it: call `clear` then call `.with.into`, how would you make `clear` finish before checking the cache during processing the request?

sjudd: I think I'm not 100% clear on the use case. If you're trying to make sure all entries are removed, there's no guarantee (asynchronous or not) that in flight requests won't still complete having loaded data from the cache after this method is called. 

Can anyone share a use case or two? What's the goal of clearing the cache? 

I'm still inclined to say a version number might be the best way to do this, because it would at least let you provide some guarantees (ie no request will complete with old data if you pass in an incremented disk cache version number to Glide as part of the GlideBuilder). Again though I'm not 100% clear on the use case.

Teovald:
My use-case is a music streaming app. We provide a setting allowing users to flush all the app caches.  
This is pretty common among music streaming apps in order to let user with limited storage space reclaim some of it.

sjudd:
Thanks for commenting, that makes sense. As long as no one expects this to be a promise as to where requests will or won't complete from, this looks good to me.

library/src/main/java/com/bumptech/glide/Glide.java
```diff
@@ -383,6 +383,13 @@ public void trimMemory(int level) {
     }
 
     /**
+     * Clears disk cache.
+     */
+    public void clearDiskCache() {
```
sjudd: Add an assertion that this is on a background thread (https://github.com/bumptech/glide/blob/master/library/src/main/java/com/bumptech/glide/util/Util.java#L135), and a comment that says it must be called from a background thread.

sjudd:
Thanks for putting this up, if you can add the assertion/comment on the method in Glide and fix the findbugs error, I'd appreciate it!

nemphys:
PR updated, please check

nemphys:
Sorry, one more update for the synchronization

sjudd:
Looks good to me, thanks!

library/src/main/java/com/bumptech/glide/load/engine/Engine.java
TWiStErRob: I found a use case: [ApplicationManifest.xml > application > @manageSpaceActivity](http://developer.android.com/guide/topics/manifest/application-element.html#logo), this attribute replaces the "Clear Data" button with a "Manage Storage" button which starts the above activity. An example I could think of is if a user needs space and sees that the app is huge, pressing "Clear Data" would wipe out all personalization/data. I know there's a "Clear Cache", but a lot of apps don't put their volatile data there for some reason. Anyway, I would put a "Clear image cache" button in my storage manager to allow freeing up space. Weirdly the [API for "Clear Data"](http://developer.android.com/reference/android/app/ActivityManager.html#clearApplicationUserData()) button was added in API 19, while the [attribute existed since API 1](http://developer.android.com/reference/android/content/pm/ApplicationInfo.html#manageSpaceActivityName)...

PR #353, opened 2015-02-24
Changes to ViewTarget class, so that it properly supports views with MATCH_PARENT dimensions.

nemphys:
Please check this one, too; I have done it a while ago and seems to work fine.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/request/target/ViewTarget.java
+++ b/library/src/main/java/com/bumptech/glide/request/target/ViewTarget.java
@@ -183,7 +183,7 @@ public abstract class ViewTarget<T extends View, Z> extends BaseTarget<Z> {
             if (isSizeValid(view.getHeight())) {
                 return view.getHeight();
             } else if (layoutParams != null) {
-                return getSizeForParam(layoutParams.height, true /*isHeight*/);
+                return getSizeForParam(view, layoutParams.height, true /*isHeight*/);
             } else {
                 return PENDING_SIZE;
             }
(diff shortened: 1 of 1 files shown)


TWiStErRob:
This doesn't look right to me. It seems to have too many assumptions about the layout hierarchy above the image view.

Consider for example a LinearLayout with an ImageView with match/match and a weight, one of those match parents will be wrongly "measured" by Glide. *Those width/height are just hints to the ViewGroups*.

Another case I can see failing is RelativeLayout where the contained ImageView has match parent but other constraints make it way smaller (toLeftOf).

I think this is why the current implementation is saying "it's pending" and lets the Android View hierarchy do the measure/layout cycle correctly.

nemphys:
Sounds fair, but I think that I went through the trouble to do this because the default implementation did not work properly (at least in my use case; please note that, as I said, I did this a while ago and can't remember exactly what was wrong). Will have to take a look at the code again and report back.

TWiStErRob:
The behaviour you referring to may have changed in Glide v3.5.0, see #274 and related commits.

nemphys:
OK, just took a look at the (unmodified) ViewTarget class and it seems that since getSizeForParam returns "param" when layoutParams is present and *not* wrap_content, isSizeValid returns false in checkCurrentDimens, which in turn just returns (no callbacks). Therefore, the view is not measured when match_parent is set.
Please correct me if I'm wrong.

TWiStErRob:
checkCurrentDimens should only be called when getWidth already returns the laid out size. Note that the entry point is getSize(cb), which also calls isSizeValid, which fails the first time for match_parent (not a valid size) and then registers the ViewTreeObserver. Which when called by Android calls checkCurrentDimens.

nemphys:
OK, you're right, but the images are still never loaded when I set my view to match_parent... Any ideas why?
Will try to debug this and see I find anything interesting.

nemphys:
I started debugging, but it gets a little complicated, since I am using a custom view inside a view pager and that's where the problem appears. Too late for today, will investigate further tomorrow.
Anyway, I am closing this PR, since it seems that the stock code works OK in simple use cases with match_parent enabled.
Thank you for your time.

TWiStErRob:
Feel free to open an issue with a minimal example project, with more context, like exact view hierarchy it's easier to see the problem. Also try just setting the image without Glide as control experiment?

nemphys:
After digging a little further, I found the culprit: the ImageView I am using in this specific scenario is actually an ImageView subclass that supports zooming/panning etc. and it handles its onMeasure event in a way that was conflicting with the callback; changed some bits and all is OK now.

The problem is that the callback method, although ingenious, causes trouble in other scenarios, too:
I am implementing shared element transitions between fragments (with match_parent ImageViews) and, in this case, the view measurements for the shared element animations are captured only once and pretty early, which means that the callback has not been called yet, leading to wrong measurements and weird animations.
All I can think of in order to overcome this is to make a custom Target subclass that will use this PR's logic to calculate dimensions (since the ImageView is the only child in my case, so it does not face the issue you mentioned). 
Any other ideas would be appreciated...

PR #376, opened 2015-03-24
Set view tag to null in clear(View view) to avoid memory leak.

zhywang:
If the ImageView is used in recyclable view (like ListView), when the request is cleared, should set ImageView tag to null, or else there maybe a reference to GlideBitmapDrawableResource, which takes a big memory.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/Glide.java
+++ b/library/src/main/java/com/bumptech/glide/Glide.java
@@ -450,6 +450,7 @@ public class Glide {
     public static void clear(View view) {
         Target<?> viewTarget = new ClearTarget(view);
         clear(viewTarget);
+        view.setTag(null);
     }
 
     /**


TWiStErRob:
It sounds like a fair point, did you find this in a memory dump or you're suspecting it?

I think `Glide` shouldn't know how `ViewTarget` stores the request, nor how any `Target` stores it.
It would probably be better and more general to put the clear line here:
```
public static void clear(Target<?> target) {
    Util.assertMainThread();
    Request request = target.getRequest();
    if (request != null) {
        request.clear();
+++     target.setRequest(null);
    }
}
```

sjudd:
Thanks for bringing this up!

I agree with @TWiStErRob, this should go into the clear method for Targets. 

However, I'm not sure this actually fixes a leak? Resources are [nulled out inside of the Request][1] when the Request is cleared already. It is possible that's not working quite right. I'd also be curious to know whether you saw this in a heap dump or not?

[1]: https://github.com/bumptech/glide/blob/master/library/src/main/java/com/bumptech/glide/request/GenericRequest.java#L308

TWiStErRob:
Oh, I forgot that it's the resource that he says leaks, but still, why keep the `Request` in memory?

By the way, @zhywang, if you're in a ListView then the next request into the recycled list item's image will overwrite the existing `Target` in the tag, and also the drawable will be cleared by the next request.

sjudd:
No good reason to keep the Request in memory, it's just a memory leak waiting to happen. I'm happy with the pull request.

However, I do want to double check and make sure this isn't leaking resources in a way I'm not currently seeing. There may be another bug, or something else we can/should do.

zhywang:
@sjudd Yes I did see the references to GlideBitmapDrawableResource from the ImageView.tag in the heap dump of our app. After more spike, I found this patch is not good enough, the root cause is that after the request is finished, it still keeps the GlideBitmapDrawableResource reference, so to avoid memory leak, maybe we should do a Glide.cancel() on the image view after  the request is finished, in the client (our app) side.

zhywang:
@TWiStErRob Yes you are correct, in a ListView the next request into the item will overwrite the Target and clear the drawable, however, if I have many types in the ListView(http://developer.android.com/reference/android/widget/BaseAdapter.html#getItemViewType%28int%29), maybe some items won't be overwrite (that's what our app does), then the  drawables is in memory, causes the memory leak. I think for this case, maybe it's better to handle it in our app.

sjudd:
@zhywang Yeah you if you have a lot of different view types you may want to consider adding a [ListView RecyclerListener][1] or a [RecyclerView RecyclerListener][2] and calling clear. Both the Request and the View will reference whatever resource you've loaded. ``clear()`` should remove both references.

[1]: http://developer.android.com/reference/android/widget/AbsListView.RecyclerListener.html
[2]: https://developer.android.com/reference/android/support/v7/widget/RecyclerView.RecyclerListener.html

zhywang:
@sjudd RecyclerListener is cool. In our app the image is in a complex layout, I resolve this problem by add a OnAttachStateChangeListener to it, to avoid impact our architecture too much.
I think this pull request can be closed now, maybe in the future glide could use the OnAttachStateChangeListener to clear the request (that's how Google Volley NetworkImageView does) :)

sjudd:
Got it thanks!

LiHanGuang:
@zhywang @TWiStErRob
I have the problem like you when  using Viewpager,
my solution below is also useful for ListView and RecycleView.

My Solution:

//ImageBrowserAdapter .java
public class ImageBrowserAdapter extends PagerAdapter{
     ...other method

     @Override
    public void destroyItem(ViewGroup container, int position, Object object) {
        Log.i(TAG, "destroyItem() position = " + position % mImages.size());
        View view = (View) object;
        container.removeView(view);
        final PhotoView photoView = (PhotoView) view.findViewById(R.id.image);
        Log.i(TAG, "destroyItem() photoView " + String.valueOf(photoView));

        Glide.clear(photoView);        //clear glide request for View, avoid memory leak
    } 
}

//ImagePagerBrowserActivity.java
public class ImagePagerBrowserActivity{
...other method

@Override
	public void onDestroy() {
		super.onDestroy();

                mScrollViewPager.setAdapter(null);         //avoid memory leak
		mAdapter = null;
		mImages.clear();
		mImages = null;
	}
}

Reason:
android.support.v4.view.ViewPager.java

/**
     * Set a PagerAdapter that will supply views for this pager as needed.
     *
     * @param adapter Adapter to use
     */
    public void setAdapter(PagerAdapter adapter) {
        if (mAdapter != null) {
            mAdapter.unregisterDataSetObserver(mObserver);
            mAdapter.startUpdate(this);
            for (int i = 0; i < mItems.size(); i++) {
                final ItemInfo ii = mItems.get(i);
                mAdapter.destroyItem(this, ii.position, ii.object); //   i am here!!!
            }
            mAdapter.finishUpdate(this);
            mItems.clear();
            removeNonDecorViews();
            mCurItem = 0;
            scrollTo(0, 0);
        }
       
        ....  other code 

        //you can also instead of using the "inteface AdapterChangeListener"  to slove this  memory leak

        if (mAdapterChangeListener != null && oldAdapter != adapter) {
            mAdapterChangeListener.onAdapterChanged(oldAdapter, adapter);  
        }
}

PR #397, opened 2015-04-10
Fix inconsistent code styles in OkHttpStreamFetcher.

floating-cat:


sjudd:
Thanks!

TWiStErRob:
Thanks for the PR!

Some (late) feedback: I prefer putting the if outside the try, because entering/exiting try may be slower.

Even though I didn't find any hard evidence on this in the last 10 minutes of research, this statement is convinced me to continue preferring it:
> Placing code inside a try-catch block inhibits certain optimizations that modern JVM implementations might otherwise perform.

It also helps devs to skip the try block if they're reading code and they know that a certain variable's nullity.

sjudd:
Makes sense, when we pull this stuff into a utility method (more pull requests always welcome :)), we can  if outside of try.

PR #401, opened 2015-04-13
Do not try to clear a non existent request

Tolriq:
During some test with Glide and FutureTarget it appears that when you cancel a FutureTarget very fast, the request may not yet be set in RequestFutureTarget leading to a crash.

This simple PR fix this case.


sjudd:
As you mentioned in #402 there's probably more going wrong here. I'm going to accept this PR anyway, because no Target should assume that its request is non-null.

PR #444, opened 2015-04-29
4.0a

AalenChan:


TWiStErRob:
Wow, that's a lot of reading... Did you make it 444 or purpose? :)
I hope you'll wait for 3.6.0 with the merge otherwise you break snapshot users.

sjudd:
Yeah we need to hold off on merging this for a little while yet :)

PR #477, opened 2015-05-29
Added GifDrawable#getFrameIndex() for GIF progress API

joshzana:
Fixes 446

### DIFF
--- a/library/src/main/java/com/bumptech/glide/load/resource/gif/GifDrawable.java
+++ b/library/src/main/java/com/bumptech/glide/load/resource/gif/GifDrawable.java
@@ -134,6 +134,13 @@ public class GifDrawable extends Drawable implements GifFrameLoader.FrameCallbac
     return state.frameLoader.getFrameCount();
   }
 
+  /**
+   * Returns the current frame index, or -1 if no frame is displayed.
+   */
+  public int getFrameIndex() {
+    return state.frameLoader.getCurrentIndex();
+  }
+
   private void resetLoopCount() {
     loopCount = 0;
   }
@@ -254,7 +261,7 @@ public class GifDrawable extends Drawable implements GifFrameLoader.FrameCallbac
 
   @TargetApi(Build.VERSION_CODES.HONEYCOMB)
   @Override
-  public void onFrameReady(int frameIndex) {
+  public void onFrameReady() {
     if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.HONEYCOMB && getCallback() == null) {
       stop();
       invalidateSelf();
@@ -263,7 +270,7 @@ public class GifDrawable extends Drawable implements GifFrameLoader.FrameCallbac
 
     invalidateSelf();
 
-    if (frameIndex == state.frameLoader.getFrameCount() - 1) {
+    if (getFrameIndex() == getFrameCount() - 1) {
       loopCount++;
     }
 
--- a/library/src/main/java/com/bumptech/glide/load/resource/gif/GifFrameLoader.java
+++ b/library/src/main/java/com/bumptech/glide/load/resource/gif/GifFrameLoader.java
@@ -46,7 +46,7 @@ class GifFrameLoader {
   private Transformation<Bitmap> transformation;
 
   public interface FrameCallback {
-    void onFrameReady(int index);
+    void onFrameReady();
   }
 
   public GifFrameLoader(Context context, GifDecoder gifDecoder, int width, int height,
@@ -119,6 +119,10 @@ class GifFrameLoader {
 
   int getSize() {
     return gifDecoder.getByteSize() + getFrameSize();
+  }
+
+  int getCurrentIndex() {
+    return current != null ? current.index : -1;
   }
 
   private int getFrameSize() {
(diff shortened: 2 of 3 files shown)


library/src/main/java/com/bumptech/glide/load/resource/gif/GifDrawable.java
```diff
@@ -65,6 +65,10 @@
    * The number of times to loop through the gif animation.
    */
   private int maxLoopCount = LOOP_FOREVER;
+  /**
+   * The index of the last drawn frame.
+   */
+  private int frameIndex = -1;
```
sjudd: I think this should default to 0. We actually pass in the first frame of the GIF into the Drawable, so it should always be at some index in the range 0..frame count -1.

library/src/main/java/com/bumptech/glide/load/resource/gif/GifDrawable.java
```diff
@@ -134,6 +138,10 @@ public int getFrameCount() {
     return state.frameLoader.getFrameCount();
   }
 
+  public int getFrameIndex() {
```
sjudd: Add a javadoc for this method (returns the current frame index in range 0 to count-1, or -1 if no frame is displayed.

sjudd:
Looks good to me, thanks for doing this. If you want to make those changes and squash your commits into one I'd be happy to  merge. 

I'm not sure what's going on with travis, it may have temporarily errored. I'll keep an eye on it and update if it looks like it wants something changed.

TWiStErRob:
I'm confused. In 3.6.0 there is a: `gifDrawable.getDecoder().getCurrentFrameIndex()`. It looks like [the 4.0 refactor got rid of that](https://github.com/joshzana/glide/commit/cf12dc225ed016f337e6c1d7685017395c3a85bf#diff-07047b243c026136d33d4621d275401dL143), but still the current frame is available through:
`state.frameLoader.gifDecoder.getCurrentFrameIndex()` which I think is preferable to having a copy of the frame index, say, to maintain compatibility.

As an alternative I think `state.frameLoader.current.index` (with a new getter on `frameLoader`) could be used to be in sync with how [`draw` uses `state.frameLoader.current`](https://github.com/joshzana/glide/blob/master/library/src/main/java/com/bumptech/glide/load/resource/gif/GifDrawable.java#L229) ([`getCurrentFrame()`](https://github.com/joshzana/glide/blob/master/library/src/main/java/com/bumptech/glide/load/resource/gif/GifFrameLoader.java#L171)).

sjudd:
@TWiStErRob maintaining a copy here is ok I think. Otherwise you'd always return the index of the frame you're loading, rather than the one you're displaying. It's also always possible that an intermediate frame will fail to load, so you you can't even guarantee that the frame you're loading is just the next frame after the one you're displaying.

Still looking in to the travis build...

TWiStErRob:
Take a look at the alternative I proposed too, because that's totally different. That would be the only way to get the same index as draw gets or the bitmap should be retrieved at the same time the index is cached.

sjudd:
Ah I see what you're saying. That would make a bit more sense with the current structure, really the Drawable doesn't need to know anything about the current index. We should also probably be keeping track of loops some other way, rather than passing the index into the callback.

joshzana:
Thanks for the feedback!  I'm happy to fix the implementation. As of now, GifFrameLoader.gifDecoder is private with no accessor, so something will have to change.  I'll go with @TWiStErRob 's "alternative" approach of exposing the index off of DelayTarget.  I'm not going to mess with loopCount as that seems a bit more involved.

joshzana:
One question - With this change, it's pretty easy to refactor out the frameIndex parameter from the onFrameReady callback interface, since that method can just call getFrameIndex().  Whats nice is that makes it consistent where the code gets the index from.  The biggest impact is the tests, which require a few extra mockito calls.  Thoughts?

TWiStErRob:
I like the idea, but wait for @sjudd.

sjudd:
Sounds reasonable to me, go for it!

joshzana:
One interesting thing I noticed during testing - the first time after starting the animation in GifDrawable#onFrameReady, the index is 1.  This is because both ByteBufferGifDecoder#decode and GifFrameLoader#loadNextFrame call advance before onFrameReady.  After looping it does properly start at 0.  I am not sure if this is by design or not, but I'm happy to open a separate issue if needed.

TWiStErRob:
Nice! Sorry for the 2 day detour ;)

library/src/main/java/com/bumptech/glide/load/resource/gif/GifDrawable.java
```diff
@@ -134,6 +134,14 @@ public int getFrameCount() {
     return state.frameLoader.getFrameCount();
   }
 
+  /**
+   * Returns the current frame index in the range of 0 to {@link #getFrameCount()}-1.
```
sjudd: nit: remove 'range of 0 to {@link #getFrameCount()} -1 or -1 if no frame is displayed.

Remove '.' and update the 'of's in the second line to 'or' and 'if' respectively

sjudd:
Awesome LGTM, thanks for doing this!

library/src/main/java/com/bumptech/glide/load/resource/gif/GifDrawable.java
joshzana: Sure, though I only added the range business due to your comment above :-)

sjudd: Urk, sorry I reworded my comment forgot to remove a critical word. Thanks, this is fine, I'll adjust the comment and merge, sorry for the confusion!

sjudd:
Merged here: https://github.com/bumptech/glide/commit/d8d32184211799f6839527bef6a20eca77c1d6b5

Thanks!

PR #498, opened 2015-06-16
Orozco Changes to FlickrSearchActivity

khorozco:
images that appear once the sample app is opened are the image results from a "puppies" search; both methods previously named registerSearchListener and unregisterSearchListener are now named registerQueryListener and unregisterQueryListener.

### DIFF
--- a/samples/gallery/src/main/java/com/bumptech/glide/samples/gallery/RecyclerAdapter.java
+++ b/samples/gallery/src/main/java/com/bumptech/glide/samples/gallery/RecyclerAdapter.java
@@ -1,8 +1,5 @@
 package com.bumptech.glide.samples.gallery;
 
-import static com.bumptech.glide.request.RequestOptions.fitCenterTransform;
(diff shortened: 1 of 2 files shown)


khorozco:
I signed the CLA.

samples/flickr/build/reports/checkstyle/checkstyle.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
sjudd: @khorozco It looks like you added this file accidentally, you can remove it with:

git rm samples/flickr/build/reports/checkstyle/checkstyle.xml

And then amend this commit with:

git commit --amend

Finally when you try to push, you will see an error because you've modified history, but you can work around that using --force:

git push origin master --force

Force pushing to your personal remote is ok because you can be pretty confident no one other than you is going to be working off of it. You see a warning because on shared repositories it's hard to contribute if history isn't linear.

PR #565, opened 2015-08-05
Make it compile on Windows and fix some deprecation warnings

TWiStErRob:


integration/gifencoder/src/test/java/com/bumptech/glide/integration/gifencoder/ReEncodingGifResourceEncoderTest.java
```diff
@@ -307,6 +310,8 @@ public void testRecyclesFrameResourceAfterWritingIfFrameResourceIsNotTransformed
   @Test
   public void testWritesBytesDirectlyToDiskIfTransformationIsUnitTransformation()
       throws IOException {
+    // Most likely an instance of http://stackoverflow.com/q/991489/253468
+    assumeTrue(!System.getProperty("os.name").startsWith("Windows"));
```
sjudd: Maybe move this into the test Util file?

TWiStErRob: I added this to tests/Util.java
> FIXME move to testutil module

There are too many utils and didn't know which one to move to; the imported project has some missing dependencies and the IDEA code is really red; so I assumed that `:library:testCompile` files are not available in `:integration:gifencoder:testCompile`.

In general module `a` may depend on module `b` but usually module `a tests` are usually not dependent on module `b test`.

sjudd: Fair enough, thanks for looking.

PR #603, opened 2015-08-28
enable to add frame with specified position for each frames

conanoc:
I changed AnimatedGifEncoder to let it add frame with different position and different size.
The frames in Animated Gif can have its own position and size. It can have only the delta image different from the previous frame, thus reduce the size of the final GIF file.

The previous version of AnimatedGifEncoder assumed that all the frames has the same size of image and the position of the frame is (0,0). Now, we can use small size images for the secondary images for animation.

Only the AnimatedGifEncoder class has changed slightly.

### DIFF
--- a/third_party/gif_encoder/src/main/java/com/bumptech/glide/gifencoder/AnimatedGifEncoder.java
+++ b/third_party/gif_encoder/src/main/java/com/bumptech/glide/gifencoder/AnimatedGifEncoder.java
@@ -22,6 +22,7 @@ import java.io.OutputStream;
  *     e.setDelay(1000);   // 1 frame per sec
  *     e.addFrame(image1);
  *     e.addFrame(image2);
+ *     e.addFrame(image3, 100, 100);    // set position of the frame
  *     e.finish();
  * </pre>
  *
@@ -44,6 +45,10 @@ public class AnimatedGifEncoder {
     private int width; // image size
 
     private int height;
+
+    private int fixedWidth;   // set by setSize()
+
+    private int fixedHeight;
 
     private Integer transparent = null; // transparent color if given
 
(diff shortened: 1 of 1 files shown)


conanoc:
I signed it

third_party/gif_encoder/src/main/java/com/bumptech/glide/gifencoder/AnimatedGifEncoder.java
```diff
@@ -140,22 +145,43 @@ public void setTransparent(int color) {
...
+     * Adds next GIF frame to the specified position. The frame is not written immediately, but is actually
+     * deferred until the next frame is received so that timing data can be
+     * inserted. Invoking <code>finish()</code> flushes all frames. If
+     * <code>setSize</code> was invoked, the size is used for all subsequent frames.
+     * Otherwise, the actual size of the image is used for each frames.
+     *
+     * @param im
+     *          BufferedImage containing frame to write.
+     * @param x
+     *          x position of this frame
```
sjudd: Can you clarify a bit more what position means here? position means offset relative to the upper left hand corner of all of the other frames of the gif? Does adding a position clip the parts of the frame that exceed the bounds of the other frames, or is the gif expanded to position + width and position + height?

A link to the GIF spec describing what position is used for and how it works would be sufficient.

third_party/gif_encoder/src/main/java/com/bumptech/glide/gifencoder/AnimatedGifEncoder.java
```diff
@@ -252,15 +278,30 @@ public void setQuality(int quality) {
...
+        if (fixedWidth < 1)
+            fixedWidth = 320;
+        if (fixedHeight < 1)
+            fixedHeight = 240;
+
+        sizeSet = true;
+    }
+
+    /**
+     * Sets current GIF frame size.
```
sjudd: Would you mind adding a clarifying comment about how this differs from setSize and when to use one vs the other?

sjudd:
This looks reasonable to me, just a few requests for more comments. If you could also use braces for your if statements, I'd appreciate. I know it doesn't quite match the current style of the encoder, but it's the practice we should use moving forward:

if (true) {
}

Thanks for putting this up!

TWiStErRob:
@sjudd & @conanoc, doesn't that 320×240 look too arbitrary to you? Is that from the spec? I think an IllegalArgumentException or an IllegalStateException would be more fitting.

sjudd:
@TWiStErRob I'm sort of inclined not to mess with it in this change, it's not new and is largely tangential to changes made here.

TWiStErRob:
Agreed, just wanted to bring it up.

third_party/gif_encoder/src/main/java/com/bumptech/glide/gifencoder/AnimatedGifEncoder.java
conanoc: see page 11 of http://giflib.sourceforge.net/gif89.txt describing Image Descriptor.
let me quote.

           ii) Image Left Position - Column number, in pixels, of the left edge
           of the image, with respect to the left edge of the Logical Screen.
           Leftmost column of the Logical Screen is 0.

           iii) Image Top Position - Row number, in pixels, of the top edge of
           the image with respect to the top edge of the Logical Screen. Top
           row of the Logical Screen is 0.

A frame overwrites the previous frame with the given position and size.

third_party/gif_encoder/src/main/java/com/bumptech/glide/gifencoder/AnimatedGifEncoder.java
conanoc: setFrameSize() sets the size of image for each frame. The width and height used here is the actual size of the image. This is a private function for internal use.
setSize() is a public function which will be used by some user who wants to set the size of the overall frame of the target GIF.

setSize was originally used for the two purposes mentioned above and we could not use different sizes for each frames. So I split it.

TWiStErRob:
@conanoc thanks for the clarifications for Sam's questions. It think what he meant is to put the information you quoted in your replies to the JavaDoc, updating your existing commit, and making it available to anyone reading the code. Same with braces: use {} where you modified lines anyway.

conanoc:
@sjudd, @TWiStErRob I added more comments and braces.

sjudd:
LGTM, thanks!

helloyako:
very nice feature. I want this.

PR #620, opened 2015-09-11
Let gif's first frame using last frame's dispose

kojilin:
When playing transparent gif in second loop, first frame will keep last frame's data in the view.
Using last frame's dispose if exist.


sjudd:
The change here looks good to me, do you mind rebasing and squashing your commits?

Also is this the expected behavior for GIFs? Do browsers typically wrap disposal of the last frame back around to the first? I looked for a bit and couldn't find much documentation on what the generally accepted behavior is.

kojilin:
@sjudd Hi, I think it would be more like browsers, when we test transparent gif with "(replace) dispose" in last frame. Safari and chrome will clear the last frame and redraw the first frame.
If we remove the "(replace) dispose" in the last frame, then it will draw the first frame over last frame in second loop.

I test some gif from http://giphy.com/search/transparent, like http://giphy.com/gifs/photobucket-pictures-smiley-qf9KfWCvhb1M4 and http://rebloggy.com/post/gif-pikachu-pokemon-kawaii-pokemon-gif-transparent-transparent-gif/45062041815

sjudd:
Ok Awesome, thanks for the explanation and the PR, LGTM!

PR #709, opened 2015-10-23
make ConnectivityMonitorFactory injectable

emakar:
Our app needs some higher level connectivity states not related to just android.net.ConnectivityManager (e.g. work via wi-fi ony or offline at despite device connectivity), so this change allows Glide to respect such states

### DIFF
--- a/library/src/main/java/com/bumptech/glide/Glide.java
+++ b/library/src/main/java/com/bumptech/glide/Glide.java
@@ -294,7 +294,7 @@ public class Glide implements ComponentCallbacks2 {
     return arrayPool;
   }
 
-  public ConnectivityMonitorFactory getConnectivityMonitorFactory() {
+  ConnectivityMonitorFactory getConnectivityMonitorFactory() {
     return connectivityMonitorFactory;
   }
 
--- a/library/src/main/java/com/bumptech/glide/Glide.java
+++ b/library/src/main/java/com/bumptech/glide/Glide.java
@@ -62,6 +62,7 @@ import com.bumptech.glide.load.resource.gif.StreamGifDecoder;
 import com.bumptech.glide.load.resource.transcode.BitmapBytesTranscoder;
 import com.bumptech.glide.load.resource.transcode.BitmapDrawableTranscoder;
 import com.bumptech.glide.load.resource.transcode.GifDrawableBytesTranscoder;
+import com.bumptech.glide.manager.ConnectivityMonitorFactory;
 import com.bumptech.glide.manager.RequestManagerRetriever;
 import com.bumptech.glide.module.GlideModule;
 import com.bumptech.glide.module.ManifestParser;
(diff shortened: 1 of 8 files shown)


emakar:
I signed it!

TWiStErRob:
Hi, here's what I would change:
 1. I think the `connectivity` package is unnecessary, it just fragments the packages unnecessarily, also not introducing it keeps more compatibility
 2. `ConnectivityMonitorFactoryImpl` -> `DefaultConnectivityMonitorFactory` as it tries to create `DefaultConnectivityMonitor`. (AFIAR the Impl suffix is not used anywhere else in Glide)
 3. The FQCN in JavaDoc links is the style Sam's going for I think.

emakar:
Thanks! Committed code style and naming fixes. I've rebased the changes as well

TWiStErRob:
Is that new constructor in `RequestManager` used anywhere?

emakar:
Nope, just telescoping constructor in case someone decides to use it direclty (so as for now it's used in another ctor only)

TWiStErRob:
I don't think you ever want to use `RequestManager` constructor directly, as far as I know it's public because it's accessed from another package. It should be always created/returned by a `Glide.with` call.

emakar:
Agreed, fixed

emakar:
Rebased.

library/src/main/java/com/bumptech/glide/Glide.java
```diff
@@ -290,6 +294,10 @@ public ArrayPool getArrayPool() {
     return arrayPool;
   }
 
+  public ConnectivityMonitorFactory getConnectivityMonitorFactory() {
```
sjudd: Can you make this package private?

library/src/test/java/com/bumptech/glide/manager/ConnectivityMonitorFactoryTest.java
```diff
@@ -18,7 +18,7 @@
 
   @Before
```
sjudd: Can you rename this DefaultConnectivityMonitorFactoryTest to match the class name?

sjudd:
Other than two small nits, looks good to me. Thanks for putting this up.

If any of the changes you need for your app are generally applicable, I'd welcome a pull request to make them to the default connectivity monitor as well.

emakar:
Thanks! Fixed encapsulation and naming.
As for the custom monitor I think it's rather specific than general. Our main purpose is just notifying ConnectivityListener. And it's performed with complicated "connectivity monitor" which besides common implementation uses several app-private restrictions (bunch of network modes, etc). So such ConnectivityMonitor is just a thin client/observer of existent complicated api.

sjudd:
Merged here: b334f4f145a899d3e2c013d222d4ebcf6eb94f0d

Sorry for the lack of credit, we're still trying to work out the kinks in this new merging scheme...

PR #717, opened 2015-10-29
Swap clearing memory cache and pool to prevent re-pooling Bitmaps

### LINKED ISSUE #687: About bitmap recycle
when LruMemoryCache full, LruMemoryCache remove some bitmap, these bitmap put BitmapPool , these bitmap reuse for other Bitmap Request。if these bitmap is using in ImageView,Will lead to errors?

TWiStErRob:
Fixes #687


sjudd:
Thanks!

PR #727, opened 2015-11-05
Implemented CENTER_INSIDE support

### LINKED ISSUE #591: Why not have centerInside()?
Sometimes we need to load image for center inside, but I have not found this method!

Can add method for center inside?

josemontiel:
Fixes #591

### DIFF
--- a/library/src/main/java/com/bumptech/glide/load/resource/bitmap/TransformationUtils.java
+++ b/library/src/main/java/com/bumptech/glide/load/resource/bitmap/TransformationUtils.java
@@ -139,6 +139,32 @@ public final class TransformationUtils {
   }
 
   /**
+   * If the Bitmap is smaller or equal to the Target it returns the original size, if not then
+   * {@link #fitCenter(BitmapPool, Bitmap, int, int)} is called instead.
+   *
+   * @param pool   The BitmapPool obtain a bitmap from.
+   * @param inBitmap  The Bitmap to center.
+   * @param width  The width in pixels of the target.
+   * @param height The height in pixels of the target.
+   * @return returns input Bitmap if smaller or equal to target, or toFit if the Bitmap's width or
+   * height is larger than the given dimensions
+   */
+  public static Bitmap centerInside(@NonNull BitmapPool pool, @NonNull Bitmap inBitmap, int width,
+                                 int height) {
+    if (inBitmap.getWidth() <= width && inBitmap.getHeight() <= height) {
+      if (Log.isLoggable(TAG, Log.VERBOSE)) {
+        Log.v(TAG, "requested target size larger or equal to input, returning input");
+      }
+      return inBitmap;
+    } else {
+      if (Log.isLoggable(TAG, Log.VERBOSE)) {
+        Log.v(TAG, "requested target size too big for input, fit centering instead");
+      }
+      return fitCenter(pool, inBitmap, width, height);
+    }
+  }
+
+  /**
    * Sets the alpha of the Bitmap we're going to re-use to the alpha of the Bitmap we're going to
    * transform. This keeps {@link android.graphics.Bitmap#hasAlpha()}} consistent before and after
    * the transformation for transformations that don't add or remove transparent pixels.
--- a/library/src/main/java/com/bumptech/glide/request/BaseRequestOptions.java
+++ b/library/src/main/java/com/bumptech/glide/request/BaseRequestOptions.java
@@ -18,6 +18,7 @@ import com.bumptech.glide.load.engine.DiskCacheStrategy;
 import com.bumptech.glide.load.resource.bitmap.BitmapDrawableTransformation;
 import com.bumptech.glide.load.resource.bitmap.BitmapEncoder;
 import com.bumptech.glide.load.resource.bitmap.CenterCrop;
+import com.bumptech.glide.load.resource.bitmap.CenterInside;
 import com.bumptech.glide.load.resource.bitmap.CircleCrop;
 import com.bumptech.glide.load.resource.bitmap.DownsampleStrategy;
 import com.bumptech.glide.load.resource.bitmap.Downsampler;
@@ -526,6 +527,34 @@ public abstract class BaseRequestOptions<CHILD extends BaseRequestOptions<CHILD>
    */
   public CHILD fitCenter(Context context) {
     return transform(context, DownsampleStrategy.CENTER_INSIDE, new FitCenter(context));
+  }
+
+  /**
+   * Applies {@link com.bumptech.glide.load.resource.bitmap.CenterInside} to all default types, and
+   * ignores unknown types.
+   *
+   * <p>This will override previous calls to {@link #dontTransform()}.
+   *
+   * @param context Any {@link android.content.Context}.
+   * @see #optionalTransform(Class, com.bumptech.glide.load.Transformation)
+   * @see #centerInside(Context) (android.content.Context)
+   */
+  public CHILD optionalCenterInside(Context context) {
+    return optionalTransform(context, DownsampleStrategy.DOWNSCALE_ONLY, new CenterInside(context));
+  }
+
+  /**
+   * Applies {@link com.bumptech.glide.load.resource.bitmap.CenterInside} to all default types and
+   * throws an exception if asked to transform an unknown type.
+   *
+   * <p>This will override previous calls to {@link #dontTransform()}.
+   *
+   * @param context Any {@link android.content.Context}.
+   * @see #transform(Class, com.bumptech.glide.load.Transformation)
+   * @see #optionalCenterInside(Context) (android.content.Context)
+   */
+  public CHILD centerInside(Context context) {
+    return transform(context, DownsampleStrategy.DOWNSCALE_ONLY, new CenterInside(context));
   }
 
   /**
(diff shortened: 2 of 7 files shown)


library/src/main/java/com/bumptech/glide/load/resource/bitmap/TransformationUtils.java
```diff
@@ -139,6 +139,55 @@ public static Bitmap fitCenter(@NonNull BitmapPool pool, @NonNull Bitmap inBitma
...
+      if (Log.isLoggable(TAG, Log.VERBOSE)) {
+        Log.v(TAG, "requested target size too big for input, fit centering instead");
+      }
+      return fitCenter(pool, inBitmap, width, height);
+    }
+    final int horizontalPadding = (width - inBitmap.getWidth()) / 2;
+    final int verticalPadding = (height - inBitmap.getHeight()) / 2;
+
+    Bitmap.Config config = getSafeConfig(inBitmap);
+    Bitmap toReuse = pool.get(width, height, config);
```
TWiStErRob: Pool may return null (at least in 3.0 it did)

library/src/main/java/com/bumptech/glide/request/BaseRequestOptions.java
```diff
@@ -529,6 +530,34 @@ public CHILD fitCenter(Context context) {
...
+
+  /**
+   * Applies {@link com.bumptech.glide.load.resource.bitmap.CenterInside} to all default types and
+   * throws an exception if asked to transform an unknown type.
+   *
+   * <p>This will override previous calls to {@link #dontTransform()}.
+   *
+   * @param context Any {@link android.content.Context}.
+   * @see #transform(Class, com.bumptech.glide.load.Transformation)
+   * @see #optionalFitCenter(android.content.Context)
```
TWiStErRob: see points to fitCenter

library/src/main/java/com/bumptech/glide/request/BaseRequestOptions.java
```diff
@@ -529,6 +530,34 @@ public CHILD fitCenter(Context context) {
...
 
   /**
+   * Applies {@link com.bumptech.glide.load.resource.bitmap.CenterInside} to all default types, and
+   * ignores unknown types.
+   *
+   * <p>This will override previous calls to {@link #dontTransform()}.
+   *
+   * @param context Any {@link android.content.Context}.
+   * @see #optionalTransform(Class, com.bumptech.glide.load.Transformation)
+   * @see #fitCenter(android.content.Context)
```
TWiStErRob: see points to fitCenter

.idea/codeStyleSettings.xml
```diff
@@ -291,6 +291,34 @@
         <option name="DOWHILE_BRACE_FORCE" value="3" />
         <option name="WHILE_BRACE_FORCE" value="3" />
         <option name="FOR_BRACE_FORCE" value="3" />
+        <Objective-C-extensions>
```
TWiStErRob: This shouldn't be here.

TWiStErRob:
I have one bigger  concern (other than my small line comments):
Applying .centerInside() to a load may produce an unnecessarily large Bitmap. Imagine one's loading a 16x16px icon into a 512x512px space. This means that Glide's centerInside will create a 512px Bitmap, but there will be 240px padding inside where there's nothing. This is a waste of memory considering the ImageView would handle this for us if it's ScaleType is set to centerInside. So I'm inclined to say that the automatic transformation for ImageViews should be a little different: return the original Bitmap if it's smaller than the requested size, and fitCenter if it's bigger. **But**, the padded version is still viable in situations like toBytes() or asBitmap() with custom targets because there it is better to have a Bitmap that has the requested size than implementing centerInside somehow every single time. The best solution would be to parameterize all centerInside calls with a `boolean padSmaller` argument, so the user can/has to decide which logic to use.

library/src/main/java/com/bumptech/glide/load/resource/bitmap/TransformationUtils.java
```diff
@@ -139,6 +139,55 @@ public static Bitmap fitCenter(@NonNull BitmapPool pool, @NonNull Bitmap inBitma
...
+    if (inBitmap.getWidth() > width || inBitmap.getHeight() > height) {
+      if (Log.isLoggable(TAG, Log.VERBOSE)) {
+        Log.v(TAG, "requested target size too big for input, fit centering instead");
+      }
+      return fitCenter(pool, inBitmap, width, height);
+    }
+    final int horizontalPadding = (width - inBitmap.getWidth()) / 2;
+    final int verticalPadding = (height - inBitmap.getHeight()) / 2;
+
+    Bitmap.Config config = getSafeConfig(inBitmap);
```
TWiStErRob: I think this needs to upgrade to 8888 if there's a padding, otherwise we'll end up with a color that someone won't like (BLACK/WHITE are both wrong choices), it should be TRANSPARENT.

josemontiel:
Thanks for the feedback @TWiStErRob. Right on it!

.idea/codeStyleSettings.xml
sjudd: The IDE adds this automatically, if you rebase this will go away (I just committed this I think).

sjudd:
I think you can disregard padding entirely, it's hard for me to imagine a scenario where you want extra transparency baked into the Bitmap.

I'd just check if the width and height are greater than the target, and if so, fitCenter, and otherwise just return the original Bitmap.

Transformations never promise that the Bitmap they return will exactly match the requested width and height.

Thanks for putting this up!

TWiStErRob:
@sjudd what if the requested transformation doesn't match scaleType?
```java
ImageView.scaleType = fitCenter
.apply(RequestOptions.centerInside())
```
I would still expect a small image to be centered in a middle of a bigger space.

sjudd:
Oh I wouldn't, I'd expect the Bitmap to remain at the smallest possible size (to minimize memory usage), but for the Bitmap to be stretched by the view when it's drawn to match the requested scale type. 

None of our other transformations try to override a conflicting scale type set on a view, I don't think we need to do so here either.

TWiStErRob:
@sjudd ah, you're right; the ImageView wins; @josemontiel sorry about that.

josemontiel:
@sjudd so just to be sure, it should check if the Bitmap's dimensions are larger than the target and call fitCenter, otherwise return the input as is?

sjudd:
@josemontiel I believe so? I think that matches what ImageView does?

.idea/codeStyleSettings.xml
josemontiel: I tried rebasing but it won't go away

TWiStErRob: You could just revert it manually in a new commit and then squash your changes.

TWiStErRob:
What do you both think about the Downsampler issue? NONE will prevent huge images from loading with centerInside.

josemontiel:
That definitely would be a problem, perhaps the NONE strategy could be overridden in the Downsampler when the source's dimensions are larger than the target's?

TWiStErRob:
Hmm, there's already a [`DownsamplerStrategy.CENTER_INSIDE`](https://github.com/bumptech/glide/blob/master/library/src/main/java/com/bumptech/glide/load/resource/bitmap/DownsampleStrategy.java#L19) but it doesn't match what we need, let's see what @sjudd says about this, he knows that part better. I think that current `CENTER_INSIDE` strategy should be called `FIT_INSIDE`.

josemontiel:
Hmm interesting @TWiStErRob . It could be something like this:

    private static class RealCenterInside extends DownsampleStrategy {
        @Override
        public float getScaleFactor(int sourceWidth, int sourceHeight, int requestedWidth,
            int requestedHeight) {
      
          if (sourceWidth > requestedWidth || sourceHeight > requestedHeight) {
            float widthPercentage = requestedWidth / (float) sourceWidth;
            float heightPercentage = requestedHeight / (float) sourceHeight;
            return Math.min(widthPercentage, heightPercentage);
          } else {
            return 1f;
          }
        }

        @Override
        public SampleSizeRounding getSampleSizeRounding(int sourceWidth, int sourceHeight,
            int requestedWidth, int requestedHeight) {
          return SampleSizeRounding.QUALITY;
        }
    }

sjudd:
Yeah I think you're going to need a new downsample strategy that's equivalent to CENTER_INSIDE but that doesn't upscale.

The names here are pretty poor, sorry for that. Is there some standard naming convention for these transformations? Alternative suggestions welcome. I'm not totally sure that FIT_INSIDE is much clearer than CENTER_INSIDE, though I agree CENTER_INSIDE is poor. If we rename CENTER_INSIDE to FIT_INSIDE, what do we call the equivalent strategy that doesn't downsample? 

@josemontiel That looks right, but I think you could simplify by just returning ``Math.min(1f, CENTER_INSIDE.getScaleFactor(...))``? We want the same transformation, except that we don't want to upscale?

TWiStErRob:
@sjudd I'm not aware of any standards here, I created #730 with my suggestions/questions, please take a look.

@josemontiel I think you should just go with the `RealCenterInside` naming and we'll rename it by fixing #730.

josemontiel:
@TWiStErRob I'll go with DownscaleOnly, makes sense.

TWiStErRob:
`gradlew :library:checkstyle` failed because of long lines in DownsampleStrategy, use _Format code_ in IDEA to wrap automatically. Can you please fix that and squash your commits into 1?

TWiStErRob:
@sjudd this looks ready.

sjudd:
LGTM, I'll merge internally as soon as I can and push out. 

I'll probably rename DOWNSCALE_ONLY per our brief discussion on your doc @TWiStErRob since this isn't the only strategy that doesn't upscale.

sjudd:
Thanks!

Issue #738, opened 2015-11-12
Some images simply do not download (Black image) - Android


thiagoluis88:
Hi,
In some devices (Moto X) do not download correctly, like the screenshot attached.
When i skipMemoryCache(true) it works, because Glide download the image again.
I already use asBitmap() with imageDecoder() and the problem still occur.

But, in my other device (Nexus 4) this problem doesn't occur. Glide works perfectly

I tried to config listener() to log Exceptions but never called onException() method. In other words, no Exceptions occurs. Its weird.

Best,
Thiago Luis.

![screenshot_2015-11-12-16-45-43](https://cloud.githubusercontent.com/assets/5911164/11127862/8dd736e6-895d-11e5-9522-ac0da09c6243.png)

TWiStErRob:
Hi there, thanks for reporting. Please follow the instructions and answer the questions below to help us diagnose further. (Alternatively you can provide GitHub mini project that reproduces this issue).

Try enabling [internal logging](https://github.com/bumptech/glide/wiki/Debugging-and-Error-Handling) to see what's going on in the logs.

When you say no `listener.onException` is being called, is `onResourceReady` called? Try to debug `onResourceReady` to see if the `Bitmap resource` value passed is a valid bitmap (AS/IDEA has "View Bitmap" in debugger or I think you can dump it to SD card via a oneliner: `resource.compress(PNG, 0, new FileOutputStream("/sdcard/temp.png"))`).

What does your `imageDecoder` do? Does it work correctly without that?

What are the installed Android versions for each of these devices?

Since you say it works on one phone, but not on an other, I rule out the possiblity that the view is invisible when you try to load, which may lead to no actual load happening and no callbacks called.

What happens if (instead of `skipMemoryCache`) you turn it off via a [`GlideModule`](https://github.com/bumptech/glide/wiki/Configuration)? (I expect it works)
```java
@Override public void applyOptions(Context context, GlideBuilder builder) {
    builder.setMemoryCache(new MemoryCacheAdapter());
}
```

thiagoluis88:
Hi,
I enabled the logging e nothing weird appear. Just logs about downloading from URL.

Yes, the onResourceReady was called. I saw the Bitmap resource and it was a valid bitmap, like the others that were downloaded. Same size, etc.

My imageDecoder was:

    public static StreamBitmapDecoder getBitmapDecoder(Context context){
        BitmapPool bitmapPool = Glide.get(context).getBitmapPool();
[...]

PR #758, opened 2015-11-20
fix for issue #738 Some images simply do not download (Black image) -…

mullender:
… Android

### DIFF
--- a/library/src/main/java/com/bumptech/glide/load/resource/bitmap/Downsampler.java
+++ b/library/src/main/java/com/bumptech/glide/load/resource/bitmap/Downsampler.java
@@ -400,17 +400,19 @@ public final class Downsampler {
     int sourceHeight = options.outHeight;
     String outMimeType = options.outMimeType;
     final Bitmap result;
+    TransformationUtils.getBitmapDrawableLock().lock();
     try {
       result = BitmapFactory.decodeStream(is, null, options);
     } catch (IllegalArgumentException e) {
       throw newIoExceptionForInBitmapAssertion(e, sourceWidth, sourceHeight, outMimeType, options);
+    } finally {
+      TransformationUtils.getBitmapDrawableLock().unlock();
     }
 
     if (options.inJustDecodeBounds) {
       is.reset();
 
     }
-
     return result;
   }
 
--- a/library/src/main/java/com/bumptech/glide/load/resource/bitmap/TransformationUtils.java
+++ b/library/src/main/java/com/bumptech/glide/load/resource/bitmap/TransformationUtils.java
@@ -20,6 +20,11 @@ import android.util.Log;
 import com.bumptech.glide.load.engine.bitmap_recycle.BitmapPool;
 import com.bumptech.glide.util.Preconditions;
 
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
 /**
  * A class with methods to efficiently resize Bitmaps.
  */
@@ -30,6 +35,17 @@ public final class TransformationUtils {
   private static final int CIRCLE_CROP_PAINT_FLAGS = PAINT_FLAGS | Paint.ANTI_ALIAS_FLAG;
   private static final Paint CIRCLE_CROP_SHAPE_PAINT = new Paint(CIRCLE_CROP_PAINT_FLAGS);
   private static final Paint CIRCLE_CROP_BITMAP_PAINT;
+  /**
+   * https://github.com/bumptech/glide/issues/738 On some devices (Moto X with android 5.1) bitmap
+   * drawing is not thread safe.
+   * This lock only locks for these specific devices. For other types of devices the lock is always
+   * available and therefore does not impact performance
+   */
+  private static final Lock BITMAP_DRAWABLE_LOCK = "XT1097".equals(Build.MODEL)
+      && Build.VERSION.SDK_INT == Build.VERSION_CODES.LOLLIPOP_MR1
+      ? new ReentrantLock()
+      : new NoLock();
+
   static {
     CIRCLE_CROP_BITMAP_PAINT = new Paint(CIRCLE_CROP_PAINT_FLAGS);
     CIRCLE_CROP_BITMAP_PAINT.setXfermode(new PorterDuffXfermode(PorterDuff.Mode.SRC_IN));
@@ -37,6 +53,11 @@ public final class TransformationUtils {
 
   private TransformationUtils() {
     // Utility class.
+  }
+
+
+  public static Lock getBitmapDrawableLock() {
+    return BITMAP_DRAWABLE_LOCK;
   }
 
   /**
@@ -74,9 +95,7 @@ public final class TransformationUtils {
     // We don't add or remove alpha, so keep the alpha setting of the Bitmap we were given.
     TransformationUtils.setAlpha(inBitmap, result);
 
-    Canvas canvas = new Canvas(result);
-    canvas.drawBitmap(inBitmap, m, DEFAULT_PAINT);
-    clear(canvas);
+    applyMatrix(inBitmap, result, m);
     return result;
   }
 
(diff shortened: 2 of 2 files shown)


mullender:
fix for #738

library/src/main/java/com/bumptech/glide/load/resource/bitmap/Downsampler.java
```diff
@@ -248,17 +248,6 @@ static void calculateScaling(DownsampleStrategy downsampleStrategy, int degreesT
     float adjustedScaleFactor = powerOfTwoSampleSize * exactScaleFactor;
 
     options.inSampleSize = powerOfTwoSampleSize;
-    // Density scaling is only supported if inBitmap is null prior to KitKat. Avoid setting
-    // densities here so we calculate the final Bitmap size correctly.
-    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
```
mullender: This is the first part of the fix, it seems that the values here are part of the problem

library/src/main/java/com/bumptech/glide/load/resource/bitmap/TransformationUtils.java
```diff
@@ -74,9 +76,11 @@ public static Bitmap centerCrop(@NonNull BitmapPool pool, @NonNull Bitmap inBitm
     // We don't add or remove alpha, so keep the alpha setting of the Bitmap we were given.
     TransformationUtils.setAlpha(inBitmap, result);
 
-    Canvas canvas = new Canvas(result);
```
mullender: This is the second part, it turns out that the canvas.drawBitmap is not thread safe on Lollipop on Motorola

TWiStErRob:
That looks like quite a significant change, I'll let @sjudd handle this one.

My view is that one it's not worth removing all that logic because it doesn't work on one or two phones out of thousands, I'd rather create an exception for those with some device detection.

library/src/main/java/com/bumptech/glide/load/resource/bitmap/TransformationUtils.java
sjudd: Can you provide documentation? Or is this reproducible on a device or device(s) in a sample app?

library/src/main/java/com/bumptech/glide/load/resource/bitmap/Downsampler.java
sjudd: You will need to be more specific. What does this fix? On what device(s)?

This is a substantial decrease in performance. If it fixes specific issues on specific devices, only disable it in those particular cases.

mullender: Agreed.
I was making this PR late last night, usually not the best time :)
Afterwards I was thinking about how this could be related to the issue. My current thought is that there is a bug in the OS code of the MotoX 5.1: bitmap scaling is not thread safe.
When you specify the inDensity and inTargetDensity it will also trigger the same non-thread-safe code path.
I will validate this hunch by applying the same lock on this operation as is applied to the drawBitmap operation.

library/src/main/java/com/bumptech/glide/load/resource/bitmap/TransformationUtils.java
mullender: This reproduces consistently in the Gallery sample app. Simple scrolling sideways and back again will result in multiple black images.

mullender:
@TWiStErRob @sjudd 
I agree with both your comments and made the patch work only for the specific device that consistently has the issue.
I measured the perf impact of adding the lock, it is significant. https://github.com/bumptech/glide/issues/738#issuecomment-158496469

But I think the perf impact is a good trade-off for this device, weighed against not being able to see the images at all.

library/src/main/java/com/bumptech/glide/load/resource/bitmap/BitmapDrawLock.java
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * https://github.com/bumptech/glide/issues/738 On some devices (Moto X with android 5.1) bitmap
+ * drawing is not thread safe.
+ *
+ * This class provides a lock implementation that only really locks for these devices. For other
+ * types of devices the lock is always available and therefore does not impact performance The
+ * actual lock logic is delegated to {@see ReentrantLock}
+ */
+final class BitmapDrawLock implements Lock {
```
TWiStErRob: How about?
```java
final class BitmapDrawLock {
	// only enable the lock for Moto X 2nd gen on android 5.1
	private static final Lock INSTANCE =
			"XT1097".equals(Build.MODEL) && Build.VERSION.SDK_INT == Build.VERSION_CODES.LOLLIPOP_MR1
			? new ReentrantLock()
			: new NoLock();

	private BitmapDrawLock() {
		// static utility class
	}

	public static Lock getInstance() {
		return INSTANCE;
	}
}

public class NoLock implements Lock {
	@Override public void lock() {
		// do nothing
	}
	@Override public void lockInterruptibly() throws InterruptedException {
		// do nothing
	}
	@Override public boolean tryLock() {
		return true;
	}
	@Override public boolean tryLock(long time, @NotNull TimeUnit unit) throws InterruptedException {
		return true;
	}
	@Override public void unlock() {
		// do nothing
	}
	@NotNull @Override public Condition newCondition() {
		throw new UnsupportedOperationException("Should not be called");
	}
}
```
That single INSTANCE field + getInstance can be even moved to some other class, like `TransformationUtils`.

library/src/main/java/com/bumptech/glide/load/resource/bitmap/Downsampler.java
```diff
@@ -40,8 +40,8 @@
       "com.bumptech.glide.load.resource.bitmap.Downsampler.DecodeFormat", DecodeFormat.DEFAULT);
   /**
    * Indicates the {@link com.bumptech.glide.load.resource.bitmap.DownsampleStrategy} option that
-   * will be used to calculate the sample size to use to downsample an image given the original
-   * and target dimensions of the image.
+   * will be used to calculate the sample size to use to downsample an image given the original and
```
TWiStErRob: Can you please revert irrelevant changes like comment/code formatting? It's harder to see what happened with all this diff. You could also squash your PR branch.

mullender: will do!

library/src/main/java/com/bumptech/glide/load/resource/bitmap/BitmapDrawLock.java
mullender: I think your proposal fits better with the existing code in the project

mullender:
Cleaned up the PR and implemented according to your suggestion @TWiStErRob

library/src/main/java/com/bumptech/glide/load/resource/bitmap/TransformationUtils.java
```diff
@@ -335,10 +348,15 @@ public static Bitmap roundedCorners(@NonNull BitmapPool pool, @NonNull Bitmap in
     paint.setAntiAlias(true);
     paint.setShader(shader);
     RectF rect = new RectF(0, 0, result.getWidth(), result.getHeight());
-    Canvas canvas = new Canvas(result);
-    canvas.drawColor(Color.TRANSPARENT, PorterDuff.Mode.CLEAR);
-    canvas.drawRoundRect(rect, roundingRadius, roundingRadius, paint);
-    clear(canvas);
+    TransformationUtils.getBitmapDrawableLock().lock();
```
TWiStErRob: Just nitpick: direct access vs get instance in the same class. I would prefer direct access where it's defined, like in applyMatrix, and getter outside the class.

library/src/main/java/com/bumptech/glide/load/resource/bitmap/TransformationUtils.java
```diff
@@ -30,6 +35,11 @@
   private static final int CIRCLE_CROP_PAINT_FLAGS = PAINT_FLAGS | Paint.ANTI_ALIAS_FLAG;
   private static final Paint CIRCLE_CROP_SHAPE_PAINT = new Paint(CIRCLE_CROP_PAINT_FLAGS);
   private static final Paint CIRCLE_CROP_BITMAP_PAINT;
+  private static final Lock BITMAP_DRAWABLE_LOCK = "XT1097".equals(Build.MODEL)
```
TWiStErRob: That comment on BitmapDrawableLock class was nice, it got lost in the process :(

TWiStErRob:
Thank you, man!

mullender:
@TWiStErRob good nitpicks ;) 
I updated the PR.

sjudd:
LGTM, thanks for tracking this down and fixing it!

PR #762, opened 2015-11-22
Clean upload script up

vanniktech:


scripts/upload.gradle
```diff
@@ -128,24 +130,20 @@ afterEvaluate { project ->
         }
 
         task androidJavadocsJar(type: Jar, dependsOn: androidJavadocs) {
+            classifier "javadoc"
             from androidJavadocs.destinationDir
             baseName "${JAR_PREFIX}${project.name}${JAR_POSTFIX}"
         }
 
         task androidSourcesJar(type: Jar) {
+            classifier = 'sources'
```
TWiStErRob: Nit: to `=` or not to `=`? see line 133.

vanniktech: classifier should have the =

I'll change that

TWiStErRob: nope, I don't see

vanniktech: Okay cool. Fixed, squashed & pushed.

TWiStErRob:
Thanks, the changes look reasonable, but I don't know the `mavenDeployer` API. @sjudd will take a look too.

scripts/upload.gradle
```diff
@@ -128,24 +130,20 @@ afterEvaluate { project ->
...
 
-    artifacts {
-        if (project.getPlugins().hasPlugin('android') || project.getPlugins().hasPlugin('android-library')) {
-            archives androidSourcesJar {
-                classifier "sources"
-            }
-            archives androidJavadocsJar {
-                classifier "javadoc"
-            }
+        artifacts {
```
sjudd: Does the indentation change exclude uploading artifacts for non-android projects? The dislrucache project is java only and needs to continue to be uploaded. I'm clearly not a Gradle pro. I'd believe this doesn't change the behavior

scripts/upload.gradle
sjudd: Any particular reason for 'sources' here and "javadoc" above?

sjudd:
Maybe I'm missing something, but I'm not sure I see where this uploads an aar?

scripts/upload.gradle
TWiStErRob: I think `artifacts` is a configuration block and each `archives` call is kind of an `add` method call. So unless the mere presence of the archives block has a side effect, I think they're equivalent.

TWiStErRob:
I was wonderimg that too, I think it's just a loosely related code cleanup and there's more to come in the other one. That's why it's a separate PR.

scripts/upload.gradle
vanniktech: The indentation change is because I merged the two same if statements into one (checking whether the project has the android application or library plugin applied).

@TWiStErRob yup that's right. You can also see that [here](https://github.com/chrisbanes/gradle-mvn-push/blob/master/gradle-mvn-push.gradle)

vanniktech:
It's just a cleanup. I chose to have it in a separate one, to now blow up one PR.

Each of the subprojects define whether they want to be uploaded as a JAR or an ARR in <subproject>/gradle.properties

scripts/upload.gradle
TWiStErRob: He means the quotes difference, in case you didn't see; it took me a few seconds to notice.

vanniktech: If you only have plain text you should use ' otherwise you can insert variables with "

'Hello World'
"Hello ${varName}"

I'll fix it.

vanniktech:
I changed all the " to ' where possible in `scripts/upload.gradle` to make it clear. Again squashed everything into one commit.

scripts/upload.gradle
```diff
@@ -102,13 +102,15 @@ afterEvaluate { project ->
...
     signing {
-        required { isReleaseBuild() && gradle.taskGraph.hasTask("uploadArchives") }
+        required { isReleaseBuild() && gradle.taskGraph.hasTask('uploadArchives') }
         sign configurations.archives
     }
 
-    if (project.getPlugins().hasPlugin('android') || project.getPlugins().hasPlugin('android-library')) {
-        def releaseVariants = project.android.libraryVariants.findAll { variant ->
-            variant.buildType.name.equalsIgnoreCase("release")
+    def isAndroidProject = project.plugins.hasPlugin('com.android.application') || project.plugins.hasPlugin('com.android.library')
```
TWiStErRob: I just noticed that you changed `android` to `com.android. application`, I guess we should have a new enough (1.2.3) android gradle plugin, right? This change of names was around 0.14 iirc.

vanniktech: You are already applying `com.android.application` and `com.android.library` everywhere in your `build.gradle` files. No need for upgrading the Android Gradle Plugin.

sjudd:
LGTM, thanks!

vanniktech:
Awesome. You're welcome. I'll within the next days add more and more off those small Gradle cleanups as I've seen some other places where Gradle code could be improved.

sjudd:
Great thanks, I look forward to it.

PR #771, opened 2015-11-25
Update to Gradle 2.9 | Checkstyle 6.12.1 | PMD 5.4.0

vanniktech:
In order to update Checkstyle I had to upgrade Gradle since with Checkstyle 6.8 there was a breaking change and Gradle fixed that with 2.7. [More information](https://github.com/gradle/gradle/pull/491)

The CI will fail since there are 5 checkstyle issues:

```
[ant:checkstyle] /Users/nbaudy/tmp/glide/library/src/main/java/com/bumptech/glide/load/Option.java:146: error: Type Javadoc comment is missing an @param <T> tag.
[ant:checkstyle] /Users/nbaudy/tmp/glide/library/src/main/java/com/bumptech/glide/load/model/UnitModelLoader.java:68: error: Type Javadoc comment is missing an @param <ResourceType> tag.
[ant:checkstyle] /Users/nbaudy/tmp/glide/library/src/main/java/com/bumptech/glide/request/transition/NoTransition.java:19: error: Type Javadoc comment is missing an @param <R> tag.
[ant:checkstyle] /Users/nbaudy/tmp/glide/library/src/main/java/com/bumptech/glide/util/pool/FactoryPools.java:109: error: Type Javadoc comment is missing an @param <T> tag.
[ant:checkstyle] /Users/nbaudy/tmp/glide/library/src/main/java/com/bumptech/glide/util/pool/FactoryPools.java:116: error: Type Javadoc comment is missing an @param <T> tag.
```

Should I fix them or would you do that?


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Wed May 20 08:28:58 PDT 2015
+#Wed Nov 25 12:31:24 CET 2015
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.2.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-2.9-bin.zip
```
TWiStErRob: I think this should stay `-all`, is there a `wrapper` task switch for this?

vanniktech: @TWiStErRob yup there is. I updated it. Have a look at the root build.gradle file

TWiStErRob:
I think those errors should be fixed, it's not good to have a failing commit on any public branch.

Thanks for doing this. Just curious: what's the motivation behind it?

gradle/wrapper/gradle-wrapper.properties
TWiStErRob: Oh, I was hoping for a less hardcoded one, but I guess it's ok.

sjudd:
I'll fix those errors, thanks for bringing them up.

vanniktech:
@TWiStErRob The motiviation is to give back to this library & get #761 sooner or later merged ; and those little changes are just to clean the Gradle files up and so that I get to know the Gradle structure of that project
@sjudd thanks for that.

I rebased to master. The checkstyle issues should be resolved now.

sjudd:
The checkstyle issues are resolved, but there are PMD issues, most of which appear to be incorrect. The unused methods are actually used, and the missing break statements I've looked at are empty and shouldn't be creating warnings according to PMD's documentation...

vanniktech:
It could be that the default PMD version has been changed in between 2.2.1 and 2.9 of Gradle.

@sjudd if you want me, I can have a look at the errors

sjudd:
Sure please do. You may also need to manually set the PMD version to avoid the false positives?

vanniktech:
Yes I'll set it to the latest version. Do you mind if I do that in a separate PR? Just to avoid having too many things in here.

sjudd:
That's ok if it's not helpful in resolving the errors.

vanniktech:
I fixed some PMD errors (simple ones), since I don't know how you want to handle those errors:

> Avoid unnecessary if..then..else statements when returning booleans
Consider using varargs for methods or constructors which take an array the last parameter.
Overridable method '<method>' called during object construction

We have 2 options:
Fixing them or ignoring the ruleset.

TWiStErRob:
I think all of those are useful checks, I took a chance at fixing/ignoring them in https://github.com/TWiStErRob/glide/commit/cac7132a12c3276feaedbd73058436256a090451 feel free to cherry-pick or just copy them over.

Can you please try to fix this warning? I think it was introduced by upgrading Gradle wrapper:
> Build file 'P:\projects\contrib\github-glide\glide\build.gradle': line 60
> The plus(Iterable<FileCollection>) method and using the '+' operator in conjunction with an Iterable<FileCollection> object have been deprecated and are scheduled to be removed in 3.0.  Please use the plus(FileCollection) method or the '+' operator with a FileCollection object instead.

vanniktech:
@TWiStErRob thanks for the PMD warnings. To be honest I don't know how to fix that Gradle warning. Sorry for that. And yes it was introduced with Updating the Gradle version.

I cherry picked your branch into this PR, squashed all of the commits into one and pushed it. Now we should get a stable CI other than the Gradle warning.

TWiStErRob:
Huh, did you rebase before squash? It's weird to see that the commit was made by sjudd and it contains his changes as well (generic comments). It's not an issue, I'm just trying to understand what's happening.

vanniktech:
I added your repo as a remote, then cherry picked your branch. After that I was one commit ahead. So I executed `git rebase -i HEAD~4` replace the pick by squash for 3 commits (can't remember which one) and then push forced.

Dunno if this is the right way or not. So far it has always worked. Other than now :/ 

Probably I chose the wrong pick commit.

sjudd:
Rebased, fixed the commit authorship, and merged here: 2a4e75a21fea39e1c069e86ed197e3e076667d43

Thanks!

PR #772, opened 2015-11-26
Reorder AppWidgetTarget arguments to help Java disambiguate varargs

TWiStErRob:
When I tried to use AppWidgetTarget I ran into a problem with varargs, there are too many `int` arguments next to each other which confused Java. Consider:
```java
AppWidgetTarget(Context, RemoteViews, int, int, int, int...);
AppWidgetTarget(Context, RemoteViews, int, int...);
```
which one would you call for `AppWidgetTarget(context, rv, R.id.image, 400, 400, appWidgetId)`?

To prevent this ambiguity from coming up I reordered the arguments to be clear where the varargs start by moving `width/height/viewId` before `RemoteViews`. This is a change that requires a major version, so I think it's a good candidate for 4.0.

To keep it consistent with other `RemoteViews` targets, I also made the same change to `NotificationTarget`, however the can be no ambiguity there.


sjudd:
Thanks!

PR #798, opened 2015-12-09
Do not re-allocate an entire MessageDigest instance for every key.

ened:
Right now, Glides `SafeKeyGenerator` creates a new `MessageDigest` instance (via `getInstance`) every time a new key needs to be generated, causing significant overhead as shown in systrace.
screen shot 2015-12-09 at 13 02 28

It should follow http://stackoverflow.com/a/13802730/375209 and use a pre-allocated MessageDigest.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/load/engine/cache/SafeKeyGenerator.java
+++ b/library/src/main/java/com/bumptech/glide/load/engine/cache/SafeKeyGenerator.java
@@ -14,14 +14,21 @@ import java.security.NoSuchAlgorithmException;
 public class SafeKeyGenerator {
   private final LruCache<Key, String> loadIdToSafeHash = new LruCache<>(1000);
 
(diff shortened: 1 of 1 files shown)


TWiStErRob:
I like the idea, except that `synchronized` keyword. Multiple background threads will do this at once that's for sure. I would rather you use a pool as everywhere else:
https://github.com/bumptech/glide/blob/master/library/src/main/java/com/bumptech/glide/util/pool/FactoryPools.java#L96
I think that method I pointed should be public (as `Resetter` interface is public already).

This is just my opinion, @sjudd has the final word.

sjudd:
Rebased and merged here: https://github.com/bumptech/glide/commit/8489cc8edad9aa43fb512cfa40abd618fe074ce1

Thanks!

ened:
@sjudd cool. Thanks for your help.

PR #854, opened 2015-12-31
fix bug of DiskCacheStrategy.RESOURCE

luxiaoyu:
This bug will cache source data at DiskCacheStrategy.RESOURCE.


luxiaoyu:
@googlebot I signed it!

sjudd:
Nice catch, thanks!

PR #855, opened 2015-12-31
Add filter to mavenDeployer for glide package to also distribute it with aar

vanniktech:
First try at this. Follow up / requirement because of #761 

I'm not quite sure whether this is the right way to go or not since I've never tried any of this before though according to the [Gradle documentation](https://docs.gradle.org/current/userguide/maven_plugin.html#sub:multiple_artifacts_per_project) this is the way. At least if I didn't misread it.

### DIFF
--- a/scripts/upload.gradle
+++ b/scripts/upload.gradle
@@ -97,6 +97,12 @@ afterEvaluate { project ->
                         }
                     }
                 }
+
+                addFilter('aar_upload') { artifact, file ->
+                    artifact.name == 'Glide'
+                }
+
+                pom('aar_upload').packaging = 'aar'
             }
         }
     }


TWiStErRob:
Few things I tried and noticed:

---

That sounds too good to be true, did you actually try it? I tried to run it like this:
```bash
cls && gradlew clean :glide:buildArchives :glide:uploadArchives -x check -PSNAPSHOT_REPOSITORY_URL=file://p:\projects\contrib\github-glide\repo-snapshot -PRELEASE_REPOSITORY_URL=file://p:\projects\contrib\github-glide\repo-release
```
Without this change the jars published fine, but when I added it, nothing was published.

---

I played around a little and found that `glide/build.gradle` needs this:
```gradle
artifacts {
    archives project(':library').tasks.bundleRelease
}
```
to have the `aar` file uploaded, because `:library` is already an `aar`-packaged module.

---

I tried to make your filter solution work, but it just doesn't want to play nice:
> If at least one custom filter has been added the default POM is not used any longer. -- [docs](https://docs.gradle.org/current/javadoc/org/gradle/api/artifacts/maven/PomFilterContainer.html#setPom(org.gradle.api.artifacts.maven.MavenPom))

This means that if we start using filters the pom.xml will be much leaner than the current, I guess it can be changed with some work.

---

I don't think it's a good idea to put this filter into `scripts/upload.gradle` because it is used by `:integration:*` as well, the same block can be added to `glide/build.gradle` though like so:
```gradle
apply from: "$rootProject.projectDir/scripts/upload.gradle"
uploadArchives {
    repositories {
        mavenDeployer {
                addFilter...
        }
    }
}
```

---

There's a problem I noticed: `classes.jar` inside the `aar` doesn't contain all files that are in the `jar`-packaged version, namely the `gifdecoder` and `disklrucache` packages are missing from the `aar`. Well, I guess that why the whole `:glide` module exists, to package those. I guess this is where the small libraries Sam mentioned in #761 come in.

---

So with the `archives ....bundleRelease` trick we can have the .aar side by side with .jar, the bigger problem is that these two share the pom file and there's a conflict there. The .jar has those two other libs packaged inside, while the aar would have to have some `<dependency>`s defined in the `pom.xml`. The artifactId would have to change to make this work as I see it now. So it would be:
```gradle
// to consume as jar (without transitive deps):
compile 'com.github.bumptech.glide:glide:4.0.0'
// to consume as aar (with transitive deps):
compile 'com.github.bumptech.glide:library:4.0.0@aar'
```
I think we could keep the `glide` artifactId if we got rid of the fat jar `:glide` module's `publishArtifacts` part (that is don't publish `:glide` into Maven), but instead had 3 jars: the module `:library` published under `artifactId=glide` and those small transtiive dependencies. This would be the same as what `:integration:*` has as they're published as both jar and aar. That way gradle users can use `glide:4.0.0` or `glide:4.0.0@aar` and non-gradlers can just download the fat jar from the GitHub releases page. @sjudd did I understand what you were saying in #761 just now?!

vanniktech:
I just quickly tested it but not in depth though. Sorry for not stating that.

I also tried basically having another script in `scripts/` like `aar-upload.gradle` though having two instances in one project does also not play nicely.

About the `classes.jar` not containing every class I think that's when the `transitive = true` in Gradle comes in handy as it will resolve those somehow. Though I'm not quite sure about that. It's just a wild guess.

Yup then there's also the problem with the artifactId as we can't have the same for JAR & AAR.

The more I think about it the easier it would be if with 4.0.0 the support for JAR would be cut off. I honestly don't know anyone who's not using Gradle yet. The only people that might still use Maven because are C++ guys because of the feature lack in AS (at least last time I checked). Then again I don't know why those C++ projects would use a Java library.

TWiStErRob:
I think my last paragraph above details the best solution (keeping compatibility as well), the problem is that we're using the fat jar in Gradle as well, if we got rid of that, then we can publish jar and aar under the same artifactId (with proper dependencies), and separately just build a fat jar for manual upload.

TWiStErRob:
As for non-gradle users: I think [all of these](http://stackoverflow.com/search?q=%5Bandroid%5D+%5Beclipse%5D+lastactive%3A3m..) are hiding potential users of the Glide fat jar.

vanniktech:
I'm totally confused now. What should we change in order to have both AAR + JAR support?

TWiStErRob:
It's a lot, see #874.

TWiStErRob:
@vanniktech #874 is merged, you can continue with #761.

vanniktech:
@TWiStErRob As far as I see this, #761 can be merged or am I missing something?

TWiStErRob:
Hmm, true that :)

PR #856, opened 2016-01-02
Upgrade to OkHttp 3.

swankjesse:


TWiStErRob:
That looks like an easy upgrade, why is it never that simple? I think this should be a separate integration library called `okhttp3`, especially since OkHttp 3 is not even released yet.

Thanks to their package change it's no longer compatible... so if anyone is stuck with OkHttp 2 due to weird company policies or other reasons they wouldn't be able to use Glide 4 without writing their own integration.

TWiStErRob:
Build failed with:

> Could not find `com.squareup.okhttp3:mockwebserver:2.3` in module `library`.

integration/okhttp3/src/main/java/com/bumptech/glide/integration/okhttp3/OkHttpGlideModule.java
```diff
@@ -0,0 +1,31 @@
+package com.bumptech.glide.integration.okhttp3;
```
swankjesse: Note the package!

swankjesse:
Fixed the build break. Didn't see that when I built locally, though it's probably a bad configuration on my end. Should be good now.

The updated PR adds a new OkHttp 3.x integration and deprecates the old 2.x integration. That gives everyone something that'll work.

integration/okhttp3/src/main/java/com/bumptech/glide/integration/okhttp3/OkHttpStreamFetcher.java
```diff
@@ -0,0 +1,95 @@
...
+ * Fetches an {@link InputStream} using the okhttp library.
+ */
+public class OkHttpStreamFetcher implements DataFetcher<InputStream> {
+  private static final String TAG = "OkHttpFetcher";
+  private final OkHttpClient client;
+  private final GlideUrl url;
+  private InputStream stream;
+  private ResponseBody responseBody;
+
+  public OkHttpStreamFetcher(OkHttpClient client, GlideUrl url) {
```
JakeWharton: Take a `Call.Factory` instead?

swankjesse: Good call. Fixed throughout.

integration/okhttp3/src/main/java/com/bumptech/glide/integration/okhttp3/OkHttpStreamFetcher.java
```diff
@@ -0,0 +1,96 @@
...
+      // Ignored
+    }
+    if (responseBody != null) {
+      responseBody.close();
+    }
+  }
+
+  @Override
+  public void cancel() {
+    // TODO: call cancel on the client when this method is called on a background thread. See #257
```
TWiStErRob: @swankjesse didn't you say this was safe now?

integration/okhttp3/src/main/java/com/bumptech/glide/integration/okhttp3/OkHttpUrlLoader.java
```diff
@@ -0,0 +1,83 @@
...
+import android.content.Context;
+
+import com.bumptech.glide.load.Options;
+import com.bumptech.glide.load.model.GlideUrl;
+import com.bumptech.glide.load.model.ModelLoader;
+import com.bumptech.glide.load.model.ModelLoaderFactory;
+import com.bumptech.glide.load.model.MultiModelLoaderFactory;
+
+import java.io.InputStream;
+
```
TWiStErRob: Build failed with:
> integration/okhttp3/OkHttpUrlLoader.java:13: error: Wrong order for 'okhttp3.Call' import.

Try `gradlew clean check jar assemble` before commit, it may be a little overkill, but it should check most things if they broke.

TWiStErRob:
Nice job! (Wow, you're really pushing OkHttp 3 :)
See my two comments and then wait for @sjudd to merge.

integration/okhttp3/src/main/java/com/bumptech/glide/integration/okhttp3/OkHttpStreamFetcher.java
swankjesse: Yeah, I’ll fix that in a separate, follow up PR.

sjudd:
LGTM, thanks all!

PR #874, opened 2016-01-07
Introduce AAR publish for :library and stop publishing :glide

TWiStErRob:
Replacing PR #855.

In this PR (from commit message):
* Don't publish `:glide` as anything to Maven repos
* Build `:glide` as 'glide-full' to have a flat jar to be published on GitHub/releases
* Publish `:library` as 'group:glide' as AAR and JAR to Maven repos
* Make `:third_party:gif_decoder` and `:third_party:disklrucache` first-class published gradle dependencies
* Remove duplicated `repositories`, `VERSION_*`, `JAR_PREFIX`, `JAR_POSTFIX` declarations  
(pull constants up to `:integration` and `:third_party` level)
* Fix JavaDoc generation (Error reading file: `${androidSdkDirectory}\docs\reference\package-list`)  
* Move JAR generation to `upload.gradle` (from `integration/build.gradle`) so it's used by `:library` as well
* Add javadoc and sources JAR generation to upload.gradle for non-android `projects as well

This PR depends on sjudd/disklrucache#3

Before commit I tried with:

```
gradlew clean build assemble check buildArchives uploadArchives \
-PSNAPSHOT_REPOSITORY_URL=file://p:\projects\contrib\github-glide\repo-snapshot \
-PRELEASE_REPOSITORY_URL=file://p:\projects\contrib\github-glide\repo-release
```

and it compiled fine.

I got the following artifacts in repo-snapshot local Maven repository:
```java
com.github.bumptech.glide:glide:4.0.0-SNAPSHOT + aar (from :library)
com.github.bumptech.glide:disklrucache:1.0.0-SNAPSHOT (from :third_party)
com.github.bumptech.glide:gifdecoder:1.0.0-SNAPSHOT + aar (from :third_party)

com.github.bumptech.glide:volley-integration:2.0.0-SNAPSHOT + aar
com.github.bumptech.glide:okhttp-integration:2.0.0-SNAPSHOT + aar
com.github.bumptech.glide:okhttp3-integration:2.0.0-SNAPSHOT + aar
com.github.bumptech.glide:recyclerview-integration:2.0.0-SNAPSHOT + aar
com.github.bumptech.glide:gifencoder-integration:2.0.0-SNAPSHOT + aar
```

and `gradlew :glide:jar` gave the following artifacts:
```gradle
glide/build/libs/glide-full-4.0.0-SNAPSHOT.jar
```

Everything above includes the `-sources` and `-javadoc` classified JARs as well.

### DIFF
--- a/integration/okhttp3/gradle.properties
+++ b/integration/okhttp3/gradle.properties
@@ -9,7 +9,3 @@ VERSION_PATCH=0
 VERSION_CODE=8
 
 POM_DESCRIPTION=An integration library to use OkHttp 3.x to fetch data over http/https in Glide
-
-# Prefix and postfix for source and javadoc jars.
-JAR_PREFIX=glide-
-JAR_POSTFIX=-integration
--- a/third_party/gif_decoder/gradle.properties
+++ b/third_party/gif_decoder/gradle.properties
@@ -1,0 +1,11 @@
+POM_NAME=Glide Gif Decoder Library
+POM_ARTIFACT_ID=gifdecoder
+POM_PACKAGING=aar
+
+VERSION_NAME=1.0.0-SNAPSHOT
+VERSION_MAJOR=1
+VERSION_MINOR=0
+VERSION_PATCH=0
+VERSION_CODE=1
+
+POM_DESCRIPTION=Implementation of GifDecoder that is more memory efficient to animate for Android devices.
--- a/glide/build.gradle
+++ b/glide/build.gradle
@@ -44,6 +44,8 @@ def getAndroidJar() {
     "${getAndroidSdkDirectory()}/platforms/${getAndroidCompileSdkVersion()}/android.jar"
 }
 
+project.archivesBaseName = "${POM_ARTIFACT_ID}-${VERSION_NAME}"
+
 // Generate javadocs and sources containing batched documentation and sources for all internal projects.
 ['release', 'debug'].each { variantName ->
 
@@ -102,5 +104,3 @@ artifacts {
         classifier 'sources'
     }
 }
-
-apply from: "$rootProject.projectDir/scripts/upload.gradle"
--- a/glide/gradle.properties
+++ b/glide/gradle.properties
@@ -1,4 +1,3 @@
-POM_NAME=Glide
-POM_ARTIFACT_ID=glide
+POM_NAME=Glide Full
+POM_ARTIFACT_ID=glide-full
 POM_PACKAGING=jar
-
--- a/integration/build.gradle
+++ b/integration/build.gradle
@@ -1,15 +1,1 @@
-evaluationDependsOnChildren()
-
-import com.android.build.gradle.api.BaseVariant
-subprojects {
-    android.libraryVariants.all { BaseVariant variant ->
-        def jarTask = project.tasks.create(name: "jar${variant.name.capitalize()}", type: Jar) {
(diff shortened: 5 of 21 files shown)


TWiStErRob:
I don't think this will build on travis until the corresponding sjudd/disklrucache#3 is merged.

TWiStErRob:
This should also solve the IDEA import issues I had last time I was setting this up as everyone now clearly depends on a normal Android lib `:library` instead of the codeless `:glide`.

integration/okhttp3/gradle.properties
```diff
@@ -2,14 +2,10 @@ POM_NAME=Glide OkHttp 3.x Integration
 POM_ARTIFACT_ID=okhttp3-integration
 POM_PACKAGING=aar
 
-VERSION_NAME=2.0.0-SNAPSHOT
-VERSION_MAJOR=2
+VERSION_NAME=1.0.0-SNAPSHOT
```
sjudd: I've been leaving these versions the same in all of the integrations just so it's easier to keep track of.

sjudd:
Does third_party/gif_encoder need to be published as well for the gifencoder library?

sjudd:
This looks great! Thanks for putting this together, just a few minor comments.

integration/okhttp3/gradle.properties
TWiStErRob: Ok, I'll pull version up to :integration like prefix.

To fully understand: Why isn't the glide version used then? If all integrations change together, they might as well change together with Glide version. Would you ever release an integration minor change without a glide minor change?

TWiStErRob:
@sjudd `third_party/gif_encoder` is not even a Gradle module, it's just a folder. For some reason it's just pulled in as an external source folder from `:integration:gifencoder`, shouldn't it be just copied to `:integration:gifencoder/src/main` to reduce confusion?

third_party/gif_decoder/gradle.properties
```diff
@@ -0,0 +1,11 @@
+POM_NAME=Glide Gif Decoder Library
+POM_ARTIFACT_ID=gifdecoder
+POM_PACKAGING=aar
+
+VERSION_NAME=1.0.0-SNAPSHOT
```
TWiStErRob: @sjudd what about these version numbers?
(including the disklrucache in sjudd/disklrucache#3)

sjudd:
Ah right, thanks for reminding me about gifencoder. The file is in a separate folder to meet Google's standards for open source code (all third party code goes into a folder called third_party). That said, it really ought to be a separate library project, just like the other two third_party projects. It's fine to leave as is for now, that's my mess I can clean up later.

As for versioning, there isn't really any reason why those integration libraries shouldn't just match Glide's. I can imagine a scenario in which someone put up a PR for one of the libraries and I might want to release just the library without releasing Glide and would be unable to do so because of the current versioning scheme, but that hasn't happened yet. I agree the right solution is to either switch them all to match Glide's or keep them all independent. You don't need to do that here either.

TWiStErRob:
Ok, I'll revert versioning changes for now, let's unblock #646.

TWiStErRob:
@sjudd I pulled `VERSION_*` constants up to `:integration`. They can be overridden if needed in the submodules.

TWiStErRob:
Yay, build successful :)

sjudd:
Thanks!

Issue #257, opened 2014-11-11
NetworkOnMainThreadException when using Glide with OkHttp


thorbenprimke:
After upgrading to 3.4, it intermittently hits a NetworkOnMainThreadException. I haven't had time to investigate it deeper except for removing the OkHttp registration line. 

Init code:
<code>
  Glide.get(this).register(GlideUrl.class, InputStream.class, new OkHttpUrlLoader.Factory());
</code>

Calling code:
<code>

    Glide.with(activity)
        .load(uri)
        .centerCrop()
        .into(imageView);
</code>

Exception trace:
<code>

    android.os.NetworkOnMainThreadException
            at android.os.StrictMode$AndroidBlockGuardPolicy.onNetwork(StrictMode.java:1145)
            at com.android.org.conscrypt.OpenSSLSocketImpl.close(OpenSSLSocketImpl.java:1009)
            at com.squareup.okhttp.Connection.closeIfOwnedBy(Connection.java:136)
            at com.squareup.okhttp.OkHttpClient$1.closeIfOwnedBy(OkHttpClient.java:70)
            at com.squareup.okhttp.internal.http.HttpConnection.closeIfOwnedBy(HttpConnection.java:134)
            at com.squareup.okhttp.internal.http.HttpTransport.disconnect(HttpTransport.java:154)
            at com.squareup.okhttp.internal.http.HttpEngine.disconnect(HttpEngine.java:447)
            at com.squareup.okhttp.Call.cancel(Call.java:126)
            at com.squareup.okhttp.Dispatcher.cancel(Dispatcher.java:134)
            at com.squareup.okhttp.OkHttpClient.cancel(OkHttpClient.java:506)
            at com.bumptech.glide.integration.okhttp.OkHttpStreamFetcher.cancel(OkHttpStreamFetcher.java:59)
            at com.bumptech.glide.load.model.ImageVideoModelLoader$ImageVideoFetcher.cancel(ImageVideoModelLoader.java:120)
            at com.bumptech.glide.load.engine.DecodeJob.cancel(DecodeJob.java:132)
            at com.bumptech.glide.load.engine.EngineRunnable.cancel(EngineRunnable.java:46)
            at com.bumptech.glide.load.engine.EngineJob.cancel(EngineJob.java:119)
            at com.bumptech.glide.load.engine.EngineJob.removeCallback(EngineJob.java:93)
            at com.bumptech.glide.load.engine.Engine$LoadStatus.cancel(Engine.java:52)
[...]

PR #887, opened 2016-01-13
Porting okhttp3 integration to Glide 3

frindly:
I ported the okhttp3 integration to the Glide 3 branch.


frindly:
I signed the CLA.

TWiStErRob:
Lgtm.

Reminder: #257 will need a re-fix for this as well.

TWiStErRob:
Oh and I opened #888 because the version looks weird compared to the v4 okhttp3 one. @frindly you needn't worry about this.

frindly:
Thanks -- I couldn't really tell what the version strategy was, so I tried to be consistent with what was in the other integrations.

TWiStErRob:
Yes, you did good. The current idea is that all integration libs have the same version, separate from the main lib's version (this might change with 4.0). I tried to make it more obvious by removing duplication, but it didn't work, resulting in [GRADLE-3380](https://issues.gradle.org/browse/GRADLE-3380).

sjudd:
LGTM, thanks! Would you mind squashing your commits? If not I'll get to it soon.

frindly:
squashed

sjudd:
Awesome thanks!

TWiStErRob:
@frindly you may be interested in #897

frindly:
Of course -- they upgrade off RC one day later.  I'll look at updating it next week.

b95505017:
Glide 3 using sync call with OkHttp without callback, so I think it could upgrade the version directly.

b95505017:
Would you mind publish a new stable release for Glide 3 with OkHttp 3.0 integration?

integration/okhttp3/build.gradle
```diff
@@ -0,0 +1,26 @@
+apply plugin: 'com.android.library'
+
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    compile project(':library')
```
TWiStErRob: I think we missed this, see #941
Library module in v3 is not published, everything there depends on `:glide` as far as I remember.

PR #891, opened 2016-01-14
cleans up Memory leak, because raw data was not released

alosdev:
We had large memory leaks with gifs. After investigating the memory we found the problem in the "raw_data" of the released gif decoder, which was in the pool.


alosdev:
I signed it!

PR #917, opened 2016-01-22
Update OkHttp version from RC to 3.0.1

frindly:
Updating to newest ok http version


frindly:
I signed it!

sjudd:
Thanks, do you mind squashing your commits?

frindly:
squashed

third_party/gif_decoder/src/main/java/com/bumptech/glide/gifdecoder/GifDecoder.java
```diff
@@ -352,6 +352,7 @@ public void clear() {
             bitmapProvider.release(previousImage);
         }
         previousImage = null;
+        rawData = null;
```
sjudd: Merge error maybe?

TWiStErRob: His branch is 1 PR behind, should the merge handle this?
Though it's probably better if @frindly rebases.

sjudd:
Ah yeah, I think this will be fine. I'll try merging and we will see :).

sjudd:
Yup worked, thanks!

TWiStErRob:
Yeah, your merge "commit" is showing the correct diff, but his single commit shows the same diff with the extra line added. So it kind-of worked :) The only weird thing I don't like is the [Blame shows](https://github.com/bumptech/glide/blame/3.0/third_party/gif_decoder/src/main/java/com/bumptech/glide/gifdecoder/GifDecoder.java#L355) his commit for that line and not the original one. So it's worth cleaning up the commits properly before accepting the PR to not confuse the VCS history.

PR #920, opened 2016-01-23
Use Tag along with Id to identify the notification

iontech:
The pair `(tag, id)` uniquely identify a notification.

Refer [The NotificationManager class description](http://developer.android.com/reference/android/app/NotificationManager.html)

### DIFF
--- a/library/src/main/java/com/bumptech/glide/request/target/NotificationTarget.java
+++ b/library/src/main/java/com/bumptech/glide/request/target/NotificationTarget.java
@@ -21,6 +21,7 @@ public class NotificationTarget extends SimpleTarget<Bitmap> {
   private final RemoteViews remoteViews;
   private final Context context;
   private final int notificationId;
+  private final String notificationTag;
   private final Notification notification;
   private final int viewId;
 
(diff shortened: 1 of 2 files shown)


iontech:
I signed it

library/src/main/java/com/bumptech/glide/request/target/NotificationTarget.java
```diff
@@ -29,35 +30,58 @@
...
+   * @param notificationId  The notificationId of the Notification that we want to load the Bitmap.
+   * @param notificationTag The notificationTag of the Notification that we want to load the Bitmap.
+   *                        May be {@code null}.
    */
-  public NotificationTarget(Context context, int width, int height,
-      int viewId, RemoteViews remoteViews, Notification notification, int notificationId) {
+  public NotificationTarget(Context context, int viewId, RemoteViews remoteViews,
+      Notification notification, int notificationId, String notificationTag) {
+    this(context, SIZE_ORIGINAL, SIZE_ORIGINAL, viewId, remoteViews, notification, notificationId,
+      notificationTag);
```
TWiStErRob: I'm a little torn here. The `notify` and `cancel` methods in the Android API break tradition of putting the optional args last, so the order becomes inconsistent with these constructors and even the doc says `(tag, id)`.

I guess we can leave it as is to help understanding the telescoping constructors better. This and the widget target is probably in need of a builder pattern (out of scope for you now).

TWiStErRob:
One comment on changes above and please squash your commits and push to your branch, other than that lgtm.

sjudd:
LGTM. @TWiStErRob I agree we could use a builder here and that it's out of scope. I think it's fine to leave the constructor as id, tag, though I share your feelings :).

TWiStErRob:
@iontech please do a squash on your branch so it has only one commit.

iontech:
@TWiStErRob I apologize, it took awhile. Pushed the squashed commit. Should be good.

PR #922, opened 2016-01-23
3.0

narren:
Need to pull the code to build it locally.


TWiStErRob:
There's no need to merge forward, as far as I know all 3.0 commits made it to master at one point. If you think one's missing please call it out in an issue or a single PR fixing the same thing on master.

If you're having build issues refer to Readme.md's build section and if that didn't help you can get help with your specific build issue on glide-library Google group or Glide's GitHub isssue tracker.

Issue #836, opened 2015-12-23
Some GIFs overdraw on previous frames


Alluretears:
**Glide Version**: 3.6.1
**Device**: Nexus 5/6.0(API 23)
When loading the transparent background gif images, there will be remnant, as shown below.
![device-2015-12-23-130049](https://cloud.githubusercontent.com/assets/7030019/11971353/21872386-a977-11e5-92ff-aa0251fa9061.gif)

**Glide load line**:
```java
Glide.with(this).load("http://7rf2lu.com2.z0.glb.qiniucdn.com/1450421326882.gif").into(imageView);
```

**Layout XML**:
```xml
<ImageView
    android:id="@+id/img_gif_Glide"
    android:layout_width="160dp"
    android:layout_height="160dp"/>
```

TWiStErRob:
I think this is already fixed by #620, but it's not released yet, only available in [`4.0.0-SNAPSHOT`](https://github.com/bumptech/glide/wiki/Snapshots). Can you please try that?

Alluretears:
@TWiStErRob Thank you very much for your help. I tried 4.0.0-SNAPSHOT(glide-4.0.0-20151212.212519-126.jar)but it did not fix，the same  effect as above:(

TWiStErRob:
@kojilin can you please help here?

kojilin:
@TWiStErRob @Alluretears I have tested and found the original gif's second frame's disposal is 1(DISPOSAL_NONE). So basically, I think the behavior is correct, but I see browsers(chrome, firefox, safari) are all clear last frame before playing first frame(Does it means first frame should not care last frame's disposal value?I'm not sure yet;;).

Alluretears:
@TWiStErRob @kojilin Thanks,How can I clear the last frame?

kojilin:
@Alluretears if u don't want to change the code of glide, u need to confirm ur gif has disposal flag at last frame. I'm not very familiar with image editing tool, but gimp gif export can select apply replace option to every frame.

But I think follow browsers behavior is needed. maybe revert #620 and check if previous frame is null, then fill the default color?

TWiStErRob:
@kojilin thank you very much for looking. @sjudd what do you think about his idea?

sjudd:
+1, let's follow the browser behavior.

TWiStErRob:
This is fixed for v4.

PR #950, opened 2016-02-01
Fixes issue #836, always clear previous frame when drawing first frame

kojilin:
Revert #620 and always clear previous frame when drawing first one. See https://github.com/bumptech/glide/issues/836.


TWiStErRob:
Thank you, awesome!

Can you please put the issue numbers from your opening comment into the commit message? It's easier to trace back when needed: there's no commitID->issue lookup other than that as far as I know.

sjudd:
Thanks!

PR #962, opened 2016-02-05
Fix comment to make it clear that target.cancel() has to be called when exception is thrown.

aam:


TWiStErRob:
I think that's not correct, you should also clear it when there was no error: "done with it", so that `catch` should be `finally`.

aam:
@TWiStErRob , thanks for taking a look, but that is exactly it: you should not need to call cancel() if get() completed successfully. That matches semantics outlined here: https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Future.html. Why do you think you should?

TWiStErRob:
It's confusing because that object is both a Future and a Target. As you say it's not good to cancel successful futures, but it's suggested to clear Targets that are not needed any more. Clear would free up any resources, but I guess it doesn't really apply to loads that end in a File, because they're not resource intensive. @sjudd?

sjudd:
Looks like the build is failing due to the comment:

```
:third_party:gif_decoder:mockableAndroidJar/home/travis/build/bumptech/glide/library/src/main/java/com/bumptech/glide/request/transition/ViewPropertyTransition.java:12: warning - Tag @link: reference not found: android.animation.ObjectAnimator /home/travis/build/bumptech/glide/library/src/main/java/com/bumptech/glide/request/RequestFutureTarget.java:50: warning - End Delimiter } missing for possible See Tag in comment string: "A {@link java.util.concurrent.Future} implementation for Glide that can be used to load resources in a blocking manner on background threads. <p> Note - Unlike most targets, RequestFutureTargets can be used once and only once. Attempting
```

A little odd that it's causing the failure but might make sense.


@TWiStErRob @aam I took another look at this. Clear and cancel are equivalent, except that cancel unblocks the future's get method. Both methods will cause the request to be cancelled and resources to be released. 

From that perspective, it's probably best if we do in fact call clear() in a finally block (clear() is also thread safe).

aam:
which clear()? `requestManager.clear(target);` or `target.getRequest().clear()`?

sjudd:
``requestManager.clear(target);``

TWiStErRob:
You don't need `cancel`, because `clear` does that in a thread-safe way, consider this:
```java
/**
 * <pre>{@code
 * FutureTarget<File> target = null;
 * RequestManager requestManager = Glide.with(context);
 * try {
 *     target = requestManager
 *              .downloadOnly()
 *              .load(model)
 *              .submit();
 *     File downloadedFile = target.get();
 *     // ... do something with the file (usually throws IOException)
 * } catch (ExecutionException | InterruptedException | IOException e) {
 *     // ... bug reporting or recovery
 * } finally {
 *     // make sure to cancel pending operations and free resources
 *     requestManager.clear(target);
 * }
 * }</pre>
 */
```
This version also formats well when rendered as JavaDoc and it compiles (requestManager is undefined for you). Also WIDTH/HEIGHT is unnecessary confusion for `downloadOnly`, probably very few people use bucketing: the downloaded image File will be the original, not a resized one.

After incorporating this, can you please squash your commits into one? Don't worry about the current build break, it looks like it's an intermittent issue and if you force push it'll be re-run and likely succeed.

aam:
thanks Rob. But what about calling Future.cancel() on the FutureTarget?
Perhaps if we use .get(TIMEOUT), it becomes evident that in the exception handler besides clearing Target, the Future needs to be cancelled too.

TWiStErRob:
Why would you need to cancel it? Your code is already past `get()`, which it must be because you're in finally, throwing or not. This means that it was either already cancelled or it returned a value and processing the returned File is completed or failed with an Exception.

aam:
When Future `get()` throws an exception, the Future itself still might be running. In case of `get(timeout)` and `TimeoutException` it is particularly clear: take a look at here: https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Future.html

TWiStErRob:
If `get()` throws `CancellationException` or `ExecutionException` then the job definitely failed. If `get()` returns a value it must have finished. If `get(timeout)` or `get()` throws `InterruptedException`, or if `get(timeout)` throws `TimeoutException` the job might still be running, you're right. So it's likely the best to keep the original `target.cancel()` wrapped in a null-check, and with parameter `true` (that param is not used, but in case it will be in the future I think `true` is better there, because the user really wants to cancel it). Glide also adheres to this sentence from the docs:
> Once a computation has completed, the computation cannot be cancelled.

This means it's safe to cancel in `finally`: the state of `isCancelled()` and `isDone()` won't be modified if `get()` completed without exception.

What do you think?

TWiStErRob:
@sjudd where is the `FutureTarget` cancelled when `requestManager.clear(target)`? There's no code in `onLoadCleared`.

aam:
Yes, right.
It should be safe to call `target.cancel(true)` in `finally`, but it seems confusing, as you only _need_ to cancel if one of those exceptions was thrown, right? I would keep it in `catch` clause in the javadoc section for clarity. Of course user is free to move the cancel around if they wish.

aam:
Or is your point that you don't want to call `cancel()` (in the `catch` clause) followed up by call to `clear()` (in `finally` clause) as second is effectively redundant in case of exception?

TWiStErRob:
Hmm, ok, I see your point now. Yes, that's one of my concerns: the target is double-cleared: `cancel` clears the target and `requestManager.clear` also clears the target. Here's one that looks like satisfies all contracts:
```java
try {
    File downloadedFile = target.get();
    // ... do something with the file (usually throws IOException)
    requestManager.clear(target);
} catch(...) {
    target.cancel(true);
}
```
It just feels a little weird to handle them differently, remember I/O streams: you just `close()` them error case or not. I personally prefer my above `try-finally` with cancel: even if the naming is weird, less stuff to worry about.

Alternatively we could add `Closable` which may confuse users more, but at the same time you would be able to say: `try (Target target = Glide.....) { work with return value of target.get() }` after API 19.

aam:
Okay, that is fine, I put `cancel()` into `finally` clause.

The build still fails unfortunately, it seems to just hang for some reason.

TWiStErRob:
Likely the parallel Gradle build. The newer plugins use much more resources with greater parallelism in-task (most notably dex), so parallelizing in-process (on Gradle level) really takes a toll on the host system (sometimes I can't even click with 8 cores and 16GB). I'm not sure if v1.2.3 is new enough to already do this though.

sjudd:
Thanks!

PR #992, opened 2016-02-22
Use only one byte[]/int[] pool

liangricha:
ByteArrayPool and LruByteArrayPool can also
be removed in favor of ArrayPool


TWiStErRob:
Nice, I love negative code!

Do you think it would be possible (/worth) to genericize ArrayPool? I guess it depends on if there's any place where Glide has a pool which stores both `byte[]`, `int[]`, maybe `char[]` in one object.

library/src/main/java/com/bumptech/glide/load/engine/cache/MemorySizeCalculator.java
```diff
@@ -129,7 +129,6 @@ private static boolean isLowMemoryDevice(ActivityManager activityManager) {
     static final int BITMAP_POOL_TARGET_SCREENS = 4;
     static final float MAX_SIZE_MULTIPLIER = 0.4f;
     static final float LOW_MEMORY_MAX_SIZE_MULTIPLIER = 0.33f;
-    // 4MB.
```
sjudd: Retain the comment? Or reference the constant in ArrayPool and remove the one here?

sjudd:
Can you delete the ByteArrayPool/LruByteArrayPool files now?

Thanks for doing this! Other than the one minor nit, lgtm.

library/src/main/java/com/bumptech/glide/load/engine/cache/MemorySizeCalculator.java
TWiStErRob: Did you mean `LruArrayPool.DEFAULT_SIZE`?

TWiStErRob:
Those two files are removed, aren't they? They're fully red in the diff.
Where's the CI build?

sjudd:
Ah yes those files are deleted, my mistake.

sjudd:
I'm not sure Travis ran, I don't see a link here or a run on the website: https://travis-ci.org/bumptech/glide/pull_requests

library/src/main/java/com/bumptech/glide/load/engine/cache/MemorySizeCalculator.java
liangricha: :+1: good call - I can revert this delete.

liangricha:
@sjudd reverted the comment deletion!

@TWiStErRob Hmm I'm not sure what you mean about genericizing `ArrayPool` - doesn't it already support arrays of any type (just have to implement the adapter interface)?

TWiStErRob:
Exactly, it supports different types of arrays in one pool. Compare it to `List x` which accepts any type of object; and `List<T> y` which only accepts `T`s. Currently `arrayPool.put(5, Number.class)` would compile just fine, because only the method is generic, if we had `ArrayPool<byte[]>` it would be more safe. The adapter structure can be kept, but the selection can be moved to the constructor.

This is out of scope for this pull, just inquiring your opinion. Also @sjudd?

sjudd:
Ah @TWiStErRob ``ArrayPool`` uses a single fixed size, so that an array of one type can use all available memory, and it also avoids having to pass through multiple variations. I agree generics are safer, but in this case we've chosen to go for more flexibility.

sjudd:
LGTM Thanks!

Issue #491, opened 2015-06-10
OOM on Glide error


Tolriq:
Don't know if it should be handled by default or offer an option for it, but Glide can currently OOM on decoding error / placeholder drawables.

Since this happens out of application control it's impossible for the application to catch that and react accordingly.

(I normally catch those OOM and clear backstack then glide cache to recover most of the time).

``` ruby
java.lang.OutOfMemoryError
       at android.graphics.BitmapFactory.nativeDecodeAsset(BitmapFactory.java)
       at android.graphics.BitmapFactory.decodeStream(BitmapFactory.java:557)
       at android.graphics.BitmapFactory.decodeResourceStream(BitmapFactory.java:394)
       at android.graphics.drawable.Drawable.createFromResourceStream(Drawable.java:785)
       at android.content.res.Resources.loadDrawable(Resources.java:2016)
       at android.content.res.ResourcesEx.loadDrawable(ResourcesEx.java:580)
       at android.content.res.Resources.getDrawable(Resources.java:697)
       at com.bumptech.glide.request.GenericRequest.onException(GenericRequest.java:548)
       at com.bumptech.glide.load.engine.EngineJob.access$200(EngineJob.java:22)
       at com.bumptech.glide.load.engine.EngineJob$MainThreadCallback.handleMessage(EngineJob.java:204)
       at android.os.Handler.dispatchMessage(Handler.java:103)
       at android.os.Looper.loop(Looper.java:194)
       at android.app.ActivityThread.main(ActivityThread.java:5468)
       at java.lang.reflect.Method.invokeNative(Method.java)
       at java.lang.reflect.Method.invoke(Method.java:525)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:936)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:703)
       at dalvik.system.NativeStart.main(NativeStart.java)

```

TWiStErRob:
Does #435's fix help you?

(Please don't forget to fence your stack traces, it breaks and becomes unreadable.)

Tolriq:
Well no it does not as it's to LOG by default and OOM are a little different to handle in Java.

TWiStErRob:
Oh, sorry I thought there was a custom handler too.

TWiStErRob:
Can't you use `THROW` with `Thread.unCaughExceptionHandler`? ... Oh, maybe `onException` is not covered by that, but worth checking.

[...]

Issue #859, opened 2016-01-04
OOM callback


tibbi:
Hello, Im sometimes loading very huge images, so out of memory on some devices is common. Its not a huge issue, but Id like to react on it. Is there any callback for Out of memory, or any way to find it out? Shouldnt onLoadFailed be called if some exception is thrown? I can create a workaround with onResourceReady and some timer to check if the image was loaded properly, but Id like some nicer solution. So how should I react to the internal Glide Out of memory exception?

tibbi:
seems to be related to https://github.com/bumptech/glide/issues/491

TWiStErRob:
Yes... the problem with delegating OOM to `onLoadFailed` is that it's not an `Exception`, it's an `Error` which are two distinct subclasses of `Throwable`. Though you may have a solution by replacing the default model loader and adding a try-catch to it, I'm looking.

TWiStErRob:
Sorry, not custom loader, custom decoder:
```java
class OOMReadyStreamBitmapDecoder extends StreamBitmapDecoder {
	public OOMReadyStreamBitmapDecoder(Context context) {
		super(context);
	} // or any other ctor you want from super
	@Override public Resource<Bitmap> decode(InputStream source, int width, int height) {
		try {
			return super.decode(source, width, height);
		} catch (OutOfMemoryError err) {
			throw new RuntimeException(err);
		}
	}
}
```

and then just use it:
```java
String url = "https://upload.wikimedia.org/wikipedia/commons/c/c6/BlankMap-World-large3.png";
BitmapRequestBuilder<String, Bitmap> oomRequest = Glide
		.with(context)
		.fromString()
		.asBitmap()
		.imageDecoder(new OOMReadyStreamBitmapDecoder(context))
		//.cacheDecoder(new FileToStreamDecoder<>(new OOMReadyStreamBitmapDecoder(context))) // needed for .diskCacheStrategy(SOURCE|ALL)
		//.error(R.drawable.ic_launcher) // displayed when OOM (because throw RuntimeException)
		//.override(Target.SIZE_ORIGINAL, Target.SIZE_ORIGINAL) // to make sure it OOMs (url is huge)
		;
```
`oomRequest` is created to prevent instantiations for each bind, it's a "template" for the loads.

```
oomRequest
		.load(url)
		.listener(new RequestListener<String, Bitmap>() {
			@Override public boolean onException(Exception e, String model, Target<Bitmap> target, boolean isFirstResource) {
[...]

PR #1057, opened 2016-03-15
Catch OutOfMemoryError in decode for v3

ihenchi:


ihenchi:
All commit authors have now been verified and signed the CLA.

TWiStErRob:
Hi, thanks.
What's the idea behind this?
How is this going to solve OOMs?

dmapr:
It is not going to solve OOMs, but it will ensure that you will get a callback (onException) in your listener when an OOM occurs. Right now if an OOM occurs then you don't get any callback (neither onException nor onResourceReady). In our particular case it happens when loading large animated GIFs and it allows us to fallback to single frame loading.

TWiStErRob:
I don't feel comfortable with that "fake" `RuntimeException`, but there isn't much we can do about the Error vs Exception hierarchy clash in v3. Let's see what @sjudd says.

Btw, are you aware of #859?

dmapr:
Thanks! We weren't aware of it; somehow it hasn't popped up when we were searching for the existing issues. @ihenchi, let's take a look at the approach suggested in the  https://github.com/bumptech/glide/issues/859 and see if we can get that to work. We also gotta add a test case for OOM in the EngineRunnableTest to bring the coverage back where it belongs.

sjudd:
Sorry for the delay. I'm torn here. I don't generally like the idea of catching OOMs, it's usually a sign that something else is wrong and catching the OOMs just delays the inevitable. Super large GIFs or other assets that we can't downsample effectively make this seem more reasonable... 

I guess it's reasonable to at least expose the error so you can respond to it. 

Thanks for putting this up!

TWiStErRob:
@ihenchi @dmapr where did #859 lead you to?

I'm starting to agree with this PR, but I think it needs to be amended a little:
 * deliver the OOM (wrapped in an exception) to `listener.onException`,
 * I would say a `ErrorWrappingGlideException` or similar would probably a good way to distinguish (internally and in listener too)
 * if there's no listener or the listener doesn't return true (=handled), trigger default behavior via:

    ```java
    throw (OutOfMemoryError)ex.getCause()
    ```
 * let the thread pool handle it with its `UncaughtThrowableStrategy` as before.

@sjudd currently default strategy is LOG anyway, so Glide catches the OOM, but there's no way to handle it, only if the developer sees it in the log during development can act on it. Also the current behavior is to just display the placeholder in perpetuity, because the thread handling the request that would've called `target.onLoadFailed`/`listener.onException` is "dead" (stopped prematurely). With this PR either a "retry with different load" or "show .error()" action can be implemented.

TWiStErRob:
Note: a transformation, or transcoder can also throw this error.

dmapr:
@TWiStErRob, at a glance it seemed like we would need to implement the https://github.com/bumptech/glide/issues/859 for multiple decoders (Bitmap, GIF, etc.) so it was less attractive than doing it in the EngineRunnable.

TWiStErRob:
This would fix #491.

sjudd:
Ok I'm ok with this if you create a new type of Exception, rather than just a runtime exception. A specific subclass will make it easier to catch this particular case. Otherwise lgtm.

dmapr:
Sorry for the delay, I was on vacation and @ihenchi had his hands full with the day job. We'll try to put this together to satisfy your requirements.

TWiStErRob:
No worries, better taking the time than getting it wrong.

library/src/main/java/com/bumptech/glide/load/engine/exceptions/ErrorWrappingGlideException.java
```diff
@@ -0,0 +1,11 @@
+package com.bumptech.glide.load.engine.exceptions;
+
+/**
+ * An exception class used for wrapping and distinguishing errors such as
+ * {@link OutOfMemoryError}.
+ */
+public class ErrorWrappingGlideException extends Exception {
+    public ErrorWrappingGlideException(Throwable throwable) {
```
TWiStErRob: I think if it's error wrapping it should expect an `Error`.

Also along the same lines maybe override getCause() with a covariant return type:
```java
public Error getCause () ...
```

library/src/main/java/com/bumptech/glide/load/engine/exceptions/ErrorWrappingGlideException.java
```diff
@@ -0,0 +1,11 @@
+package com.bumptech.glide.load.engine.exceptions;
```
TWiStErRob: @ihenchi I don't think a new package is necessary here.

@sjudd how will we forward-port this? Use https://github.com/bumptech/glide/blob/master/library/src/main/java/com/bumptech/glide/load/engine/GlideException.java or keep this error-wrapping one?

TWiStErRob:
I was wondering why you ignored my 3rd point in https://github.com/bumptech/glide/pull/1057#issuecomment-201005522, but then I investigated and I saw that it's not really possible, because that would crash the host application, because the exception is moved from the fetcher thread to the UI thread by the time we ask the listener if it's handled. @sjudd Do you think it is safe to ignore that point?

library/src/main/java/com/bumptech/glide/load/engine/exceptions/ErrorWrappingGlideException.java
dmapr: I don't think the GlideException is there on the 3.x branch?

TWiStErRob: Yes, but we'll need to forward-port this change it to v4, otherwise v3 and v4 diverges.
And if we keep this exception when forward-porting it would make sense to have them in the same package from the beginning.

dmapr:
Yes, sorry, forgot to reply to the third point. Yes, I don't think we can guarantee that the unprocessed exception can be delivered to the UncaughtThrowableStrategy once the EngineRunnable completes.

TWiStErRob:
There would be way, but it's hacky: somehow manually post a new Runnable to the thread pool which just re-throws the error.

dmapr:
I believe such an approach would increase the complexity beyond our comfort level. Let's wait to see what @sjudd says before we start down that path.

TWiStErRob:
Agreed, I'm not comfortable with it either, just an option.

TWiStErRob:
@ihenchi @dmapr can you please replace the exception with this (changes: package, ctor, getter):
```java
package com.bumptech.glide.load.engine;

/**
 * An exception class used for wrapping and distinguishing errors such as {@link OutOfMemoryError}.
 */
public class ErrorWrappingGlideException extends Exception {
    public ErrorWrappingGlideException(Error error) {
        super(error);
        if (error == null) {
            throw new NullPointerException("The causing error must not be null");
        }
    }

    @Override
    public Error getCause() {
        // cast is safe because constructor ensures there must be a cause, and it must be an Error
        return (Error) super.getCause();
    }
}
```

and then squash all the commits into one.

@TWiStErRob I've noted [above](https://github.com/bumptech/glide/pull/1057#issuecomment-201006281) that transcoders and transformations can also throw, but they're both deep within `decode()` so they're all caught.

@sjudd please acknowledge that it is safe to ignore my third point above, I think that's a breaking change as it invalidates the existence of the central `UncaughtThrowableStrategy` for OOM and would force users to attach listeners to each load to even see OOMs in logs. Catching the OOM will delegate it outside the executor's reach, and we only know on the UI thread in GenericRequest whether we actually have a listener or not to decide what to do. Throwing at that point would always crash the app, instead of the current perpetual placeholder behavior.

sjudd:
UncaughtThrowableStrategy isn't really an API, it's a way to expose errors that occur, primarily for developers. I think what we're saying with this PR is that an OOM shouldn't really be an error because there are times where OOMs are unavoidable.

TWiStErRob:
Ah, true that.

> would force users to attach listeners to each load to even see OOMs in logs.

Actually, this is currently true for all exceptions, if you want the stack trace. Since the exceptions are logged in `log.tag.EngineRunnable` anyway, the OOMs would be logged too.

So the change that will occur when merging this is that instead of perpetual placeholder OOMs would be showing the error drawable like any other exception, and the lifecycle methods will be called as expected. Degrading OOM from unexpected error to expected error.

@ihenchi @dmapr I think we're all on the same page. Please do the actions in https://github.com/bumptech/glide/pull/1057#issuecomment-231612179 and we're good to go.

dmapr:
@TWiStErRob @sjudd — updated as requested.

sjudd:
Thanks! I appreciate your patience here.

dmapr:
Thanks! Do you have an ETA for 3.8.0?

TWiStErRob:
Not sure, but you can use it right away: https://github.com/bumptech/glide/wiki/Snapshots

PR #1071, opened 2016-03-18
Introduce issue templates "the right way"

### LINKED ISSUE #332: Introduce issue templates
We know well that Google Code has a nice template for each bug, sadly Github doesn't. Almost every issue here has "Glide/Android/Phone version?" as the first response after a bug is reported.

This could be prevented/"automated" with the following steps:
* add "report issue" button/link to Readme.MD at the top
* add "report issue" button/link to wiki
* add contributing.md describing how to make a nice & efficient issue (formatted and full of *relevant* info)
With a copy-pasteable markdown template and a "report issue" link.

See https://github.com/isaacs/github/issues/99 for more info.

The "report issue" link would look like this (example, could be refined):
https://github.com/bumptech/glide/issues/new?body=**Glide%20Version/Integration%20library%20(if%20any)**:%0A**Device/Android%20Version**:%0A**Is%20it%20consistently%20reproducible?**:%0A**Issue%20details/Repro%20steps**:%0A```%0AGlide.with(context)...%3B%0A```%0A%0A**Stack%20trace**:%0A```%0Apaste%20stack%20trace%20here%0A```

TWiStErRob:
## Description
Re-fix #332 by adding a GitHub supported issue (and PR) template

## Motivation and Context
Finally GitHub has provided us way to force users to notice the issue template we have in place.
It was announced in https://github.com/blog/2111-issue-and-pull-request-templates

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,31 +12,6 @@ When in doubt, file an issue. We'd rather close a few duplicate issues than let 
 Similarly if you support a particular feature request, feel free to let us know by commenting on the issue or [subscribing][6] to the issue.
 
 To file a new issue, please use our issue template and fill out the template as much as possible (remove irrelevant parts).
-
-<pre>**Glide Version**:
-**Integration libraries**:
-**Device/Android Version**:
-**Issue details / Repro steps / Use case background**:
-
-**Glide load line**:
-```java
-Glide.with(...).....load(...).....into(...);
-```
-
-**Layout XML**:
-```xml
-&lt;...Layout&gt;
-    &lt;ImageView android:scaleType="..." ... /&gt;
-&lt;/..Layout&gt;
-```
-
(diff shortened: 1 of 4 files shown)


TWiStErRob:
Something to consider: extract those HTML comments to a wiki page and keep a link at the top only with a few keywords.

sjudd:
Awesome thanks for doing this!

TWiStErRob:
What do you think about my comment above?

sjudd:
You're more in tune with how much people pay attention to the template. If you think it would be better to link out to a wiki page, we can do that.

We could try it as is for a few weeks and see what happens, then switch to a link and see if there's any difference?

TWiStErRob:
They don't, really. I hope by having it there already it will be more prominent. I'm worried a little that it's too long that they'll just delete it and start writing the usual reports.
I agree, a little temporal A/B testing makes sense.

Issue #1062, opened 2016-03-17
how to release memory when using glide together with Html.fromHtml,thanks


ourgit:

ourgit:
I am trying parse the dynamic emotion to GlideImageGetter by Html.fromHtml, the gif show corretly,but it leak memory out. I wanna release the memory at instant when the listview scrolls or the datachaged,which make some gif off the screen. below is the details code,hope guys help me out.thanks.

1.
` Spanned spanned = EmotionParser.parseEmotion(instance, guildchatmsg.getBody(), cache.tv_chatcontent_out_mul);
                        cache.tv_chatcontent_out_mul.setText(spanned);`

2.
`public static Spanned parseEmotion(Context context, String content, TextView tv, int type) {
        Pattern pattern = Pattern.compile(FACE_PATTERN, Pattern.CASE_INSENSITIVE);
        if (null == content || content.length() < 1) return null;
        Matcher matcher = pattern.matcher(content);
        StringBuilder sb = new StringBuilder();
        int start = 0;
        int emotionCount = 0;
        while (matcher.find()) {
            int matchStart = matcher.start();
            int matchEnd = matcher.end();
            String group = matcher.group();
            emotionCount++;
            String resultFile = group.substring(0, group.length());
            String resourcePath = "android.resource://" + context.getPackageName() + "/drawable/" + resultFile;
            sb.append(content.substring(start, matchStart)).append("<img src=\"" + resourcePath + "\"/>");
            start = matchEnd;
        }
        String result;
        Spanned spanned = null;
        if (start > 0) {
            int len = content.length();
            if (start == len) result = sb.toString();
            else result = sb.append(content.substring(start, len)).toString();
            int width = 0;
            if (type == 1) {
                width = getRecommendScaleOnBig(context);
                if (emotionCount > 8)
                    spanned = Html.fromHtml(result, new GlideImageGetter(context, tv, false, width), null);
                else
                    spanned = Html.fromHtml(result, new GlideImageGetter(context, tv, true, width), null);
            } else if (type == 2) {
                width = getSpecifiedWidthOnSmall(context);
                spanned = Html.fromHtml(result, new GlideImageGetter(context, tv, false, width), null);
            }
        } else {
            result = content;
            spanned = Html.fromHtml(result, null, null);
        }
        return spanned;
    }`

3.
`public final class GlideImageGetter implements Html.ImageGetter, Drawable.Callback {

    private static final String TAG = "GlideImageGetter";
[...]

PR #1090, opened 2016-03-25
Fix thread-unsafe behaviours in GifDecoder

### LINKED ISSUE #1068: black square background appears under the transparent gif.
I am using glide to show multi emotion gifs, most of time the glide shows the gifs correctly,however,there would be black square background appearing under the transparent gif sometimes ,and the black square background will disappear if I scroll the listview. Any solution for that? thanks

TWiStErRob:
* use a local copy of `act`: fixes #1068  
See investigation below https://github.com/bumptech/glide/issues/1068#issuecomment-199776115
 * don't modify `bgColor` unnecessarily  
There was a TODO in `GifHeader` and it coincided with the code I'm modifying. I also checked all other `GifHeader` and `GifFrame` fields, after this change all of them are used read-only. Please double-check that this move doesn't modify the behavior (note: `bgColor` is only used in this single place)

Tested in Glide Support with #1062's example:
https://github.com/TWiStErRob/glide-support/tree/c3de297d0daec8d2650a3a640c0404c35b937a30/src/glide3/java/com/bumptech/glide/supportapp/github/_1062_html_gif_spans

### DIFF
--- a/third_party/gif_decoder/src/main/java/com/bumptech/glide/gifdecoder/GifDecoder.java
+++ b/third_party/gif_decoder/src/main/java/com/bumptech/glide/gifdecoder/GifDecoder.java
@@ -105,7 +105,10 @@ public class GifDecoder {
 
     // Global File Header values and parsing flags.
     // Active color table.
+    // Maximum size is 256, see GifHeaderParser.readColorTable
     private int[] act;
+    // Private color table that can be modified if needed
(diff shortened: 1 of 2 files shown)


sjudd:
LGTM, thank you!

TWiStErRob:
@sjudd why didn't this close the related issue? It says "fixes #1068" in the commit message.

Issue #1060, opened 2016-03-16
GIF - Significant delay before showing the second frame for the very first loop.


vishalpawale:
**Glide Version:** 3.6.1
**Issue details:**
While loading GIF's there is significant delay before showing **only** the second frame, and this happens only for the first loop, all the later loops goes with expected delay.

So after doing some research, I thought it may be because Glide is resizing the frames, so I override resize operation using `override()`.

**Below is the Glide load line:**
`Glide.with(this).load(trainingItem.getUrl()).asGif().override(Target.SIZE_ORIGINAL, Target.SIZE_ORIGINAL).diskCacheStrategy(DiskCacheStrategy.SOURCE).into(draweeView);`

But still there is significant delay before showing the next frame for the very first loop.
Any suggestions/pointers?

TWiStErRob:
GIF frames are decoded in the same pool as normal images. Are you displaying multiple GIFs or having a lot of other images loaded at the same time?

vishalpawale:
No, I'm showing only one GIF in the whole activity. And even the view hierarchy of that activity is very simple, not so deep. GIF size is around 96KB. If I use high res GIF sizing around 500KB, delay goes upto 20s.

TWiStErRob:
Hmm, that's a lot, those kinds of delays occurred when people are using RESULT cache, which you're not.

I just noticed you call it `draweeView`, are you loading into a standard `ImageView`? If not, please try that.

vishalpawale:
Also, going through code, it looks like we have 30ms as min frame time, and 100ms is default. 

Also it looks to me that we don't have any provision to set maximum frame timing, but in case if we have that provision, I'll set that max frame time to 4-5s and get a workaround for this issue.

vishalpawale:
Yes, I'm loading into standard `ImageView`. 
Please ignore the naming for now, initially I was using Fresco library, but later found Glide really cool. Awesome job man, Thanks for the Glide. :)

TWiStErRob:
Yep, too fast animation is not supported, intentionally, by Glide. I think that min frame time is hardcoded and not configurable. If you want smooth, use a video or an Android animation. Though you're right that if the GIF file contains frame delay value that's 20 seconds Glide will wait.

It's hard to imagine there would be difference in delay on first and other playbacks on purpose. Can you share the image or the whole demo app?

vishalpawale:
Yes, min frame time is not configurable. And I can not go for video or an Android animation, since its a big change on server side as well.

[...]

PR #1091, opened 2016-03-25
Get frame delay before advancing to show the current frame for the correct amount of time

TWiStErRob:
See https://github.com/bumptech/glide/issues/1060#issuecomment-201023023
Confirmed working by manually testing with the image I created from the original.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/load/resource/gif/GifFrameLoader.java
+++ b/library/src/main/java/com/bumptech/glide/load/resource/gif/GifFrameLoader.java
@@ -97,8 +97,8 @@ class GifFrameLoader {
         }
         isLoadPending = true;
 
-        gifDecoder.advance();
         long targetTime = SystemClock.uptimeMillis() + gifDecoder.getNextDelay();
+        gifDecoder.advance();
         DelayTarget next = new DelayTarget(handler, gifDecoder.getCurrentFrameIndex(), targetTime);
         requestBuilder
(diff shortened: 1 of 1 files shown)


third_party/gif_decoder/src/main/java/com/bumptech/glide/gifdecoder/GifDecoder.java
```diff
@@ -195,7 +195,7 @@ public void advance() {
      */
     public int getDelay(int n) {
         int delay = -1;
-        if ((n >= 0) && (n < header.frameCount)) {
+        if (0 <= n && n < header.frameCount) {
```
sjudd: nit: I prefer the other ordering for <= or ==, it's easier for me to read "n is greater than or equal to 0" than it is "0 is less than or equal to n".

TWiStErRob: How about "`n` is *between* 0 and `frameCount`"?
It's literally between the two bounds in the code as well.

sjudd: Yeah I'd prefer 0 <= n < frameCount if the language supported it, but as is, I'm used to reading variable <=> number.

TWiStErRob:
nit fixed

sjudd:
Thanks!

Issue #1059, opened 2016-03-16
The Gif load problem, can't display well


JKingdom:
Hi, I load a gif image from local
The gif：
![Image](http://file.baixing.net/201511/40dfc16a7965fff85bc6f6ea2d854f17.gif)
and other gifs:
![Image](http://file.baixing.net/201511/59aa0ca0c1d0c330983a110d605ff940.gif)
![Image](http://file.baixing.net/201511/080408bac9579a07500c212e40bb96ee.gif)

I use 
`      int source = R.drawble.xxx;`
`
Glide.with(this).load(source).asGif().diskCacheStrategy(DiskCacheStrategy.SOURCE).into(imageView);
`
to load those Gifs

But I find the gif1 and gif2 can't display very well, though the gif can display on the phone, but it flick(flash , twinkle, glint  sorry I'm not a English speaker, I don't know how to express it well, all in all the Gif1 and Gif2 in glide are not playing well, It seems that the last frame doesn't  disappear)

However the 3rd Gif is normal.

I'm confused about the question, I try many gifs and find it seems that when the gif's background is transparent it will appear(not for all, just  some of them)

I use Glide 3.6.1
and also find this problem in 3.7.0

But when I choose another gif lib like:
[koral--/android-gif-drawable](https://github.com/koral--/android-gif-drawable)
it display all gifs very well.

So I help to solve this problem.
 Thank you very much!

TWiStErRob:
I think what you see is #836, the fix is done for v4, but needs to be copied into v3, this is tracked in #970. Please confirm that the "screenshot" in #836 is similar to your "flash".

There's also a possibility of using external libraries to display stuff, see #805.

**Tip**: don't use `.diskCacheStrategy(SOURCE)` for local images, that will duplicate the file (in APK, and on SDcard) without any performance benefit. Use NONE for local GIFs.

**Tip**: use `adb screen record /sdcard/demo.mp4`, then `adb pull /sdcard/demo.mp4`, then upload at http://ezgif.com/video-to-gif to make a "moving screenshot" that you can embed on GitHub/StackOverflow.

JKingdom:
Thank you very much for you answer, I will check soon.
But I have another question: 
I use `.diskCacheStrategy(SOURCE)` because I think I load gif is too slow, so I find the issue [#600](https://github.com/bumptech/glide/issues/600)

so I use `.diskCacheStrategy(SOURCE)`  in my code

[...]

PR #1093, opened 2016-03-25
Fill only the area that was modified by the previous GIF frame

TWiStErRob:
## Description
As seen in #1059:
flickering GIF animation

If you look closely you can see there are multiple different flickers:
 * one where the frame is partially shown
 * one where a single frame is almost fully transparent

This is how it should look like:
original image

This change achieves that by clearing only part of the frame when need to dispose to background.

## Motivation and Context
I sent the above image through http://ezgif.com/split ("Ignore optimizations") to see what how they look:
image

As you can see there's an actual half-frame and there are many almost-empty frames, most importantly the one where there's a little pink on the left. These two frames cause the trouble. Their dispose is set to "background" which triggers a code block that fills the entire frame with `bgColor`. The problem is that the frame doesn't contain a full-sized image, just a partial one. This is also confirmed in debug (closed frames look identical to 0th frame): 
image

I checked the GIF spec and it says:
>  iv) Disposal Method - Indicates the way in which the graphic is to be treated after being displayed.
>  2 -   Restore to background color. The **area used by the graphic** must be restored to the background color. -- [GIF89a](https://www.w3.org/Graphics/GIF/spec-gif89a.txt), emphasis mine

This means that when preparing the frame transition from 2 to 3, 2 wants it's partial area to be cleared as defined by the (ix,iy iw,ih) rect. I took a shot at implementing that, I went through these iterations to arrive at the code in the PR.

1. simple iteration with cached end values

	```java
	for (int y = previousFrame.iy, yEnd = y + previousFrame.ih; y < yEnd; y++) {
	    for (int x = previousFrame.ix, xEnd = x + previousFrame.iw; x < xEnd; x++) {
	        dest[y * width + x] = c;
	    }
	}
	```

2. pre-calculate values and name them

	```java
	int top = previousFrame.iy;
	int bottom = previousFrame.iy + previousFrame.ih;
	int left = previousFrame.ix;
	int right = previousFrame.ix + previousFrame.iw;
	for (int y = top; y < bottom; y++) {
	    for (int x = left; x < right; x++) {
	        dest[y * width + x] = c;
	    }
	}
	```

3. lift multiplication up

	```java
	int top = previousFrame.iy;
	int bottom = previousFrame.iy + previousFrame.ih;
	for (int y = top; y < bottom; y++) {
	    int left = y * width + previousFrame.ix;
	    int right = y * width + previousFrame.ix + previousFrame.iw;
	    for (int p = left; p < right; p++) {
	        dest[p] = c;
	    }
	}
	```

4. lift multiplication up again and rename variables

	```java
	int topLeft = previousFrame.iy * width + previousFrame.ix;
	int bottomLeft = topLeft + previousFrame.ih * width;
	for (int left = topLeft; left < bottomLeft; left += width) {
	    int right = left + previousFrame.iw;
	    for (int pointer = left; pointer < right; pointer++) {
	        dest[pointer] = c;
	    }
	}
	```

Not sure which one is best, I think the final version is readable enough and all multiplications are lifted outside, so it should be efficient as well.

### DIFF
--- a/third_party/gif_decoder/src/main/java/com/bumptech/glide/gifdecoder/GifDecoder.java
+++ b/third_party/gif_decoder/src/main/java/com/bumptech/glide/gifdecoder/GifDecoder.java
@@ -34,7 +34,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
-import java.util.Arrays;
 
 /**
  * Reads frame data from a GIF image source and decodes it into individual frames
@@ -434,7 +433,15 @@ public class GifDecoder {
                         c = 0;
                     }
                 }
-                Arrays.fill(dest, c);
+                // The area used by the graphic must be restored to the background color.
+                int topLeft = previousFrame.iy * width + previousFrame.ix;
+                int bottomLeft = topLeft + previousFrame.ih * width;
+                for (int left = topLeft; left < bottomLeft; left += width) {
+                    int right = left + previousFrame.iw;
+                    for (int pointer = left; pointer < right; pointer++) {
+                        dest[pointer] = c;
+                    }
+                }
             } else if (previousFrame.dispose == DISPOSAL_PREVIOUS && previousImage != null) {
                 // Start with the previous frame
                 previousImage.getPixels(dest, 0, width, 0, 0, width, height);


TWiStErRob:
Worth noting that probably the `DISPOSAL_PREVIOUS` handling should be parametrized similarly as well, something like:
```java
previousImage.getPixels(dest, topLeft, width, ix, iy, iw, ih);
```

sjudd:
LGTM, awesome work!

Issue #394, opened 2015-04-05
Support contact uris


sjudd:
http://stackoverflow.com/questions/29438701/how-to-load-a-uri-with-content-prefix-using-glide-android

R4md4c:
I'm wondering, What do I need to know in order to be able to add this feature to Glide?

TWiStErRob:
PR or for yourself?

R4md4c:
Yes, a PR, I'd like to send a PR.

TWiStErRob:
Cool :) 3.8.0 or v4?
(I'm going out now, I'll give you details later)

TWiStErRob:
Just saw your update, check out https://github.com/TWiStErRob/glide-support/ in the meantime.

R4md4c:
I guess 3.8.0 will be good, as this feature would reach many users who could be afraid of migrating to v4 when it gets released.

TWiStErRob:
Thank you very much for coming with a request like this, it's awesome!

To start dev: first do [Build](https://github.com/bumptech/glide#build) (probably worth doing it on your fork). Then if it works in command line, then [Development](https://github.com/bumptech/glide#development).

This feature needs to be integrated into the leftmost arrow in [this flow](https://docs.google.com/drawings/d/1KyOJkNd5Dlm8_awZpftzW7KtqgNR6GURvuF6RfB210g/edit)

Once your env is up take a look at `StringLoader`, `UriLoader` and its descenants. I think you'll need modify `getLocalUriFetcher` or `StreamLocalUriFetcher.loadResource` (latter looks more reusable) to check the authority (`Uri.getHost()`) and then load the best available version with `ContactsContract.Contacts.openContactPhotoInputStream` or similar. Use the latest available API with `SDK_INT` conditions.

A thing to look out for may be that this needs to work after this new feature is introduced (on API 14+):
```java
Glide.with(...)
    .load(Uri.withAppendedPath(contactUri, Contacts.Photo.DISPLAY_PHOTO)
```
because the authority matches, but it doesn't need to be pushed through `openContactPhotoInputStream` for it to work (it should work with current impl).

Useful resource for debugging: https://github.com/bumptech/glide/wiki/Debugging-and-Error-Handling.

R4md4c:
Thanks for the super helpful introduction! 

I'll try to do what I could, and I think I can do it, but I might ask some questions especially when writing tests. 

The branch that I should work on is 3.0, right?

TWiStErRob:
[...]

PR #1119, opened 2016-04-07
Support contact Uri #394

R4md4c:
So I've supported these kind of Uris for issue #394 , also wrote a simple sample for this feature.

content://com.android.contacts/contacts/5/display_photo
content://com.android.contacts/contacts/lookup/3570i61d948d30808e537
content://com.android.contacts/contacts/38
content://com.android.contacts/contacts/lookup/3570i61d948d30808e537


R4md4c:
I've signed it!

library/src/main/java/com/bumptech/glide/load/data/StreamLocalUriFetcher.java
```diff
@@ -12,17 +19,78 @@
...
 
     @Override
     protected void close(InputStream data) throws IOException {
         data.close();
     }
+
+    private InputStream loadResourceFromContactUri(Uri uri, ContentResolver contentResolver)
+            throws FileNotFoundException {
+        Uri contactUri = uri;
+        int matchedUri = URI_MATCHER.match(contactUri);
```
TWiStErRob: I think `match()` result could be passed in, it's enough to parse the Uri once.

library/src/main/java/com/bumptech/glide/load/data/StreamLocalUriFetcher.java
```diff
@@ -12,17 +19,78 @@
...
+                if (SDK_INT < ICE_CREAM_SANDWICH) {
+                    return ContactsContract.Contacts.openContactPhotoInputStream(contentResolver, contactUri);
+                } else {
+                    return openContactPhotoInputStreamIcs(contentResolver, contactUri);
+                }
+            case ID_THUMBNAIL:
+            case ID_DISPLAY_PHOTO:
+                return contentResolver.openInputStream(contactUri);
+            default:
+                throw new FileNotFoundException("Contact cannot be found");
```
TWiStErRob: I think it would be better to still try to go through `openInputStream` in case of `default` to future-proof it; in case something new is introduced, or something we're not aware of.

library/src/main/java/com/bumptech/glide/load/data/StreamLocalUriFetcher.java
```diff
@@ -12,17 +19,78 @@
...
+    }
+
     public StreamLocalUriFetcher(Context context, Uri uri) {
         super(context, uri);
     }
 
     @Override
     protected InputStream loadResource(Uri uri, ContentResolver contentResolver) throws FileNotFoundException {
+        if (ContactsContract.Contacts.CONTENT_URI.getHost().equals(uri.getHost())
+                && URI_MATCHER.match(uri) != UriMatcher.NO_MATCH) {
```
TWiStErRob: I think match() result could be passed in, it's enough to parse the Uri once.

library/src/main/java/com/bumptech/glide/load/data/StreamLocalUriFetcher.java
```diff
@@ -12,17 +19,78 @@
...
+    /** A contact uri (e.g. content://com.android.contacts/contacts/38) */
+    private static final int ID_CONTACT = 3;
+    /**
+     * A contact display photo (high resolution) uri
+     * (e.g. content://com.android.contacts/5/display_photo)
+     */
+    private static final int ID_DISPLAY_PHOTO = 4;
+
+
+    private static final UriMatcher URI_MATCHER;
```
TWiStErRob: Name should include `Contact(s)`, because it's specific to that; if it was extended in the future the host/authority wouldn't match? Or leave it as is, but change the matching logic to decide solely on this and not the host (doesn't the UriMatcher start with looking up the host in an authority hash anyway?)

library/src/main/java/com/bumptech/glide/load/data/StreamLocalUriFetcher.java
```diff
@@ -12,17 +19,78 @@
...
+            case ID_LOOKUP:
+            case ID_CONTACT:
+                // If it was a Lookup uri then resolve it first, then continue loading the contact uri.
+                if (matchedUri == ID_LOOKUP) {
+                    contactUri = ContactsContract.Contacts.lookupContact(contentResolver, contactUri);
+                    if (contactUri == null) {
+                        throw new FileNotFoundException("Contact cannot be found");
+                    }
+                }
+                if (SDK_INT < ICE_CREAM_SANDWICH) {
```
TWiStErRob: This whole block could be inside a method it's more visible what's the difference between the versions then. It increases the scope of `TargetApi`, but I think it's worth it.

settings.gradle
```diff
@@ -6,6 +6,7 @@ include ':third_party:gif_encoder'
 include ':samples:flickr'
 include ':samples:giphy'
 include ':samples:svg'
+include ':samples:contacturi'
```
TWiStErRob: I think if we're introducing a new project, it may worth doing a catch-all one, so there's a place where we can put single-activity manual tests (I would go as far as merging the other samples into one big sample project in the future, of course it's out of scope for this PR), so instead of `contactsuri` it would be better to call this `misc`/`tests`/`manualTest` or similar.

TWiStErRob:
Nice!

See my comments, treat them as idea-sparking ones, not as sacred texts = feel free to argue them ;)

TWiStErRob:
@sjudd is there any point doing this for `FileDescriptorLocalUriFetcher` too?

library/src/main/java/com/bumptech/glide/load/data/StreamLocalUriFetcher.java
R4md4c: I did it this way because Lint complained about this, if I moved this block into another method Lint will complain and the build will fail with: 
**NewApi: Calling new methods on older versions**

So the only way avoiding this error was doing it that way, if you have another way for it then I'm all ears of course.

TWiStErRob: This seems to work for me (`gradlew library:build library:lint` built with 0 issues):
```java
@TargetApi(Build.VERSION_CODES.ICE_CREAM_SANDWICH)
private InputStream openContactPhotoInputStream(ContentResolver contentResolver, Uri contactUri) {
	if (SDK_INT < ICE_CREAM_SANDWICH) {
		return ContactsContract.Contacts.openContactPhotoInputStream(contentResolver, contactUri);
	} else {
		return ContactsContract.Contacts.openContactPhotoInputStream(contentResolver, contactUri, true);
	}
}
```
IDEA/AS is sometimes sluggish to refresh the warning status, and/or it's possible that you have an older version of lint that doesn't like static imports, or similar issues. Usually to get rid of all NewApi and friends you need to correctly protect the new methods with `SDK_INT` conditions AND `@TargetApi` on the scope.

R4md4c:
@TWiStErRob 

Thanks a lot for your comments, it was really helpful!
I changed the code so that the .match() call happens only once, instead of several times.

library/src/main/java/com/bumptech/glide/load/data/StreamLocalUriFetcher.java
```diff
@@ -12,17 +19,74 @@
...
+
     public StreamLocalUriFetcher(Context context, Uri uri) {
         super(context, uri);
     }
 
     @Override
     protected InputStream loadResource(Uri uri, ContentResolver contentResolver) throws FileNotFoundException {
+        final int matchedUri  = URI_MATCHER.match(uri);
+        if (matchedUri != UriMatcher.NO_MATCH) {
+            return loadResourceFromContactUri(uri, contentResolver, matchedUri);
```
TWiStErRob: What do you think about inlining this call to simplify?
`matchedUri == UriMatcher.NO_MATCH` is the same as the default in the switch.

R4md4c: I'm thinking of doing something like
```
return matchedUri != UriMatcher.NO_MATCH ? loadResourceFromContactUri(uri, contentResolver, matchedUri)  : contentResolver.openInputStream(uri);
```

TWiStErRob: That's not readable, consider:
```java
switch (URI_MATCHER.match(uri)) {
    case ID_LOOKUP:
    case ID_CONTACT:
        ...
        return openContactPhotoInputStream(contentResolver, contactUri);
    case ID_THUMBNAIL:
    case ID_DISPLAY_PHOTO:
    case UriMatcher.NO_MATCH:
    default:
        return contentResolver.openInputStream(contactUri);
}
```
this is what happens anyway, just with bigger cyclomatic complexity and duplicated call to `openInputStream`.

R4md4c: I've updated the PR, into to your more simplified version.

samples/contacturi/src/main/java/com/bumptech/glide/samples/contacturi/MainActivity.java
```diff
@@ -0,0 +1,69 @@
+package com.bumptech.glide.samples.contacturi;
```
sjudd: two space indents please.

library/src/main/java/com/bumptech/glide/load/data/StreamLocalUriFetcher.java
```diff
@@ -12,17 +19,72 @@
...
+                return contentResolver.openInputStream(contactUri);
+        }
+    }
+
+    @TargetApi(Build.VERSION_CODES.ICE_CREAM_SANDWICH)
+    private InputStream openContactPhotoInputStream(ContentResolver contentResolver, Uri contactUri) {
+        if (SDK_INT < ICE_CREAM_SANDWICH) {
+            return ContactsContract.Contacts.openContactPhotoInputStream(contentResolver, contactUri);
+        } else {
+            return ContactsContract.Contacts.openContactPhotoInputStream(contentResolver, contactUri, true);
```
sjudd: this is a nit, but I find it helpful to document null, and non-obvious primitive arguments like so: openContactPhotoInputStream(contentResolver, contactUri, true /*preferHighres*/).

library/src/main/java/com/bumptech/glide/load/data/StreamLocalUriFetcher.java
```diff
@@ -12,17 +19,72 @@
...
+        final int matchedUri  = URI_MATCHER.match(uri);
+        return loadResourceFromUri(uri, contentResolver, matchedUri);
     }
 
     @Override
     protected void close(InputStream data) throws IOException {
         data.close();
     }
+
+    private InputStream loadResourceFromUri(Uri contactUri, ContentResolver contentResolver, int matchedUri)
```
sjudd: rename contactUri -> uri

sjudd:
@TWiStErRob No need to do this for file descriptors, at least not right now. They're currently only used for loading videos. That said, I actually think they'd be more efficient than InputStreams for local content retrieved via a a ContentResolver, so we may revisit that in the future.

sjudd:
Also thanks for putting this up! Other than one or two minor comments looks good.

samples/contacturi/src/main/java/com/bumptech/glide/samples/contacturi/MainActivity.java
TWiStErRob: This is v3, didn't you introduce Google style in v4?

R4md4c:
Hi @TWiStErRob, 

It is been 4 days since the last patch, but I haven't heard any feedback from you, is there something missing from my side?

TWiStErRob:
Oh, sorry, I'm ok with it, but @sjudd has the final word (=presses the merge button). He said those two nits, which you fixed, but we don't get notifications for force push commits (nor normal commits for that matter), you have to comment manually, which you've done now :)

samples/contacturi/src/main/java/com/bumptech/glide/samples/contacturi/MainActivity.java
sjudd: Yup good point, disregard.

sjudd:
Thanks!

PR #1253, opened 2016-06-05
Fix GIF not display with EOF error

### LINKED ISSUE #1145: can not show gif image with last frame broken
this is image address:http://img1.famulei.com/z/0/p/164/2113351333687.gif.

this image cause this exception.
8lq le f 07 170miybuh

the image showed up when we deleted last frame of this gif image.

s1rius:
Fix #1145 
Some image urls to test:
http://img1.famulei.com/z/0/p/164/2113351333687.gif
http://7vzseb.com1.z0.glb.clouddn.com/12nuDeDV4xU3yE.gif
http://p3.pstatp.com/large/2ce000064e9f0b034f1

### DIFF
--- a/library/src/main/java/com/bumptech/glide/load/resource/gif/ByteBufferGifDecoder.java
+++ b/library/src/main/java/com/bumptech/glide/load/resource/gif/ByteBufferGifDecoder.java
@@ -94,7 +94,7 @@ public class ByteBufferGifDecoder implements ResourceDecoder<ByteBuffer, GifDraw
       GifHeaderParser parser) {
     long startTime = LogTime.getLogTime();
     final GifHeader header = parser.parseHeader();
-    if (header.getNumFrames() <= 0 || header.getStatus() != GifDecoder.STATUS_OK) {
+    if (header.getNumFrames() <= 0) {
       // If we couldn't decode the GIF, we will end up with a frame count of 0.
       return null;
     }
--- a/third_party/gif_decoder/src/main/java/com/bumptech/glide/gifdecoder/GifHeaderParser.java
+++ b/third_party/gif_decoder/src/main/java/com/bumptech/glide/gifdecoder/GifHeaderParser.java
@@ -345,7 +345,13 @@ public class GifHeaderParser {
     int blockSize;
     do {
       blockSize = read();
-      rawData.position(rawData.position() + blockSize);
+      int newPosition = rawData.position() + blockSize;
+      if (newPosition <= rawData.limit()) {
+        rawData.position(newPosition);
+      } else {
+        header.status = STATUS_FORMAT_ERROR;
+        return;
+      }
     } while (blockSize > 0);
   }
 


s1rius:
I signed it!

sjudd:
Thanks for putting this up. I'm inclined to accept this I think. Can you briefly describe the new behavior for this images? I'd imagine it will match Chrome by running through all valid frames once and then failing when it reaches an invalid frame?

The only issue I could see is that it doesn't allow the app developer to easily tell that the GIF load won't behave as they expect. We don't have a way of communicating a partial failure at the moment.

s1rius:
Sorry for late.
With this PR, Glide will treat a damaged GIF as a normal GIF which contains some valid frames. The invalid frames will be ignored. And the valid frames will default keep looping not just run once.
If an user see a damaged GIF on web or gallery. I think, make the user see a valid GIF on an app which used Glide is better than display a placeholder image.
@sjudd

TWiStErRob:
@sjudd when looking at the code this is what I concluded too. I guess there's no point it crashing the app when a bad image is loaded, but probably a warning log would be useful to communicate partial failure when `header.getStatus() != GifDecoder.STATUS_OK`.

@s1rius did you actually look into what's wrong with these files? Are there bytes missing from the end (partial file), or is it an encoder error, or maybe something else?

s1rius:
@TWiStErRob I actually look into the wrong with this [file](http://7vzseb.com1.z0.glb.clouddn.com/12nuDeDV4xU3yE.gif), and I compared the bad file and the good file.The bad file lost many bytes. 
compare

TWiStErRob:
> The only issue I could see is that it doesn't allow the app developer to easily tell that the GIF load won't behave as they expect. We don't have a way of communicating a partial failure at the moment.

@sjudd how about putting this logic behind an Option? The default would be as before in v3, but users who expect partial files (like @s1rius) can enable the option? Or the reverse: default is to decode and show everything, but strict users can disable partial decoding.

sjudd:
Option is ok with me for now, good idea.

Ideally you'd want some partial success indicator, but I can't think of way to do that without expanding a bunch of interfaces or changing the behavior of Targets.

s1rius:
I think decode and show everything is a good idea.

If I am an strict user, I will guarantee every GIF upload to server is good. So my App displays every GIF correctly from server.
And I want to ignore bad GIFs stored in local. But I don't know which GIF is bad before decoding it. Instead, I display a placeholder image on the bad Gif position.

s1rius:
Maybe decode and show everything is OK.
If I am an strict developer, I will guarantee every GIF upload to server is good. So my App displays every GIF correctly from server.
And I want to ignore bad GIFs stored in local. But I don't know which GIF is bad before decoding it. Instead, I display a placeholder image on the bad Gif position in my gallery Activity. It is not make sense.
So, I think decode and show everything is a good idea.

chenwulue:
so,how to show gif like those,them can be show by fresco or broswer

Issue #1277, opened 2016-06-21
GifResourceDecoder leaks activity once in memory cache


TWiStErRob:
Based on https://github.com/bumptech/glide/issues/859#issuecomment-227491567

**Glide Version**: 3.7.0

**Integration libraries**: N/A

**Device/Android Version**: all

**Issue details / Repro steps / Use case background**:
Using the `GifResourceDecoder` created with activity context leaks it because it is stored in memory cache. A simple workaround is to use `context.getApplicationContext()` in the constructor, but likely this should be fixed in `GifResourceDecoder`.

**Glide load line / `GlideModule` (if any) / list Adapter code (if any)**:
```java
Glide.with(...).load(gifUrl).decoder(new GifBitmapWrapperResourceDecoder(null, new GifResourceDecoder(context), Glide.get(context).getBitmapPool()));
```

TWiStErRob:
PR merged

PR #1281, opened 2016-06-22
Use application Context in GifResourceDecoder

mcoded:
Very quick fix for #1277
This fixes a memory leak which causes `GifResourceDecoder` to leak `Activity` context once in memory cache. The solution is to save the application context instead when initializing the decoder.


TWiStErRob:
Did you check if it has any possible side effects?

mcoded:
I did. From what I gathered, that `Context` is only used to initialize `GifDrawable`s, which only use it to initialize a `GifState` (which already calls `context.getApplicationContext()` on the `Context` that's provided to it).

sjudd:
Thanks!

PR #1303, opened 2016-07-02
Update build.gradle

Dimous:
## Description
Changed dependancy to official google repository.

## Motivation and Context

mcxiaoke's mirror for volley is deprecated

PR #1316, opened 2016-07-06
Update gradle.properties

Dimous:


TWiStErRob:
This should be in he same PR as #1303 you can force push to your branch and the commits will be picked up.

Dimous:
ok

Issue #1094, opened 2016-03-25
Check disposal handling in GIFs


TWiStErRob:
While building #1093 I noted that probably the `DISPOSAL_PREVIOUS` handling should be parametrized similarly as well, but this needs confirmation by a GIF that does this and fails to render correctly, code is probably:

``` java
previousImage.getPixels(dest, topLeft, width, ix, iy, iw, ih);
```

based on

>  iv) Disposal Method - Indicates the way in which the graphic is to be treated after being displayed.
> 3 - Restore to previous. The decoder is required to restore **the area overwritten by the graphic** with what was there prior to rendering the graphic. -- [GIF89a](https://www.w3.org/Graphics/GIF/spec-gif89a.txt), emphasis mine

TWiStErRob:
A good source to double-check this is http://www.imagemagick.org/Usage/anim_basics/

TWiStErRob:
Creating a branch locally where every current PR related to GIF is merged:
 * ![](http://www.imagemagick.org/Usage/anim_basics/anim_bgnd.gif)  
leaves a white trace instead of transparent
 * ![](http://www.imagemagick.org/Usage/anim_basics/bunny_anim.gif)  
frames drawn over each other instead of after another

TWiStErRob:
I made progress so the bunny is displayed correctly.
I also applied my hunch in the initial comment and it seems everything is working correctly.
More tests need to be done to confirm that partial "dispose to previous" is correct.
If anyone wants to picks this up, here are the changes: https://github.com/TWiStErRob/glide/commit/c8c331ed60b0a8e5133aa4c0e29780c45afdde17

---

The other image seems to be correct according to my reading of the spec, but it seems that browsers and other image software agree on "background" always meaning transparent and not `bgColor`.

sjudd:
Was this fixed by #2521?

michaeimm:
#2521 only fixed that bunny GIF.

michaeimm:
This seems be fixed by #3002 .

sjudd:
Great thanks!

PR #1323, opened 2016-07-10
Forward port GIF fixes

### LINKED ISSUE #1059: The Gif load problem, can't display well

### LINKED ISSUE #1068: black square background appears under the transparent gif.

### LINKED ISSUE #1277: GifResourceDecoder leaks activity once in memory cache

TWiStErRob:
## Description
 * Forward port PR #1281 which fixed #1277
 * Forward port PR #1090 which fixed #1068
 * Forward port PR #1093 which fixed #1059

## Motivation and Context
3.8.0 has better GIF support than 4.0, forward port fixes

---

@sjudd There's still an outstanding issue though: the transparency calculation based on first frame (`isFirstFrameTransparent`) just doesn't imply that the rest of the frames have the same transparency. For example, http://www.imagemagick.org/Usage/anim_basics/canvas_bgnd.gif starts with an opaque frame and then the others keep clearing to transparent color. Actually none of the frames in this GIF have transparency (in header), because the frames are just little square images, but the disposal is background which clears to transparent. You can find the description of this at http://www.imagemagick.org/Usage/anim_basics/#background. Incorrectly using 565 Bitmaps negates some of the fixes above. Not sure what to do with this.
You can find runnable examples with expectations in https://github.com/TWiStErRob/glide-support for #1094

### DIFF
--- a/library/src/main/java/com/bumptech/glide/load/resource/gif/ByteBufferGifDecoder.java
+++ b/library/src/main/java/com/bumptech/glide/load/resource/gif/ByteBufferGifDecoder.java
@@ -66,7 +66,7 @@ public class ByteBufferGifDecoder implements ResourceDecoder<ByteBuffer, GifDraw
       ArrayPool arrayPool,
       GifHeaderParserPool parserPool,
       GifDecoderFactory gifDecoderFactory) {
-    this.context = context;
+    this.context = context.getApplicationContext();
     this.bitmapPool = bitmapPool;
     this.gifDecoderFactory = gifDecoderFactory;
     this.provider = new GifBitmapProvider(bitmapPool, arrayPool);
--- a/third_party/gif_decoder/src/main/java/com/bumptech/glide/gifdecoder/GifHeader.java
+++ b/third_party/gif_decoder/src/main/java/com/bumptech/glide/gifdecoder/GifHeader.java
@@ -33,7 +33,6 @@ public class GifHeader {
   int bgIndex;
   // Pixel aspect ratio.
   int pixelAspect;
-  //TODO: this is set both during reading the header and while decoding frames...
   int bgColor;
   int loopCount;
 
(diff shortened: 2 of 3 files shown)


sjudd:
LGTM, thanks!

isFirstFrameTransparent was always a bit of a hack. We ought to be able to be more accurate about this since we iterate over every pixel anyway. I can't actually remember why it's written the way it is...

PR #1096, opened 2016-03-26
.asBitmap().crossFade()

TWiStErRob:
## Description
Allow `.animate(factory)`, `.asBitmap().crossFade()`, `.asBitmap().transcode(...).crossFade()`.
This opens up the API to allow custom transformations via factory (e.g. `PaletteBitmap`), as well as stock usage of `.crossFade()` in most cases: now default, `asGif` and `asBitmap` all allow cross-fading.
Closes #1007 (v4 already done that) and fixes #840 for v3 (v4 TODO)

## Motivation and Context
It has been asked many times (#840, #605, #903, #847, #1007, #63, #155, #261, #362, #405, #749, #1083, [UPq4H4009OY](https://groups.google.com/d/msg/glidelibrary/UPq4H4009OY/v84MFKD7yYsJ)) how to do this and I figured out a few workarounds, this incorporates those into a builtin method to ease use while also allowing more customization.

---
Questions to @sjudd:
* should I squash?
* should we set .crossFade() as default for .asBitmap() like it is for drawables?

PR #1324, opened 2016-07-10
Forward port PR #1096: .asBitmap().transition(withCrossFade())

### LINKED ISSUE #840: Change format for specific loads and still get transition drawable
Is it possible to change the load format between ARGB and RGB_565 for specific loads and still load a drawable to get the transition effect between thumbnail and actual image?

TWiStErRob:
Forward port PR #1096 which fixes #840.

The v4 port contains less tricks than the v3 version due to simplified request builders.
The Bitmap related TransitionFactory classes also got cleaner responsibilities (fully decoupled from crossfade).
`animate(GlideAnimationFactory)`, or as it's called in v4: `TransitionOptions.transition(TransitionFactory)` was already public.

Here's a demo of new methods introduced. Irrelevant parts (`load`/`into`/etc.) are omitted for brevity.
The class names are pulled to comments to show how it would look like with static imports.

```java
DrawableCrossFadeFactory.Builder fadeOver = new DrawableCrossFadeFactory.Builder()
    .setCrossFadeEnabled(false);

// mirrored convenience methods from DrawableTransitionOptions
Glide.with(this).asBitmap()
    .transition(withCrossFade(3000)) // BitmapTransitionOptions
;
// mirrored convenience methods from DrawableTransitionOptions 
Glide.with(this).asBitmap()
    .transition(withCrossFade(android.R.anim.fade_in, 3000)) // BitmapTransitionOptions
;
// wrap Drawable crossFade over a Bitmap
Glide.with(this).asBitmap()
    .transition(withCrossFade(fadeOver.build())) // // BitmapTransitionOptions 
;
// wrap Drawable transition over a Bitmap (not necessarily a DrawableCrossFadeFactory)
Glide.with(this).asBitmap()
    .transition(withWrapped(new ViewAnimationFactory<Drawable>(android.R.anim.fade_in))) // BitmapTransitionOptions
;
// factories for custom TransitionOptions.transition(*)
Glide.with(this)
    .as(PaletteBitmap.class)
    .transition(with(new PaletteBitmapTransitionFactory(fadeOver.build()))) // GenericTransitionOptions
;
// mirrored typed factory for custom TransitionOptions.transition(*)
Glide.with(this)
    .asBitmap()
    .transition(with(new ViewAnimationFactory<Bitmap>(android.R.anim.fade_in))) // BitmapTransitionOptions
;
// mirrored typed factory for custom TransitionOptions.transition(*)
Glide.with(this)
    .asDrawable()
    .transition(with(new ViewAnimationFactory<Drawable>(android.R.anim.fade_in))) // DrawableTransitionOptions
;
```

Idea nicely finds all the factory methods, all the user has to remember is `transition(with|)` and press Ctrl+Space twice in AS/IDEA to get method suggestions:
image

### DIFF
--- a/library/src/main/java/com/bumptech/glide/request/transition/BitmapContainerTransitionFactory.java
+++ b/library/src/main/java/com/bumptech/glide/request/transition/BitmapContainerTransitionFactory.java
@@ -1,0 +1,55 @@
+package com.bumptech.glide.request.transition;
+
+import android.content.res.Resources;
+import android.graphics.Bitmap;
+import android.graphics.drawable.BitmapDrawable;
+import android.graphics.drawable.Drawable;
+
+import com.bumptech.glide.load.DataSource;
+
+/**
+ * A {@link TransitionFactory} for complex types that have a {@link android.graphics.Bitmap} inside.
+ * The transitioning bitmap is wrapped in a {@link android.graphics.drawable.BitmapDrawable}.
+ * Most commonly used with {@link DrawableCrossFadeFactory}.
+ *
+ *  @param <R> The type of the composite object that contains the {@link android.graphics.Bitmap} to
+ *            be transitioned.
+ */
+public abstract class BitmapContainerTransitionFactory<R> implements TransitionFactory<R> {
+  private final TransitionFactory<Drawable> realFactory;
+
+  public BitmapContainerTransitionFactory(TransitionFactory<Drawable> realFactory) {
+    this.realFactory = realFactory;
+  }
+
+  @Override
+  public Transition<R> build(DataSource dataSource, boolean isFirstResource) {
+    Transition<Drawable> transition = realFactory.build(dataSource, isFirstResource);
+    return new BitmapGlideAnimation(transition);
+  }
+
+  /**
+   * Retrieve the Bitmap from a composite object.
+   * <br>
+   * <b>Warning:</b> Do not convert any arbitrary object to Bitmap via expensive drawing here.
+   *
+   * @param current composite object containing a Bitmap and some other information
+   * @return the Bitmap contained within {@code current}
+   */
+  protected abstract Bitmap getBitmap(R current);
+
+  private class BitmapGlideAnimation implements Transition<R> {
+    private final Transition<Drawable> transition;
+
+    public BitmapGlideAnimation(Transition<Drawable> transition) {
+      this.transition = transition;
+    }
+
+    @Override
+    public boolean transition(R current, ViewAdapter adapter) {
+      Resources resources = adapter.getView().getResources();
+      Drawable currentBitmap = new BitmapDrawable(resources, getBitmap(current));
+      return transition.transition(currentBitmap, adapter);
+    }
+  }
+}
--- a/library/src/main/java/com/bumptech/glide/GenericTransitionOptions.java
+++ b/library/src/main/java/com/bumptech/glide/GenericTransitionOptions.java
@@ -1,5 +1,8 @@
 package com.bumptech.glide;
 
+import com.bumptech.glide.request.transition.TransitionFactory;
+import com.bumptech.glide.request.transition.ViewPropertyTransition;
+
 /**
  * Implementation of {@link TransitionOptions} that exposes only generic methods that can be applied
  * to any resource type.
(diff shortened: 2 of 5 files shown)


library/src/main/java/com/bumptech/glide/request/transition/BitmapContainerTransitionFactory.java
```diff
@@ -0,0 +1,55 @@
...
+  @Override
+  public Transition<R> build(DataSource dataSource, boolean isFirstResource) {
+    Transition<Drawable> transition = realFactory.build(dataSource, isFirstResource);
+    return new BitmapGlideAnimation(transition);
+  }
+
+  /**
+   * Retrieve the Bitmap from a composite object.
+   * <br>
+   * <b>Warning:</b> Do not convert any arbitrary object to Bitmap via expensive drawing here.
```
sjudd: minor nit: Use <p> instead of <br>

library/src/main/java/com/bumptech/glide/request/transition/BitmapContainerTransitionFactory.java
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * A {@link TransitionFactory} for complex types that have a {@link android.graphics.Bitmap} inside.
+ * The transitioning bitmap is wrapped in a {@link android.graphics.drawable.BitmapDrawable}.
+ * Most commonly used with {@link DrawableCrossFadeFactory}.
+ *
+ *  @param <R> The type of the composite object that contains the {@link android.graphics.Bitmap} to
+ *            be transitioned.
+ */
+public abstract class BitmapContainerTransitionFactory<R> implements TransitionFactory<R> {
```
sjudd: Do you have another use case in mind for this other than BitmapTransitionFactory below?

If not, it might be simpler to combine the two in a concrete class for now. We could extract something more general if there's a use case?

sjudd:
Neat! Just a couple small comments

library/src/main/java/com/bumptech/glide/request/transition/BitmapContainerTransitionFactory.java
TWiStErRob:
```java
class PaletteBitmap {
    Palette palette;
    Bitmap bitmap;
}

Glide
    .with(this)
    .as(PaletteBitmap.class)
    .transition(with(new PaletteBitmapCrossFadeFactory()));

public class PaletteBitmapCrossFadeFactory extends BitmapContainerTransitionFactory<PaletteBitmap> {
    public PaletteBitmapCrossFadeFactory() {
        super(new DrawableCrossFadeFactory.Builder().build());
    }

    @Override
    protected Bitmap getBitmap(PaletteBitmap current) {
        return current.bitmap;
    }
}
```

library/src/main/java/com/bumptech/glide/request/transition/BitmapContainerTransitionFactory.java
TWiStErRob: I'll change it to `<p>` in a subsequent PR, I have a few small fixes like this.
*(Don't forget to put `` ` `` around HTML on GitHub.)*

sjudd: In retrospect, this is one of my favorite nit PR comments ever. 

Thanks for the reminder about wrapping.

sjudd:
Awesome thanks!

library/src/main/java/com/bumptech/glide/request/transition/BitmapContainerTransitionFactory.java
TWiStErRob: Yeah, the above is a funny one. I was throwing a WTF, but luckily I can edit all comments here and was able to peek at the markdown.

library/src/main/java/com/bumptech/glide/request/transition/BitmapContainerTransitionFactory.java
dkanada: @TWiStErRob any chance you know about a more recent solution for crossfades with PaletteBitmap targets? I had originally tried GenericTransitionOptions with a basic fade animation, but it doesn't work with placeholders, which are removed before the animation starts. This solution looked promising (had to make adjustments since it's four years old) but the animation is completely ignored with the above method, even with a custom duration.

TWiStErRob: I think that may be due to the fact that your placeholder and final image are not the same type, so there is no transition between them. I haven't used Glide 4 enough to know more details, it could be a simple behavior change between 3 and 4. Try to look around the factories and targets to see who is removing the placeholder and try to work backwards from there to find how it could be avoided. You could also try to follow your animation object to see where it gets ignored.

dkanada: Thank you so much! It ended up being a stray line of code in `onResourceReady()` for a custom target. I'm now encountering another issue though, and would appreciate any insight if possible. When the images are leave the frame in a RecyclerView and then enter again while scrolling, the placeholder shows up indefinitely. My first theory was that the TransitionDrawable was somehow reset and the initial layer was shown with no animation start. I also thought it might be an issue with the BitmapPool or recycled images. Does either of those sound more plausible based on the symptoms?

dkanada: I was able to find the issue, so I'll add it here for anyone else who might have a similar problem.

```
        @Override
        protected void setResource(PaletteBitmap resource) {
            view.setImageBitmap(resource.bitmap);
        }
```

I had originally moved this line of code to `onResourceReady()` since it was crashing the app in `setResource()` which fixed the crash. It had also broken the animation but I didn't notice that until recently. I realized this might be the issue since it was in your code snippet above, so I added it back with a null check before trying to set the bitmap. This ended up fixing the issue while keeping the animation. I'm still not sure why it will sometimes be null, but that's a problem for another day.

TWiStErRob: Nice. Btw, `null` answer is easy:
https://github.com/bumptech/glide/blob/a3a3a6573addd83a5be031bbd55d29e7514ca446/library/src/main/java/com/bumptech/glide/request/target/ImageViewTarget.java#L56-L96
I've updated the code above.

Issue #1044, opened 2016-03-09
ViewTarget MemoryLeak


DenBond7:
**Glide Version**: com.github.bumptech.glide:glide:3.7.0
**Integration libraries**:

``` gradle
    debugCompile 'com.squareup.leakcanary:leakcanary-android:1.4-beta1'
    compile fileTree(dir: 'libs', include: ['*.jar'])
    compile 'com.android.support:appcompat-v7:23.1.1'
    compile 'com.github.bumptech.glide:glide:3.7.0'
    compile 'com.github.bumptech.glide:okhttp-integration:1.3.1@aar'
    compile 'com.google.code.gson:gson:2.6.2'
    compile 'com.squareup.retrofit:retrofit:1.9.0'
    compile 'com.squareup.okhttp:okhttp:2.5.0'
    compile 'com.squareup.okhttp:okhttp-urlconnection:2.5.0'
    compile 'com.android.support:recyclerview-v7:23.1.1'
    compile 'com.android.support:cardview-v7:23.1.1'
    compile 'com.android.support:design:23.1.1'
```

**Device/Android Version**: Android 4.1.1 tested, Genemotion
**Issue details / Repro steps / Use case background**:
1. Open my app
2. Open an activity where showing RecycelerView with images
3. Then press Back -> Back -> .... Home Screen.
4. After I wait about 1 minute and press to run GC. 
5. I have leakcanary.  I saw next memory leak.

![image](https://cloud.githubusercontent.com/assets/2863246/13631264/6c4ae8ba-e5ec-11e5-8f3d-80903f11c2ce.png)
1. Then I collected glide.hprof and saw a memory leak. 
   _**(SELECT \* FROM instanceof android.support.v7.app.AppCompatActivity)**_

![2016-03-09_11-40-04](https://cloud.githubusercontent.com/assets/2863246/13631155/cadca496-e5eb-11e5-90b2-3db41af76f3f.png)

**Glide load line**:
Next code running in RecyclerView.Adapter.

``` java
Glide.with(context)
        .load(images.get(i))
        .error(android.R.drawable.ic_delete)
        .centerCrop()
        .into(viewHolder.imageViewCar);
```

context - GalleryFragment -> original methot getContext() from android.support.v4.app.Fragment;

**Layout XML**:

``` xml
<?xml version="1.0" encoding="utf-8"?>
<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
[...]

Issue #616, opened 2015-09-10
Memory leak related to Glide


RomainPiel:
```
In com.example:2.0:45.
* com.example.ArtistActivity has leaked:
* GC ROOT thread com.bumptech.glide.load.engine.executor.FifoPriorityThreadPoolExecutor$DefaultThreadFactory$1.<Java Local> (named 'fifo-pool-thread-0')
* references com.bumptech.glide.load.engine.executor.FifoPriorityThreadPoolExecutor$LoadTask.sync
* references java.util.concurrent.FutureTask$Sync.callable
* references java.util.concurrent.Executors$RunnableAdapter.task
* references com.bumptech.glide.load.engine.EngineRunnable.manager
* references com.bumptech.glide.load.engine.EngineJob.cbs
* references java.util.ArrayList.array
* references array java.lang.Object[].[0]
* references com.bumptech.glide.request.GenericRequest.target
* references com.bumptech.glide.request.target.GlideDrawableImageViewTarget.view
* references android.widget.ImageView.mContext
* references android.support.v7.internal.view.ContextThemeWrapper.mBase
* leaks com.example.ArtistActivity instance
```

Leak Canary just found that memory leak in my app and it looks like it's related to Glide.
This call is not doing anything fancy:
```Java
Glide.with(heroImage.getContext())
                .load("http://....")
                .into(heroImage);
```

I only reproduced the issue on API 16.

Any ideas?

TWiStErRob:
Glide should be hooking into the activity life cycle, if Leak Canary does the same, there may be an ordering issue: that is Leak Canary checks too soon, before Glide has the chance to clean up in response to a stopped activity.

If you can reproduce consistently, you could check a heap dump few seconds after you get this message to prove that the leak is real and not a false positive.

LiHanGuang:
@TWiStErRob 
@RomainPiel
I also have this problem,  but i solve it.

Solution:
MyActivity.java
onDestory(){
        Glide.get(this).clearMemory(); //  call this method manual 
}

Question:
"Glide should be hooking into the activity life cycle",
Is Glide auto call "clearMemory()" when the callback method "onDestory" be called?

TWiStErRob:
[...]

PR #1433, opened 2016-08-29
Prevent leaking activities due to a call without context

raycoarana:
## Description
We detect some memory leaks due to the ViewTarget being leaking the target view. If glide is called without a context, this ViewTarget is never cancelled and bad things happens. This is a change to prevent that to happen.

## Motivation and Context
Due to a bad use of the library or because of architecture restrictions, you may be calling Glide with a ViewTarget but without any activity or fragment context, so your view could be leaked and with it, the fragment and the activity that contains it. With this change, making ViewTarget to handle the view as a WeakReference, prevent it from doing any kind of leak, no matter how the library is used.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/request/target/ViewTarget.java
+++ b/library/src/main/java/com/bumptech/glide/request/target/ViewTarget.java
@@ -44,11 +44,11 @@ public abstract class ViewTarget<T extends View, Z> extends BaseTarget<Z> {
   private static boolean isTagUsedAtLeastOnce = false;
   @Nullable private static Integer tagId = null;
 
-  protected final T view;
+  protected final WeakReference<T> viewReference;
   private final SizeDeterminer sizeDeterminer;
 
   public ViewTarget(T view) {
-    this.view = Preconditions.checkNotNull(view);
+    this.viewReference = new WeakReference<T>(Preconditions.checkNotNull(view));
     sizeDeterminer = new SizeDeterminer(view);
   }
 
@@ -56,7 +56,7 @@ public abstract class ViewTarget<T extends View, Z> extends BaseTarget<Z> {
    * Returns the wrapped {@link android.view.View}.
    */
   public T getView() {
-    return view;
+    return viewReference.get();
   }
 
   /**
@@ -118,10 +118,20 @@ public abstract class ViewTarget<T extends View, Z> extends BaseTarget<Z> {
 
   @Override
   public String toString() {
-    return "Target for: " + view;
+    View view = viewReference.get();
+    if (view == null) {
+      return "Target for an already destroyed view.";
+    } else {
+      return "Target for: " + view;
+    }
   }
 
   private void setTag(@Nullable Object tag) {
+    View view = viewReference.get();
+    if (view == null) {
+      return;
+    }
+
     if (tagId == null) {
       isTagUsedAtLeastOnce = true;
       view.setTag(tag);
@@ -132,6 +142,10 @@ public abstract class ViewTarget<T extends View, Z> extends BaseTarget<Z> {
 
   @Nullable
   private Object getTag() {
+    View view = viewReference.get();
+    if (view == null) {
+      return null;
+    }
     if (tagId == null) {
       return view.getTag();
     } else {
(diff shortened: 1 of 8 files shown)


library/src/main/java/com/bumptech/glide/request/transition/BitmapContainerTransitionFactory.java
```diff
@@ -46,9 +48,14 @@ public BitmapGlideAnimation(Transition<Drawable> transition) {
...
     @Override
     public boolean transition(R current, ViewAdapter adapter) {
-      Resources resources = adapter.getView().getResources();
-      Drawable currentBitmap = new BitmapDrawable(resources, getBitmap(current));
-      return transition.transition(currentBitmap, adapter);
+      View view = adapter.getView();
+      if (view == null) {
+        return false;
+      } else {
+        Resources resources = adapter.getView().getResources();
```
TWiStErRob: Is this safe? What if GC kicks in while checking `view == null`?

TWiStErRob:
I have mixed feelings. It adds a lot of nullity to the code and breaks every single custom target out there in the wild. On the other hand it's a positive change for preventing leaks and a very good observation.

@sjudd & @raycoarana Wouldn't it be possible to keep the target as it was and change Glide's internals to hold weak Targets? That way the public interface would be kept simple. What if `SingleRequest.target` was weak instead? Or if that could cause a problem (e.g. with downloadOnly), keep it strong while the load is queued/in progress and weaken it only after the resource has been delivered in `onResourceReady`.

sjudd:
You can't hold on to Targets weekly without breaking pretty much every Target that is not a ViewTarget (SimpleTarget, notification targets etc). 

As is this will also cause NPEs for custom ViewTargets, so it's probably not something we can accept. 

Typically when Targets are leaked, it's only for the duration of the load, which will eventually finish, often relatively quickly. Did you find a case where Targets are leaked permanently?

library/src/main/java/com/bumptech/glide/request/target/ViewTarget.java
```diff
@@ -44,19 +44,19 @@
   private static boolean isTagUsedAtLeastOnce = false;
   @Nullable private static Integer tagId = null;
 
-  protected final T view;
+  protected final WeakReference<T> viewReference;
```
TWiStErRob: @sjudd NPE only if extenders use `getView()`, but compile error if they use `super.view`.

TWiStErRob:
Hmm, I thought `Glide.with(appContext).load(blah).into(imageView)` leaks permanently... but I just read that `ApplicationLifecycle` doesn't hold on to the `Target`s the same way `ActivityFragmentLifecycle` does. (Rayco, `Target` is a `LifecycleListener`)

raycoarana:
Mmmm, sorry guys maybe is a false positive of LeakCanary. What we have is this leak:

> GC ROOT com.bumptech.glide.load.engine.executor.FifoPriorityThreadPoolExecutor$DefaultThreadFactory$1.<Java Local> (anonymous subclass of java.lang.Thread)
> references com.bumptech.glide.load.engine.executor.FifoPriorityThreadPoolExecutor$LoadTask.sync
> references java.util.concurrent.FutureTask$Sync.callable
> references java.util.concurrent.Executors$RunnableAdapter.task
> references com.bumptech.glide.load.engine.EngineRunnable.manager
> references com.bumptech.glide.load.engine.EngineJob.cbs
> references java.util.ArrayList.array
> references array java.lang.Object[].[0]
> references com.bumptech.glide.request.GenericRequest.target
> references com.tuenti.glide.GlideMapper$5.view (anonymous subclass of com.bumptech.glide.request.target.BitmapImageViewTarget)
> references com.tuenti.view.PhotoZoomView.mContext
> leaks com.tuenti.PhotoActivity instance

I'm thinking that maybe this is true specially in cases where users have a very slow internet connection and try to load many images, that could leak activities for some time, even crashing the app because of OOM exception.

I will close this PR and further investigate it.

Thanks!

library/src/main/java/com/bumptech/glide/request/target/ViewTarget.java
```diff
@@ -44,19 +44,19 @@
...
   private final SizeDeterminer sizeDeterminer;
 
   public ViewTarget(T view) {
-    this.view = Preconditions.checkNotNull(view);
+    this.viewReference = new WeakReference<T>(Preconditions.checkNotNull(view));
     sizeDeterminer = new SizeDeterminer(view);
```
TWiStErRob: This also has a strong ref.

TWiStErRob:
Also read #616 and #1044.

Issue #1441, opened 2016-08-31
Non-2xx responses leak connections with OkHttp integrations


JakeWharton:
In both the OkHttp 2 and OkHttp 3 integrations on `master`, the `ResponseBody` is not closed when `isSuccessful()` is `false` (aka non-2xx response). The `else` branch should immediately call `response.body().close()` or at the very least set the `responseBody` field for later closing (the former seems more logical).

TWiStErRob:
Good catch, in v3 the sync version assigned `responseBody` every time.

Issue #1234, opened 2016-05-25
Loading from file (Raw image) doesn't resize the loaded Image | OOM 


zetbrush:
**Glide Version**: 4.0.0-SNAPSHOT 20151124.173231-105
**Integration libraries**: no diff
**Device/Android Version**: All/Galaxy S4/GT-I9500 5.0.1

**Issue details / Repro steps / Use case background**: 

When loading raw image from file with for example 2k/4k images, we should override width and height using original width/height which leads to huge memory usages and OOM, as because of loading couple of that size images.

**Glide load line / `GlideModule` (if any) / list Adapter code (if any)**:

``` java
Glide.with().asBitmap().load(file).listener(listener).apply(RequestOptions.dontTransform().override(width, height)).into(imageView);
```

as width and height should be the original size Glide loads as is.

So is there any way to override sizes after decode, so glide could load  the low res and put into the cache the low res one?

TWiStErRob:
Do we have built-in RAW decoders in v4? I don't remember that. If we don't, that means you wrote the decoder.

There are two things that contribute to sizing:

 * If you allow Glide to figure out the size and try to downsample in decoder (this is what the built-in BitmapFactory wrapping Downsampler class does, start from `StreamBitmapDecoder`), you get the target size in `ResourceDecoder.decode(...)` and you can adhere to those requested sizes by loading a close enough amount of pixels from the raw image. E.g. you can load every second row/column, or calculate an average of an NxN grid where `N = max(1, sourceWidth / targetWidth)`, this saves memory outright.
 * After this you can apply a transformation to create an even smaller bitmap, again trying to match the target size, from the loaded one, e.g. `.fitCenter()`. This transformed image will be saved in RESULT cache (I think the equivalent is `RESOURCE` in v4).

I think your problem is that you need to tell the decoder somehow how many bytes are there because the raw file doesn't have a header; and you chose `.override()` to do this. Is this correct?

zetbrush:
yeah, thanks for suggestions, you got me right! will something like this case handling be integrated in Glide?

TWiStErRob:
[...]

PR #1443, opened 2016-09-01
Close the OkHttp response in case of error. Fixes https://github.com/…

kypeli:
## Description
Closes OkHttp and OkHttp3 Responses in case of error. Fixes issue https://github.com/bumptech/glide/issues/1441

### DIFF
--- a/integration/okhttp3/src/main/java/com/bumptech/glide/integration/okhttp3/OkHttpStreamFetcher.java
+++ b/integration/okhttp3/src/main/java/com/bumptech/glide/integration/okhttp3/OkHttpStreamFetcher.java
@@ -52,9 +52,9 @@ public class OkHttpStreamFetcher implements DataFetcher<InputStream> {
 
       @Override
       public void onResponse(Call call, Response response) throws IOException {
+        responseBody = response.body();
         if (response.isSuccessful()) {
           long contentLength = response.body().contentLength();
-          responseBody = response.body();
           stream = ContentLengthInputStream.obtain(responseBody.byteStream(), contentLength);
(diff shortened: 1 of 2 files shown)


kypeli:
Happened to see this trivial issue in Glide, so I hope my PR is useful. Didn't notice any tests related to OkHttp in Glide, but I am very new to the Glide source code, so please correct me if I am wrong. 

Anything else that is needed for this?

integration/okhttp3/src/main/java/com/bumptech/glide/integration/okhttp3/OkHttpStreamFetcher.java
```diff
@@ -58,6 +58,7 @@ public void onResponse(Call call, Response response) throws IOException {
           stream = ContentLengthInputStream.obtain(responseBody.byteStream(), contentLength);
         } else if (Log.isLoggable(TAG, Log.DEBUG)) {
           Log.d(TAG, "OkHttp got error response: " + response.code() + ", " + response.message());
+          response.body().close();
```
TWiStErRob: What if logging is disabled?

TWiStErRob:
Thanks, there's only one issue: it's a partial fix, see comment on code.

I think to keep it compatible (similar) with v3 it would be better to assign `responseBody` as soon as there is one. `cleanup` is called in a ` finally` as far as I know.

kypeli:
You are absolutely right, that's a lot better. Thanks. I updated the pull request as suggested.

TWiStErRob:
Tip for future: you don't need the full url in the commit message `#1234` or `user/repo#1234` works as well.

sjudd:
Thanks!

PR #1459, opened 2016-09-08
return the old MemoryCategory used by Glide.

frannois:
## Description
return the previous `MemoryCategory` used by Glide when setting a new one.

## Motivation and Context
If the app enters a section of the app that would benefit from a boost of Glide's memory, the developer has the option to boost the memory used by Glide.
However there is no way currently to know what was the current `MemoryCategory` used by Glide before entering that high memory usage portion of the app, which makes it impossible to restore the `MemoryCategory` once that portion of the app is done.

Example  I have an image picker RecyclerView activity. I would like to boost Glide's memory in onCreate but I can't restore it to the previous value in the onDestroy of the activity.


frannois:
Another thing we could do is to check current MemoryCategory value before setting the multipliers and evicting.
However that change breaks a test and since it's my first PR I wanted to have the maintainers point of view on this first.

sjudd:
Looks good to me. Setting the same size multiplier repeatedly shouldn't cause any evictions, the caches should already be below the maximum.

Issue #382, opened 2015-03-29
Choose crop location


Kev1000000:
While centerCrop does a great job for most situations, there are still quite a few situations where choosing a focal point where Glide would crop from would be very handy.  For example, I would love to crop starting from the lop left of the image.

TWiStErRob:
I think we've discussed this here: https://groups.google.com/forum/#!search/crop$20face$20glide/glidelibrary/ePa-mpY4doc/ldiAUnuF140J

TWiStErRob:
@sjudd, actually, would it be a big deal to allow something like:
```
Glide.with(...).load(stuff)
        .asBitmap()
        .crop(new Cropper() {
            public Rect findCropPosition(Bitmap bitmap, int targetWidth, int targetHeight) {
                Rect r = new Rect(0, 0, targetWidth, targetHeight);
                r.offset(-targetWidth/2, -targetHeight/2); // center crop-rect around 0, 0
                r.offsetTo(bitmap.getWidth()/2, bitmap.getHeight()/2); // move to middle of bitmap
                return r;
            }
        })
        .into(...)
;
```
The above is simulating centerCrop, but you can imagine a "face-searching" library call there.
Also, for @Kev1000000's top left it would be just: `return new Rect(0, 0, targetWidth, targetHeight);`.

This would mean that Glide supplies the `.crop()` method, a `CropTransformation` and a `CropTransformation.Cropper` interface. What do you think?

Kev1000000:
Does this mean starting a crop at 0,0 us not currently possible?

TWiStErRob:
You just have to roll your own transformation, extend `BitmapTransformation` and then take a look at what `CenterCrop` does, you can modify it to your needs. The code in `CenterCrop` helps to re-use `Bitmap`s, and `TransformationUtils.centerCrop` does the actual cropping, that's the core you need to modify, but it'll be much simple since you're bound to 0, 0.

Then after you wrote your own `TopLeftCrop`, you can just add `.transform(new TopLeftCrop())` to your Glide line. I suggest to use a singleton pattern like `UnitTransformation` to avoid creating a new stateless object every so often.

sjudd:
This is actually relatively straight forward. The approach @TWiStErRob suggested is correct, subclass [BitmapTransformation][1] and write whatever transformation you'd like. You can base it on the existing [CenterCrop][2] if that's useful.

[...]

PR #1491, opened 2016-09-28
[WIP] Added PositionedCrop Feature as requested in #382

bjornson:
## Description

This adds the feature requested in requested in #382 
CenterCrop now inherits from PositionedCrop.
## Motivation and Context

The ability to position the crop area in addition to the CenterCrop Transformation. Using a percentage allows for many different usages.

The tests were succesful.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/load/resource/bitmap/CenterCrop.java
+++ b/library/src/main/java/com/bumptech/glide/load/resource/bitmap/CenterCrop.java
@@ -1,9 +1,7 @@
 package com.bumptech.glide.load.resource.bitmap;
 
 import android.content.Context;
-
 import com.bumptech.glide.load.engine.bitmap_recycle.BitmapPool;
-
 import java.security.MessageDigest;
 
 /**
@@ -18,16 +16,16 @@ public class CenterCrop extends PositionedCrop {
   private static final byte[] ID_BYTES = ID.getBytes(CHARSET);
 
   public CenterCrop(Context context) {
-    super(context, 0.5f, 0.5f);
+    super(context, PositionedCrop.CENTER, PositionedCrop.CENTER);
   }
 
   public CenterCrop(BitmapPool bitmapPool) {
-    super(bitmapPool, 0.5f, 0.5f);
+    super(bitmapPool, PositionedCrop.CENTER, PositionedCrop.CENTER);
   }
 
   @Override
   public boolean equals(Object o) {
-    return o instanceof CenterCrop;
+    return o != null && getClass() == o.getClass();
   }
 
   @Override
--- a/library/src/main/java/com/bumptech/glide/load/resource/bitmap/CenterCrop.java
+++ b/library/src/main/java/com/bumptech/glide/load/resource/bitmap/CenterCrop.java
@@ -1,9 +1,9 @@
 package com.bumptech.glide.load.resource.bitmap;
 
 import android.content.Context;
-import android.graphics.Bitmap;
-import android.support.annotation.NonNull;
+
 import com.bumptech.glide.load.engine.bitmap_recycle.BitmapPool;
+
 import java.security.MessageDigest;
 
 /**
@@ -13,24 +13,16 @@ import java.security.MessageDigest;
  *
  * Does not maintain the image's aspect ratio
  */
-public class CenterCrop extends BitmapTransformation {
+public class CenterCrop extends PositionedCrop {
   private static final String ID = "com.bumptech.glide.load.resource.bitmap.CenterCrop";
   private static final byte[] ID_BYTES = ID.getBytes(CHARSET);
 
   public CenterCrop(Context context) {
-    super(context);
+    super(context, 0.5f, 0.5f);
   }
 
   public CenterCrop(BitmapPool bitmapPool) {
-    super(bitmapPool);
-  }
-
-  // Bitmap doesn't implement equals, so == and .equals are equivalent here.
-  @SuppressWarnings("PMD.CompareObjectsWithEquals")
-  @Override
-  protected Bitmap transform(@NonNull BitmapPool pool, @NonNull Bitmap toTransform, int outWidth,
-      int outHeight) {
-    return TransformationUtils.centerCrop(pool, toTransform, outWidth, outHeight);
+    super(bitmapPool, 0.5f, 0.5f);
   }
 
   @Override
--- a/library/src/main/java/com/bumptech/glide/load/resource/bitmap/PositionedCrop.java
+++ b/library/src/main/java/com/bumptech/glide/load/resource/bitmap/PositionedCrop.java
@@ -1,0 +1,80 @@
+package com.bumptech.glide.load.resource.bitmap;
+
+import static android.os.Build.ID;
+
+import android.content.Context;
+import android.graphics.Bitmap;
+import android.support.annotation.FloatRange;
+
+import com.bumptech.glide.load.engine.bitmap_recycle.BitmapPool;
+
+import java.security.MessageDigest;
+
+/**
+ * Scale the image so that either the width of the image matches the given width and the height of
+ * the image is greater than the given height or vice versa, and then crop the larger dimension to
+ * match the given dimension.
+ *
+ * Using percentages the crop area can be adjusted
+ *
+ * Example usage:
+ *
+ * Crop top-left: new PositionedCrop(context, 0, 0);
+ *
+ * Crop top-right: new PositionedCrop(context, 0, 1);
+ *
+ * Crop bottom-right: new PositionedCrop(context, 1, 1);
+ *
+ * Crop top-center: new PositionedCrop(context, 0,5f, 0);
+ *
+ * Does not maintain the image's aspect ratio
+ */
+public class PositionedCrop extends BitmapTransformation {
+  private static final byte[] ID_BYTES = ID.getBytes(CHARSET);
+
+  private float xPercentage = 0.5f;
+  private float yPercentage = 0.5f;
+
+  public PositionedCrop(Context context, @FloatRange(from = 0.0, to = 1.0) float xPercentage, @FloatRange(from = 0.0, to = 1.0) float yPercentage) {
+    super(context);
+    this.xPercentage = xPercentage;
+    this.yPercentage = yPercentage;
+  }
+
+  public PositionedCrop(BitmapPool bitmapPool, @FloatRange(from = 0.0, to = 1.0) float xPercentage, @FloatRange(from = 0.0, to = 1.0) float yPercentage) {
+    super(bitmapPool);
+    this.xPercentage = xPercentage;
+    this.yPercentage = yPercentage;
+  }
+
+  // Bitmap doesn't implement equals, so == and .equals are equivalent here.
+  @SuppressWarnings("PMD.CompareObjectsWithEquals")
+  @Override
+  protected Bitmap transform(BitmapPool pool, Bitmap toTransform, int outWidth, int outHeight) {
+
+    return TransformationUtils.cropPosition(pool, toTransform, outWidth, outHeight, xPercentage, yPercentage);
+  }
+
+  public String getId() {
+    return "PositionedCrop.com.bumptech.glide.load.resource.bitmap.x:" + xPercentage + ".y:" + yPercentage;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (o instanceof PositionedCrop) {
+      return ((PositionedCrop) o).xPercentage == xPercentage && ((PositionedCrop) o).yPercentage == yPercentage;
+    } else {
+      return false;
+    }
+  }
+
+  @Override
+  public int hashCode() {
+    return getId().hashCode();
+  }
+
+  @Override
+  public void updateDiskCacheKey(MessageDigest messageDigest) {
+    messageDigest.update(ID_BYTES);
+  }
+}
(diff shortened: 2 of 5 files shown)


bjornson:
I signed it!

library/src/main/java/com/bumptech/glide/load/resource/bitmap/TransformationUtils.java
```diff
@@ -98,6 +98,49 @@ public static Bitmap centerCrop(@NonNull BitmapPool pool, @NonNull Bitmap inBitm
...
+   */
+  public static Bitmap cropPosition(@NonNull BitmapPool pool, @NonNull Bitmap inBitmap, int width,
+      int height, float xPercentage, float yPercentage) {
+    if (inBitmap.getWidth() == width && inBitmap.getHeight() == height) {
+      return inBitmap;
+    }
+    // From ImageView/Bitmap.createScaledBitmap.
+    final float scale;
+    float dx = 0, dy = 0;
+    Matrix m = new Matrix();
```
TWiStErRob: There are an awful lot of `0.5`s hardcoded here, it feel suspicious with 0.5 meaning center, but also having x/yPercentage.

library/src/main/java/com/bumptech/glide/load/resource/bitmap/TransformationUtils.java
```diff
@@ -181,11 +224,11 @@ public static Bitmap centerInside(@NonNull BitmapPool pool, @NonNull Bitmap inBi
 
   /**
    * Sets the alpha of the Bitmap we're going to re-use to the alpha of the Bitmap we're going to
-   * transform. This keeps {@link android.graphics.Bitmap#hasAlpha()}} consistent before and after
+   * transform. This keeps {@link Bitmap#hasAlpha()}} consistent before and after
```
TWiStErRob: I think Sam prefers FQCNs in docs.

library/src/test/java/com/bumptech/glide/load/resource/bitmap/PositionedCropTest.java
```diff
@@ -0,0 +1,149 @@
...
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+import org.robolectric.RobolectricTestRunner;
+import org.robolectric.Shadows;
+import org.robolectric.annotation.Config;
+
+@RunWith(RobolectricTestRunner.class)
+@Config(manifest = Config.NONE, sdk = 18)
+public class PositionedCropTest {
```
TWiStErRob: This doesn't test the logic of the target class. There's nothing that verifies that all possible values work. At least add tests for a 3&times3 grid combinations of values (top, center, left) &times; (left, center, right). I think if you create a 3px&times;3px Bitmap of 9 colors and corp to 1px&times1px you can check the pixel colors to see if the crop was correct.

library/src/main/java/com/bumptech/glide/load/resource/bitmap/CenterCrop.java
```diff
@@ -13,24 +13,16 @@
...
  * Does not maintain the image's aspect ratio
  */
-public class CenterCrop extends BitmapTransformation {
+public class CenterCrop extends PositionedCrop {
   private static final String ID = "com.bumptech.glide.load.resource.bitmap.CenterCrop";
   private static final byte[] ID_BYTES = ID.getBytes(CHARSET);
 
   public CenterCrop(Context context) {
-    super(context);
+    super(context, 0.5f, 0.5f);
```
bjornson: Absolutely correct. I suggest providing these:
```
public static final float START = 0;
public static final float CENTER = 0.5f;
public static final float END = 1;
```

library/src/main/java/com/bumptech/glide/load/resource/bitmap/PositionedCrop.java
```diff
@@ -0,0 +1,80 @@
...
+ * Crop top-right: new PositionedCrop(context, 0, 1);
+ *
+ * Crop bottom-right: new PositionedCrop(context, 1, 1);
+ *
+ * Crop top-center: new PositionedCrop(context, 0,5f, 0);
+ *
+ * Does not maintain the image's aspect ratio
+ */
+public class PositionedCrop extends BitmapTransformation {
+  private static final byte[] ID_BYTES = ID.getBytes(CHARSET);
```
bjornson: Yes, this should use the dynamic id from `getId()`

library/src/main/java/com/bumptech/glide/load/resource/bitmap/PositionedCrop.java
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  public PositionedCrop(BitmapPool bitmapPool, @FloatRange(from = 0.0, to = 1.0) float xPercentage, @FloatRange(from = 0.0, to = 1.0) float yPercentage) {
+    super(bitmapPool);
+    this.xPercentage = xPercentage;
+    this.yPercentage = yPercentage;
+  }
+
+  // Bitmap doesn't implement equals, so == and .equals are equivalent here.
+  @SuppressWarnings("PMD.CompareObjectsWithEquals")
```
bjornson: :-) ok, this will be removed

library/src/main/java/com/bumptech/glide/load/resource/bitmap/PositionedCrop.java
```diff
@@ -0,0 +1,80 @@
...
+  @Override
+  protected Bitmap transform(BitmapPool pool, Bitmap toTransform, int outWidth, int outHeight) {
+
+    return TransformationUtils.cropPosition(pool, toTransform, outWidth, outHeight, xPercentage, yPercentage);
+  }
+
+  public String getId() {
+    return "PositionedCrop.com.bumptech.glide.load.resource.bitmap.x:" + xPercentage + ".y:" + yPercentage;
```
bjornson: Good idea, this is more elegant:
`...bitmap(x=" + xPercentage + ",y=" + yPercentage + ")";`

library/src/main/java/com/bumptech/glide/load/resource/bitmap/PositionedCrop.java
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  @Override
+  public int hashCode() {
+    return getId().hashCode();
+  }
+
+  @Override
+  public void updateDiskCacheKey(MessageDigest messageDigest) {
+    messageDigest.update(ID_BYTES);
```
bjornson: Yes, I will change it to the dynamic
`messageDigest.update(getId().getBytes(CHARSET));`

library/src/main/java/com/bumptech/glide/load/resource/bitmap/TransformationUtils.java
```diff
@@ -98,6 +98,49 @@ public static Bitmap centerCrop(@NonNull BitmapPool pool, @NonNull Bitmap inBitm
...
+   *
+   * @param pool     The BitmapPool to obtain a bitmap from.
+   * @param inBitmap   The Bitmap to resize.
+   * @param width    The width in pixels of the final Bitmap.
+   * @param height   The height in pixels of the final Bitmap.
+   * @param xPercentage The horizontal percentage of the crop. 0.0f => left, 0.5f => center, 1.0f => right or anything in between 0 and 1
+   * @param yPercentage The vertical percentage of the crop. 0.0f => top, 0.5f => center, 1.0f => bottom or anything in between 0 and 1
+   * @return The resized Bitmap (will be recycled if recycled is not null).
+   */
+  public static Bitmap cropPosition(@NonNull BitmapPool pool, @NonNull Bitmap inBitmap, int width,
```
bjornson: Good idea!

library/src/main/java/com/bumptech/glide/load/resource/bitmap/TransformationUtils.java
bjornson: There was one set too many which would have resulted in wrong results.
```

dy = (height - inBitmap.getHeight() * scale);
dy *= yPercentage;
```

library/src/main/java/com/bumptech/glide/load/resource/bitmap/TransformationUtils.java
```diff
@@ -98,6 +98,49 @@ public static Bitmap centerCrop(@NonNull BitmapPool pool, @NonNull Bitmap inBitm
...
+      int height, float xPercentage, float yPercentage) {
+    if (inBitmap.getWidth() == width && inBitmap.getHeight() == height) {
+      return inBitmap;
+    }
+    // From ImageView/Bitmap.createScaledBitmap.
```
bjornson: Regarding the comment: what would be the best practice? Remove the comment or add more detail to it that explains what was modified from the original createScaledBitmap?
Regarding the duplication: The centerCrop should call cropPosition with (..., CENTER, CENTER), it is exactly the same since cropPosition only replaced the original 0.5f parts with the percentage variables. In my first commit it was doubled, which was wrong.

library/src/test/java/com/bumptech/glide/load/resource/bitmap/PositionedCropTest.java
bjornson: Good idea. I will need some time for this as I am not very experienced in writing tests.

library/src/main/java/com/bumptech/glide/load/resource/bitmap/TransformationUtils.java
bjornson: Ok, this was done by the auto-formatter. Is there a way to import the default rules that are used in the glide project?

bjornson:
The new test is still missing, so this is [WIP]

library/src/main/java/com/bumptech/glide/load/resource/bitmap/CenterCrop.java
TWiStErRob: Hmm, good idea; I was only thinking locally in CenterCrop.

I would expand to left/right, top/bottom, and maybe center/middle, because start-end is misleading. There's no reason to abstract it away, especially since Gravity already has these start/end names, which are direction-sensitive, but these constants are not that!

In the future (in a later PR or if you have a lot of time) it would be nice to have a specialized constructor that accepts a Gravity `int` and decompose that to the corresponding float values.

library/src/main/java/com/bumptech/glide/load/resource/bitmap/PositionedCrop.java
TWiStErRob: I think the value should be cached in a field (percents are final, so id doesn't change) to prevent multiple StringBuilder allocations and charset lookups per load.

library/src/main/java/com/bumptech/glide/load/resource/bitmap/TransformationUtils.java
TWiStErRob: That looks more reasonable. This is why exhaustive tests are needed.

library/src/main/java/com/bumptech/glide/load/resource/bitmap/TransformationUtils.java
TWiStErRob: There should be one included: https://github.com/bumptech/glide/tree/master/.idea

The code formatter of IDEA is not perfect, many times it has its own unconfigurable rules. Most noticed during refactors when static imports and inner class imports are freely changed.

What I usually do is do all the coding, use the formatter, and then while committing and reviewing my diff locally I just revert the code blocks I didn't specifically change. This prevents those weird diffs like you have after your second commit with the space alignments. They're not part of this PR and just confusing if later the commit diff is reviewed. I suggest you revert those.

library/src/test/java/com/bumptech/glide/load/resource/bitmap/PositionedCropTest.java
TWiStErRob: Take your time, there's no rush.

library/src/main/java/com/bumptech/glide/load/resource/bitmap/PositionedCrop.java
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Scale the image so that either the width of the image matches the given width and the height of
+ * the image is greater than the given height or vice versa, and then crop the larger dimension to
+ * match the given dimension.
+ *
+ * Using percentages the crop area can be adjusted
+ *
+ * Example usage:
+ *
+ * Crop top-left: new PositionedCrop(context, 0, 0);
```
TWiStErRob: Update the examples to use the constants, especially `0,5f`!
Hmm, actually, after having all the constants, do we really need to say this?

> "Crop top-left: `new PositionedCrop(context, LEFT, TOP)`"

---

Note: this is something I've never understood: if there's a new line between each line then there's no benefit of having the new lines, because they evenly space things out, hence provide no grouping of source code lines. If people prefer bigger line-height than normal, they should set it in their browser/text editor, not in the source code.

TWiStErRob:
Thanks for taking the time to do this!

library/src/main/java/com/bumptech/glide/load/resource/bitmap/CenterCrop.java
TWiStErRob: Hmm, I think I figured out where you got start/end from: fitStart/fitEnd (`Matrix.ScaleToFit`).

sjudd:
Assuming this is stale, happy to continue the review if anyone wants to pick this up.

bjornson:
I am really busy at the moment, but would like to update it this summer. But as this was for an older version, I might have to redo the whole thing.

sp00ne:
Any updates on this? :) Has it gone completely stale or is something similar already out? 🙏

Issue #1555, opened 2016-10-28
Please remove synthetic methods


RiteshChandnani:
**Glide Version**: 3.7.0

**Issue details**: 
There’s no true nesting in Java. Hence, any access to private members of outer class from nested class will have to go via a package scoped method generated by Java to facilitate the call.

To reduce method, make methods of outer class package scoped.

More information:
https://realm.io/news/360andev-jake-wharton-java-hidden-costs-android/

TWiStErRob:
Yep, I watched that too, agree, it's a good one.

Feel free to implement a PR that increases some scopes where the IDEA inspection flags there's a synthetic accessor, I think @sjudd will be happy to accept it.

PR #1556, opened 2016-10-28
Removed synthetic methods

RiteshChandnani:
## Description

This removes all the synthetic methods.
Issue #1555 
## Motivation and Context

This change helps us reduce the method count and therefore also helps us avoid the trampoline created because of synthetic methods.

### DIFF
--- a/integration/okhttp/src/main/java/com/bumptech/glide/integration/okhttp/OkHttpStreamFetcher.java
+++ b/integration/okhttp/src/main/java/com/bumptech/glide/integration/okhttp/OkHttpStreamFetcher.java
@@ -6,6 +6,7 @@ import com.bumptech.glide.load.DataSource;
 import com.bumptech.glide.load.data.DataFetcher;
 import com.bumptech.glide.load.model.GlideUrl;
 import com.bumptech.glide.util.ContentLengthInputStream;
+import com.bumptech.glide.util.Synthetic;
 import com.squareup.okhttp.OkHttpClient;
 import com.squareup.okhttp.Request;
 import com.squareup.okhttp.Response;
@@ -24,8 +25,8 @@ public class OkHttpStreamFetcher implements DataFetcher<InputStream> {
   private static final String TAG = "OkHttpFetcher";
   private final OkHttpClient client;
   private final GlideUrl url;
-  private InputStream stream;
-  private ResponseBody responseBody;
+  @Synthetic InputStream stream;
+  @Synthetic ResponseBody responseBody;
 
   public OkHttpStreamFetcher(OkHttpClient client, GlideUrl url) {
     this.client = client;
--- a/integration/okhttp3/src/main/java/com/bumptech/glide/integration/okhttp3/OkHttpStreamFetcher.java
+++ b/integration/okhttp3/src/main/java/com/bumptech/glide/integration/okhttp3/OkHttpStreamFetcher.java
@@ -7,6 +7,7 @@ import com.bumptech.glide.load.DataSource;
 import com.bumptech.glide.load.data.DataFetcher;
 import com.bumptech.glide.load.model.GlideUrl;
 import com.bumptech.glide.util.ContentLengthInputStream;
+import com.bumptech.glide.util.Synthetic;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.Map;
@@ -22,8 +23,8 @@ public class OkHttpStreamFetcher implements DataFetcher<InputStream> {
   private static final String TAG = "OkHttpFetcher";
   private final Call.Factory client;
   private final GlideUrl url;
-  private InputStream stream;
-  private ResponseBody responseBody;
+  @Synthetic InputStream stream;
+  @Synthetic ResponseBody responseBody;
   private volatile Call call;
 
   public OkHttpStreamFetcher(Call.Factory client, GlideUrl url) {
--- a/library/pmd-ruleset.xml
+++ b/library/pmd-ruleset.xml
@@ -35,4 +35,24 @@
     </rule>
 
 
+    <!--Overrides default check to avoid violation when @Synthetic annotation is present-->
+    <rule ref="rulesets/java/design.xml/UncommentedEmptyConstructor"
+          message="Document empty constructor">
+
+        <properties>
+            <property name="xpath">
+                <value>
+                    <![CDATA[
+//ConstructorDeclaration[@Private='false'][count(BlockStatement) = 0 and
+($ignoreExplicitConstructorInvocation = 'true' or not(ExplicitConstructorInvocation)) and @containsComment = 'false'
+ and not(../Annotation/MarkerAnnotation/Name[@Image='Synthetic'])]
+ ]]>
+                </value>
+            </property>
+
+        </properties>
+
+    </rule>
+
+
 </ruleset>
--- a/library/src/main/java/com/bumptech/glide/ListPreloader.java
+++ b/library/src/main/java/com/bumptech/glide/ListPreloader.java
@@ -5,6 +5,7 @@ import android.widget.AbsListView;
 import com.bumptech.glide.request.target.BaseTarget;
 import com.bumptech.glide.request.target.SizeReadyCallback;
 import com.bumptech.glide.request.transition.Transition;
+import com.bumptech.glide.util.Synthetic;
 import com.bumptech.glide.util.Util;
 import java.util.List;
 import java.util.Queue;
@@ -210,8 +211,11 @@ public class ListPreloader<T> implements AbsListView.OnScrollListener {
   }
 
   private static class PreloadTarget extends BaseTarget<Object> {
-    private int photoHeight;
-    private int photoWidth;
+    @Synthetic int photoHeight;
+    @Synthetic int photoWidth;
+
+    @Synthetic
+    PreloadTarget() { }
 
     @Override
     public void onResourceReady(Object resource, Transition<? super Object> transition) {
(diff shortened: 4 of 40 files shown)


TWiStErRob:
Looking at the diff I think it would be a good idea to leave /*private*/ in the code and put a comment in the default constructors so that a future cleanup wouldn't reverse the changes.

How does adding a default scoped ctor help? Without a default scoped ctor (1 method) a private one is generated with an accessor (2 methods)?

RiteshChandnani:
@TWiStErRob 
Thanks for checking my PR.
I will put a comment in default constructors to avoid cleanup. But, I can't get my arms around - " I think it would be a good idea to leave /private/ in the code", can you please elaborate a bit more.

And, yes, if we don't have a default scoped ctor, two ctors will be synthesized, one will be the private constructor with no params and one will be a package private constructor with an arbitrary param (its type is dependent on compiler implementation, and hence one addtional empty class to support ctor overloading.
I also made a dummy implementation and used reflection and found the same.

Sorry, due to different timezones I was unable to get back to you immediately.

TWiStErRob:
> I think it would be a good idea to leave `/*private*/` in the code

If you look at this code out of context, what you see is that one of the fields are non-private for seemingly no reason (all usages of it are within the same class).

```java
  private final Glide glide;
  final Lifecycle lifecycle;
  private final RequestTracker requestTracker;
```

This would make it more clear that it was supposed to be private, but cannot due to some limitation.
```java
  private final Glide glide;
  /*private*/ final Lifecycle lifecycle;
  private final RequestTracker requestTracker;
```
potentially an even better approach would be a `@PreventSynthetic` or `@Private` source-level annotation to prevent thinking of commenting out as a forget-to-revert-after-debug thing.

Actually, thinking of it, isn't there a library out there which does this automatically? I mean if there is a way to have lambdas in Java 5, some APT code should be able to increase visibility of fields.

RiteshChandnani:
Got it.
So, we can have a `@PreventSynthetic` source-level annotation to make the code more clear.

Like this one:
https://github.com/hidroh/materialistic/blob/master/app/src/main/java/io/github/hidroh/materialistic/annotation/Synthetic.java

So, I can go forward and implement that source-level annotation and update the code accordingly?

Yeah, there should be a library which does this automatically, but I was unable to find something after doing a quick search.

TWiStErRob:
Yep, that example looks good, hidroh/materialistic#642 is where it was introduced.

So the changes are:
```diff
-   private final Type name;
+   @Synthetic final Type name;
```
I like the shorter `@Synthetic` name. Annotate the constructors as well, then they can be left empty.

TWiStErRob:
Btw, don't know what @sjudd's stand is on this, I hope he'll accept, but not sure.

RiteshChandnani:
Great.
Will do it right away!

Yeah, even I hope the same about that.

RiteshChandnani:
Btw, I think we should create a new directory named "annotation" in "glide/library/src/main/java/com/bumptech/glide/" 

What's your take?

RiteshChandnani:
Also, `diskLruCache` is a git submodule, can I go and update that one too?

TWiStErRob:
Yes, you can contribute to `disklrucache` as well, it's http://github.com/sjudd/disklrucache that you need to fork. You need two separate PRs (this and another) and you can't depend on code in each other (i.e. the annotation). As far as I remember the cache doesn't have many classes, so it should be less likely to have synthetics.

I would put the annotation in library/src/main/java/com/bumptech/glide/util, because it's not part of the public API.

RiteshChandnani:
@TWiStErRob 
Got it.
Thanks!
I will first finish with this PR and later look into `disklrucache`.

Will move the annotation directory under `library/src/main/java/com/bumptech/glide/util`

RiteshChandnani:
No worries. I will fix it.

TWiStErRob:
You can run `gradlew checkstyle pmd` to see the problems.

You'll likely need `@SuppressWarnings("PMD.SpecificViolationName")` on the constructors, or try to modify the PMD config if it's possible to suppress the violations when the `Synthetic` annotation is present.

RiteshChandnani:
All checks passed locally, but I used `@SuppressWarnings("PMD.UncommentedEmptyConstructor")` rather than changing PMD config because I don't know much about writing custom PMD rules.

What's your take?
Can I go forward and push that code?

TWiStErRob:
Why suppress if you can write a comment inside? Suppressions usually need to be explained, but if there's a comment, it explains itself.

I don't know much either, but based on a quick research it's quite simple. I found this: http://pmd.sourceforge.net/pmd-5.1.1/howtowritearule.html and the source for the rule is https://github.com/pmd/pmd/blob/master/pmd-java/src/main/resources/rulesets/java/design.xml#L1401-L1431 . I don't have time right now, but it looks like it should be as easy as adding `and not(hasAnnotation(....Synthetic))` (`junit.xml` has examples of how you can do the `hasAnnotation` part). Since `xpath` is just a property I think it could be overridden like `allowCommentedBlocks` in `<glide>/library/pmd-ruleset.xml`.

I say if you have time or curious go ahead with changing the rule, if not, just go with comments.

RiteshChandnani:
No worries, I will give it a try!
Thanks Robert!

TWiStErRob:
Nice. Just noticed something: the samples are not part of the API, they're just usage examples. I think it's better to keep synthetics there, so they're more portable (copy-paste).

RiteshChandnani:
Thanks a lot Robert!
So, I will remove synthetics from samples.

TWiStErRob:
Thanks :) Now, we just have to wait for @sjudd to ack.

Note: there are some weird new-lines left in diff
https://github.com/bumptech/glide/pull/1556/files#diff-4fdac1e84ca614a23bd46483d941e644

RiteshChandnani:
Yeah, no worries.

I will remove those weird lines.

TWiStErRob:
I think you misunderstood me there, I mean that there are some 1-line changes where the change is just inserting a new line without any other change:
image

The import groups should be separated by a new line, and Checkstyle/PMD may fail on this.

TWiStErRob:
No, but I mostly saw them in samples after your last push removing code from those. I suggest you squash your commits locally first, then amend the last commit to see if there are irrelevant modifications in the squashed diff. We would squash-merge anyway, so it doesn't matter that the commits are lost.

RiteshChandnani:
Btw, `I mean that there are some 1-line changes where the change is just inserting a new line without any other change`

I have to just double sure that you mean to say that I have to get rid of that new line, right?

TWiStErRob:
If there's no related change (like adding/removing an import or method), then there should no random whitespace changes in the diff either. So, yes, make sure any standalone added/removed newline is not in the diff, especially in files with no other changes.

RiteshChandnani:
Btw, looking at the original source files I found that grouping is not applied in imports. Hence, to maintain consistency across the project I think I should also follow the same path.

TWiStErRob:
Huh, that's interesting. Here's the commit that added groups: https://github.com/bumptech/glide/commit/f7a6d65cf7c1a41908dd48e0dab68ee5b881387e, but you're right the current code doesn't have groups, and that's because of https://github.com/bumptech/glide/commit/3f5543a1519e2b711a677f9072ad56734f492f86. This is a recent change, I haven't been following the `master` branch closely enough. So ,no separation. I wonder how Sam formatted the code... it looks like [the import order in settings](https://github.com/bumptech/glide/blob/master/.idea/codeStyleSettings.xml#L22-L256) wasn't updated to reflect the style.

RiteshChandnani:
Yeah! You are right too!

TWiStErRob:
The current diff looks good, sorry about the confusion.

RiteshChandnani:
No worries at all Robert!

I was honored to work on this!
Was not aware about PMD before this, now will go back and update my projects!

Thanks a lot! :)

sjudd:
Thanks for putting this together!

Can you do a quick before/after on the class/method count to verify the changes?

RiteshChandnani:
Yeah Sam here it is -

Before count: 54
After count: 3

TWiStErRob:
Huh, what's that 3 remaining?

RiteshChandnani:
But why Flag is not being raised by the IDE for these :/

screenshot from 2016-10-31 21-40-54

TWiStErRob:
IDEA/AS is not perfect, they are always working to make them better, just check out the [EAP release notes](https://confluence.jetbrains.com/display/IDEADEV/IDEA+2016.3+EAP). I reported [quite a few myself](https://youtrack.jetbrains.com/issues?q=created%20by:%20TWiStErRob%20) the latest one being what you just found: https://youtrack.jetbrains.com/issue/IDEA-163379

RiteshChandnani:
Ohh!
Thanks!
I was able to bring that down to 0 by making annotating those variables as `@Synthetic`

TWiStErRob:
Yep, the class file is what matters in the end, IDE warnings just helpers. Don't forget to commit and push the changes ;)

RiteshChandnani:
Yeah! Will do it right away! :+1:

RiteshChandnani:
After analyzing both versions of dex files I found total method count as -

Before: 3282
After: 3202

RiteshChandnani:
I found class count as -

Before: 470
After: 456

sjudd:
LGTM, Thanks!

I wonder if it's possible to write a build check for this?

RiteshChandnani:
@sjudd 

I am thankful to you too for writing such a great library! :)

`I wonder if it's possible to write a build check for this?`
Even I am musing about the same..

TWiStErRob:
I think PMD works on class files, so we can check if they have `access\$\d\d[02]` methods.

TWiStErRob:
@RiteshChandnani bug you found with sibling-synthetics is fixed already.

RiteshChandnani:
Yeah I checked too!
https://youtrack.jetbrains.com/issue/IDEA-16337

Nice that it has been fixed fixed! :)
Thanks Robert! :+1:

PR #1666, opened 2016-12-28
added note for DexGuard only

alashow:
https://github.com/bumptech/glide/wiki/Configuration#keeping-a-glidemodule

## Motivation and Context

Not everybody uses DexGuard and adding `-keepresourcexmlelements` will cause failing builds if one doesn't use DexGuard.


alashow:
I signed it!

sjudd:
Thanks! Sorry for the delay.

PR #1742, opened 2017-02-20
Add an annotation processor for GlideModules.

sjudd:
I've added an annotation processor that finds and "merges" all GlideModule implementations. This approach avoids issues with manifest parsing where it fails on some devices as well as the added parsing time at startup.

For now the annotation processor generates a simple class that just aggregates all GlideModules. In the long run we could make it more efficient by just dropping all GlideModules and generating a single class with all the methods merged.

I have not figured out a good way to make GlideModules optional with this system. With AndroidManifest based discovery, users could just use an exclude in their main manifest to exclude a GlideModule for a particular library. We may be able to do something similar using gradle, but it seems like it either requires two separate modules or a cycle in the dependency graph.


TWiStErRob:
For optionality: how about adding an annotation parameter so that you can exclude existing modules from your modules (similar to `@dagger.Module(includes={...})`, say:
```java
@com.bumptech.glide.annotation.GlideModule(overrides = {
    com.bumptech.glide.integration.okhttp3.OkHttpGlideModule.class
})
class MyOkHttpModule implements com.bumptech.glide.module.GlideModule {
    ...
}
```
Notice the array, because the users may want to override multiple modules with a single one. As far as I remember the `{}` are optional if there's only a single class, so the common use case would look nice as well.

sjudd:
That's an interesting idea, I like it! I'll try that after I fix the main issue here:

It turns out this doesn't work directly as written because the annotation processor only runs on annotated files that are being compiled from source, not those included in aars. So the way this works right now, if you write a GlideModule in your application, the processor will pick it up, but if you depend on an integration library with a GlideModule, the processor will ignore it. 

I can get around this by generating intermediate classes that are in a particular package and then adding an additional otherwise unused annotated class that's only included in applications, not libraries. The annotation processor will pick up the new annotated class and then in the processor, we can retrieve all the intermediate classes in our special package and generate the final module. 

It's a little awkward, annotation processors don't really like combining classes.

library/src/main/java/com/bumptech/glide/Glide.java
```diff
@@ -145,24 +147,97 @@ public static Glide get(Context context) {
...
         }
       }
     }
 
     return glide;
   }
 
+  private static Glide initGlide(Context context) {
+    Context applicationContext = context.getApplicationContext();
+    List<GlideModule> manifestModules = new ManifestParser(applicationContext).parse();
```
joshzana: Just a reminder to ensure there is a way to totally disable manifest parsing in keeping with #1754   Thanks!

sjudd: Done thanks!

sjudd:
Ok I think this is pretty much ready for review. This change does the following:

1. Adds ChildGlideModule and RootGlideModule abstract classes (Android convention to make it easier to change the interface later).
2. Adds @GlideModule annotation to find all Child and Root GlideModule implementations.
3. Adds @Excludes annotation to allow RootGlideModule implementations to exclude any ChildGlideModule or any legacy GlideModule.

Going forward RootGlideModule will be the place to specify all application specific options. Only RootGlideModules will have the ability to change cache sizes/implementations and perform other configurations that affect Applications the most. There can be only one RootGlideModule per application.

ChildGlideModules only allow registering additional components (ModelLoaders, decoders etc). They're meant for libraries. There can be many ChildGlideModules per library or application.

This change will work side by side with the existing manifest registered GlideModule interface. Applications that don't include a RootGlideModule will see a log warning them to do so. All of the existing libraries that Glide provides offer both an annotation and a manifest version of their modules. When Applications have fully migrated to the annotation processor, they can disable manifest parsing in their RootGlideModule. If Applications want to migrate slowly, they can switch one module at a time and use the @Excludes annotation to remove legacy GlideModule implementations as they go.

This change will require one additional gradle/maven dependency for each application and library for the annotation processor.

TWiStErRob:
Didn't look deeply yet, but based on your comments: forcing apps to have `RootGlideModule` sounds strange. What if I'm happy with the defaults, do I need to include a module then?

sjudd:
They're not going to be forced to have a RootGlideModule, but we will have a warning log because if you don't have a RootGlideModule and you want to include a library that has a ChildGlideModule, the ChildGlideModule will be ignored. 

We can try to word the warning so that it doesn't seem scary if you're happy with defaults.

sjudd:
Some documentation for the requirements is here: http://sjudd.github.io/glide/doc/configuration.html.

It's a bit more work for users who want to use the annotation processor than it was for GlideModules, but it's backwards compatible and not strictly required if users don't have modules. It can be substantially faster than parsing the manifest for metadata. It also allows configuring Glide even on devices where there are weird errors trying to read metadata from manifests (one way this happens is when the amount of metadata in an Application's manifest exceeds the platform's memory limit).

library/src/main/java/com/bumptech/glide/module/ChildGlideModule.java
```diff
@@ -0,0 +1,21 @@
...
+ * Registers a set of components to use when initializing Glide within an app when
+ * Glide's annotation processor is used.
+ *
+ * <p>Any number of ChildGlideModules can be contained within any library or application.
+ *
+ * <p>ChildGlideModules are called in no defined order. If ChildGlideModules within an application
+ * conflict, {@link RootGlideModule}s can use the {@link com.bumptech.glide.annotation.Excludes}
+ * annotation to selectively remove one or more of the conflicting modules.
+ */
+public abstract class ChildGlideModule implements RegistersComponents {
```
TWiStErRob: Why class and not interface? Aren't these going to end up the way RecyclerView inner classes should be interfaces and not classes?

sjudd: Class so that they're more extensible. Same logic that the framework is using. If you ever want to change or add a method, it's possible without breaking everyone.

TWiStErRob: Ah, I see, a workaround for non-available Java 8 default methods.

Issue #1233, opened 2016-05-25
Forward port 3.8.0 changes


TWiStErRob:
There are some changes in 3.8.0 which are not yet available in 4.0.
This ticket is here to remind that [3.8.0 milestone](https://github.com/bumptech/glide/issues?utf8=%E2%9C%93&q=milestone%3A3.8.0) need to revised for commits that should be forward ported into the new version, see also commits: [v3.7.0..v3.8.0](https://github.com/bumptech/glide/compare/v3.7.0...3.0).

If anyone wants to do it, it's up for grabs!

TWiStErRob:
Quick check of what needs to be done:

- [ ] 1281 Use application Context in GifResourceDecoder  
2cf9d2b forward port into ByteBufferGifDecoder
- [ ] 1119 Support contact Uri 394  
4f633ec forward port
- [ ] 1090 Fix thread-unsafe behaviours in GifDecoder  
dbb7be3 forward port into StandardGifDecoder
- [ ] 1093 Fill only the area that was modified by the previous GIF frame  
258cc10 forward port into StandardGifDecoder
- [ ] 1096 .asBitmap().crossFade()  
20b57a2 forward port, and close 840

More details:

| Issue ID |                                         Issue Title                                         |              Issue labels              | v3 commit |   v4 commit   |                 action                 |
|:--------:|:-------------------------------------------------------------------------------------------:|:--------------------------------------:|:---------:|:-------------:|:--------------------------------------:|
| 394      | Support contact uri                                                                         | feature                                | 1119      |               | ok, see PR                             |
| 637      | GIFs drop frames when many of them are displayed                                            | bug duplicate enhancement reproducible | 1060      |               | ok, see duplicate                      |
| 941      | OkHttp3 integration lib wrong pom xml                                                       | bug no-library                         | 9f53341   | v3 only       | ok, no port needed                     |
| 957      | Gradle compile failed on okhttp3 integration using dependency                               | bug duplicate                          | 941       |               | ok, see duplicate                      |
| 970      | Backport GIF overdraw fixes                                                                 | bug enhancement                        | 1092      |               | ok, see PR                             |
| 972      | Change invalid okhttp3-integration dep                                                      | bug                                    | 941       | v3 only       | ok, no port needed                     |
| 1003     | Could not find glide-parent:library:unspecified                                             | bug duplicate                          | 941       |               | ok, see duplicate                      |
[...]

PR #1753, opened 2017-02-23
Add Glide.setModulesEnabled API to allow apps to disable Manifest parsing

### LINKED ISSUE #684: Glide.get fails if PackageManager IPC fails
PackageManager has a failure mode where it throws a RuntimeException with "Package manager has died" when IPC calls fail.  We see this in the Dropbox app quite frequently.  In our internal calls to PackageManager we are careful to catch this RuntimeException and handle it in cases where we can.  

Unfortunately, adding Glide to our dependency list has started showing crashes with this stack:

``` ruby
Caused by: java.lang.RuntimeException: Package manager has died
    at android.app.ApplicationPackageManager.getApplicationInfo(ApplicationPackageManager.java:250)
    at com.bumptech.glide.module.ManifestParser.parse(panda.py:25)
    at com.bumptech.glide.Glide.get(panda.py:147)
    at com.dropbox.android.taskqueue.AnimatedGifDownloader.<init>(panda.py:139)
    at com.dropbox.android.DropboxApplication$AppState.initializeMainProcess(panda.py:899)
    at com.dropbox.android.DropboxApplication$AppState.<init>(panda.py:681)
    at com.dropbox.android.DropboxApplication.onCreate(panda.py:207)
```

For some reason not quite clear to me, we don't see this on anything above 4.4.4.

The request here would be to have a reasonable fallback when PackageManager is unavailable, or a way for apps to opt out of manifest scanning to pick up custom modules, since we don't use any.

joshzana:
## Description
Adding a static flag to Glide to disable parsing the manifest to find configured modules.
Added unit tests.
Fixes #684 

## Motivation and Context
The Dropbox app has been crashing on launch approximately 9000 times a day due to `RuntimeException`s  thrown by `PackageManager`.  We'd like to remove the dependency on `PackageManager` from our startup path, and we don't use any `GlideModule`s.  While #1742 in 4.0 will address this long term, we'd like a short term way to run Glide without runtime manifest parsing.

In order to enable unit testing, I decided to make this a set method that takes a `boolean`.

Not sure if I found the best way to unit test this.  Feel free to propose a cleaner approach.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/Glide.java
+++ b/library/src/main/java/com/bumptech/glide/Glide.java
@@ -75,6 +75,7 @@ import java.io.InputStream;
 import java.net.URL;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
 /**
@@ -87,6 +88,7 @@ public class Glide implements ComponentCallbacks2 {
   private static final String DEFAULT_DISK_CACHE_DIR = "image_manager_disk_cache";
   private static final String TAG = "Glide";
   private static volatile Glide glide;
+  private static boolean modulesEnabled = true;
 
   private final Engine engine;
   private final BitmapPool bitmapPool;
@@ -137,6 +139,17 @@ public class Glide implements ComponentCallbacks2 {
   }
 
   /**
+   * Enable or disable the parsing of AndroidManifest.xml
+   * looking for {@link GlideModule} implementations.
+   * Must be called before accessing the Glide singleton; otherwise, has no effect.
+   */
+  public static void setModulesEnabled(boolean enabled) {
+    synchronized (Glide.class) {
+      modulesEnabled = enabled;
+    }
+  }
+
+  /**
    * Get the singleton.
    *
    * @return the singleton
@@ -146,9 +159,8 @@ public class Glide implements ComponentCallbacks2 {
       synchronized (Glide.class) {
         if (glide == null) {
           Context applicationContext = context.getApplicationContext();
-          List<GlideModule> modules = new ManifestParser(applicationContext).parse();
-
           GlideBuilder builder = new GlideBuilder(applicationContext);
+          List<GlideModule> modules = parseGlideModules(applicationContext);
           for (GlideModule module : modules) {
             module.applyOptions(applicationContext, builder);
           }
@@ -161,6 +173,18 @@ public class Glide implements ComponentCallbacks2 {
     }
 
     return glide;
+  }
+
+  /**
+   * If modules are enabled, parses the application manifest and returns the configured modules.
+   * Otherwise, returns an empty list.
+   */
+  private static List<GlideModule> parseGlideModules(Context applicationContext) {
+    if (modulesEnabled) {
+      return new ManifestParser(applicationContext).parse();
+    } else {
+      return Collections.emptyList();
+    }
   }
 
   @VisibleForTesting
--- a/library/src/test/java/com/bumptech/glide/GlideTest.java
+++ b/library/src/test/java/com/bumptech/glide/GlideTest.java
@@ -2,7 +2,10 @@ package com.bumptech.glide;
 
 import static com.bumptech.glide.request.RequestOptions.decodeTypeOf;
 import static com.bumptech.glide.request.RequestOptions.placeholderOf;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
 import static org.mockito.Matchers.anyInt;
 import static org.mockito.Matchers.eq;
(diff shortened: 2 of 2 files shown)


library/src/test/java/com/bumptech/glide/GlideTest.java
```diff
@@ -547,6 +552,44 @@ public void testClone() throws IOException {
     verify(secondTarget).onResourceReady(notNull(Drawable.class), isA(Transition.class));
   }
 
+  @Test
+  public void testSetModulesEnabledTrue() throws Exception {
+    // teaDown glide instance first so we have a clean slate and not the Glide.get() call in setUp
```
TWiStErRob: "teaDown"

library/src/test/java/com/bumptech/glide/GlideTest.java
```diff
@@ -547,6 +552,44 @@ public void testClone() throws IOException {
...
+
+    // With modules enabled, SetupModule below should have been initialized,
+    // so the default modelLoader should NOT be configured
+    assertEquals(1, modelLoaders.size());
+    assertFalse(modelLoaders.get(0) instanceof HttpGlideUrlLoader);
+  }
+
+  @Test
+  public void testSetModulesEnabledFalse() throws Exception {
+    // teaDown glide instance first so we have a clean slate and not the Glide.get() call in setUp
```
TWiStErRob: "teaDown"

library/src/test/java/com/bumptech/glide/GlideTest.java
```diff
@@ -547,6 +552,44 @@ public void testClone() throws IOException {
...
+      Glide glide = Glide.get(getContext());
+      List<ModelLoader<GlideUrl, ?>> modelLoaders =
+          glide.getRegistry().getModelLoaders(mock(GlideUrl.class));
+
+      // With modules disabled, SetupModule below should not have been initialized,
+      // so the default modelLoader should remain
+      assertEquals(1, modelLoaders.size());
+      assertTrue(modelLoaders.get(0) instanceof HttpGlideUrlLoader);
+    } finally {
+      Glide.setModulesEnabled(true);
```
TWiStErRob: I think this should be part of Glide.tearDown, it resets Glide to pre-initialized state, which IMO includes both static variables. (It should probably even reset ViewTarget.tagId)

joshzana:
@TWiStErRob  - Why the v4 label?  I'm hoping this can go in before that.

joshzana:
Oops, yes.  I want this on the v3 branch.  Want a separate PR or should we land on master and then cherry-pick?

TWiStErRob:
We can't pick it directly, there are breaking changes in v4, though looking at the code this'll only impact a few lines in tests. Probably best to get two PRs in (see for example what I did in https://github.com/bumptech/glide/issues/1233... didn't know about cherry back then :)
Let's see what @sjudd thinks as well, he's the mastermind.

library/src/test/java/com/bumptech/glide/GlideTest.java
```diff
@@ -132,6 +137,7 @@ public Boolean answer(InvocationOnMock invocation) throws Throwable {
   @After
   public void tearDown() {
     Glide.tearDown();
+    Glide.setModulesEnabled(true);
```
TWiStErRob: I meant `Glide.tearDown` not the `@After`, though I guess it doesn't matter much.

library/src/main/java/com/bumptech/glide/Glide.java
```diff
@@ -137,6 +139,17 @@ public static File getPhotoCacheDir(Context context, String cacheName) {
   }
 
   /**
+   * Enable or disable the parsing of AndroidManifest.xml
+   * looking for {@link GlideModule} implementations.
+   * Must be called before accessing the Glide singleton; otherwise, has no effect.
+   */
+  public static void setModulesEnabled(boolean enabled) {
+    synchronized (Glide.class) {
```
sjudd: I'd probably add a check here to make sure that the Glide singleton hasn't already been instantiated.

joshzana:
OK I just made #1754 against the 3.0 branch.  Turns out a lot has changed around the test code, so that part is pretty different.

RE #1742. as I read it, it still parses the manifest every time, so it wouldn't fix #684 even if we were able to ship Glide 4.0 right now.  Maybe 4.0 needs its own implementation of this same method?

sjudd:
Yup that's true, but we'd probably enable a parameter similar to the one @TWiStErRob proposed for excluding modules to exclude parsing the manifest entirely.

It's mostly still there to not instantly break everything.

joshzana:
OK great.  How about I drop this PR, and you add a parameter to your PR?

TWiStErRob:
@joshzana so you want to close this PR in favor of the promise #1742 has?

joshzana:
Yep!  I'll add a comment over there just to make sure this doesn't get lost.

PR #1754, opened 2017-02-24
Add Glide.setModulesEnabled API for 3.x to allow apps to disable Manifest parsing

### LINKED ISSUE #684: Glide.get fails if PackageManager IPC fails

joshzana:
## Description
Adding a static flag to Glide to disable parsing the manifest to find configured modules.
Added unit tests.
Same as #1753 but backported to 3.0 branch
Fixes #684

## Motivation and Context
he Dropbox app has been crashing on launch approximately 9000 times a day due to RuntimeExceptions thrown by PackageManager. We'd like to remove the dependency on PackageManager from our startup path, and we don't use any GlideModules. While #1742 in 4.0 will address this long term, we'd like a short term way to run Glide without runtime manifest parsing.

In order to enable unit testing, I decided to make this a set method that takes a boolean.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/Glide.java
+++ b/library/src/main/java/com/bumptech/glide/Glide.java
@@ -72,6 +72,7 @@ import com.bumptech.glide.util.Util;
 import java.io.File;
 import java.io.InputStream;
 import java.net.URL;
+import java.util.Collections;
 import java.util.List;
 
 /**
@@ -83,6 +84,7 @@ public class Glide {
 
     private static final String TAG = "Glide";
     private static volatile Glide glide;
+    private static boolean modulesEnabled = true;
 
     private final GenericLoaderFactory loaderFactory;
     private final Engine engine;
@@ -137,6 +139,21 @@ public class Glide {
     }
 
     /**
+     * Enable or disable the parsing of AndroidManifest.xml
+     * looking for {@link GlideModule} implementations.
+     * @throws IllegalArgumentException if the Glide singleton has already been created.
+     */
+    public static void setModulesEnabled(boolean enabled) {
+        synchronized (Glide.class) {
+            if (glide != null) {
+                throw new IllegalArgumentException("Glide singleton already exists.");
+            }
+
+            modulesEnabled = enabled;
+        }
+    }
+
+    /**
      * Get the singleton.
      *
      * @return the singleton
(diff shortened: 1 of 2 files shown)


library/src/main/java/com/bumptech/glide/Glide.java
```diff
@@ -137,6 +139,21 @@ public static File getPhotoCacheDir(Context context, String cacheName) {
     }
 
     /**
+     * Enable or disable the parsing of AndroidManifest.xml
+     * looking for {@link GlideModule} implementations.
```
sjudd: Line break is a little early here.

sjudd:
Thanks!

joshzana:
Is there a snapshot build in maven anywhere that has this fix?  as far as I can tell, https://oss.sonatype.org/content/repositories/snapshots/com/github/bumptech/glide/glide/3.8.0-SNAPSHOT/ is pretty out of date.

If not, I can always build myself.

sjudd:
Not totally sure how this works, but it appears that there are snapshots built from 2/25 in that same folder: https://oss.sonatype.org/content/repositories/snapshots/com/github/bumptech/glide/glide/3.8.0-SNAPSHOT/glide-3.8.0-20170225.215140-15.jar

It looks like it also keeps around a few older versions?

PR #1798, opened 2017-03-20
Split GIF support out into an integration library

sjudd:
## Description
Moves GIF support into an integration library, leveraging the new annotation processor and GlideExtensions.


sjudd:
I need to uncomment and fix the tests here.

integration/gifdecoder/src/main/java/com/bumptech/glide/integration/gifdecoder/GifFrameLoader.java
```diff
@@ -1,7 +1,4 @@
-package com.bumptech.glide.load.resource.gif;
-
-import static com.bumptech.glide.request.RequestOptions.diskCacheStrategyOf;
-import static com.bumptech.glide.request.RequestOptions.signatureOf;
+package com.bumptech.glide.integration.gifdecoder;
```
TWiStErRob: RequestOptions used around here could be cached?

integration/gifdecoder/src/main/java/com/bumptech/glide/integration/gifdecoder/RequestManagerExtension.java
```diff
@@ -0,0 +1,24 @@
+package com.bumptech.glide.integration.gifdecoder;
+
+import android.graphics.drawable.Drawable;
+import com.bumptech.glide.TransitionOptions;
+import com.bumptech.glide.load.resource.drawable.DrawableTransitionOptions;
+import com.bumptech.glide.request.RequestOptions;
+
+/**
+ * Testing.
```
TWiStErRob: ?

integration/gifdecoder/src/test/java/com/bumptech/glide/integration/gifdecoder/ByteBufferGifDecoderTest.java
```diff
@@ -17,14 +14,15 @@
...
-import com.bumptech.glide.tests.GlideShadowLooper;
+import com.bumptech.glide.testlib.GlideShadowLooper;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.List;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
+import org.mockito.Matchers;
```
TWiStErRob: is this a new code style? It makes test code with arg matchers hard to read.

TWiStErRob: I think `Assert.*`, `Mockito.*`, `Matchers.*`, `ArgumentMatchers.*` etc should be always static imported.

integration/gifdecoder/src/test/java/com/bumptech/glide/integration/gifdecoder/GifDrawableTransformationTest.java
```diff
@@ -46,36 +40,49 @@ public void testSetsTransformationAsFrameTransformation() {
...
   public void testEquals() throws NoSuchAlgorithmException {
-    doAnswer(new Util.WriteDigest("first")).when(wrapped)
-        .updateDiskCacheKey(isA(MessageDigest.class));
-    KeyAssertions.assertSame(transformation, new GifDrawableTransformation(wrapped, bitmapPool));
-
-    @SuppressWarnings("unchecked") Transformation<Bitmap> other = mock(Transformation.class);
-    doAnswer(new Util.WriteDigest("other")).when(other)
-        .updateDiskCacheKey(isA(MessageDigest.class));
-    KeyAssertions.assertDifferent(transformation, new GifDrawableTransformation(other, bitmapPool));
+//    Mockito.doAnswer(new WriteDigest("first")).when(wrapped)
```
TWiStErRob: forgot this?

integration/gifencoder/build.gradle
```diff
@@ -9,6 +9,7 @@ dependencies {
     testCompile "org.mockito:mockito-all:${MOCKITO_VERSION}"
     testCompile "org.robolectric:robolectric:${ROBOLECTRIC_VERSION}"
     testCompile "com.android.support:support-v4:${SUPPORT_V4_VERSION}"
+    compile project(path: ':integration:gifdecoder')
```
TWiStErRob: `path: ` seems unnecessary, not used in other places

library/src/main/java/com/bumptech/glide/RequestManager.java
```diff
@@ -14,7 +14,7 @@
 import android.view.View;
 import com.bumptech.glide.load.engine.DiskCacheStrategy;
 import com.bumptech.glide.load.resource.drawable.DrawableTransitionOptions;
-import com.bumptech.glide.load.resource.gif.GifDrawable;
+//import com.bumptech.glide.load.resource.gif.GifDrawable;
```
TWiStErRob: forgot?

library/src/main/java/com/bumptech/glide/RequestManager.java
```diff
@@ -46,7 +46,7 @@
  */
 public class RequestManager implements LifecycleListener {
   private static final RequestOptions DECODE_TYPE_BITMAP = decodeTypeOf(Bitmap.class).lock();
-  private static final RequestOptions DECODE_TYPE_GIF = decodeTypeOf(GifDrawable.class).lock();
+//  private static final RequestOptions DECODE_TYPE_GIF = decodeTypeOf(GifDrawable.class).lock();
```
TWiStErRob: forgot?

library/src/main/java/com/bumptech/glide/load/resource/drawable/DrawableResource.java
```diff
@@ -39,8 +39,8 @@ public final T get() {
   public void initialize() {
     if (drawable instanceof BitmapDrawable) {
       ((BitmapDrawable) drawable).getBitmap().prepareToDraw();
-    } else if (drawable instanceof GifDrawable) {
-      ((GifDrawable) drawable).getFirstFrame().prepareToDraw();
+//    } else if (drawable instanceof GifDrawable) {
```
TWiStErRob: forgot?

if you need it how about introducing an interface (so there's only need for BitmapDrawable and that interface for instanceof)

library/src/main/java/com/bumptech/glide/request/BaseRequestOptions.java
```diff
@@ -24,10 +24,10 @@
 import com.bumptech.glide.load.resource.bitmap.Downsampler;
 import com.bumptech.glide.load.resource.bitmap.FitCenter;
 import com.bumptech.glide.load.resource.bitmap.VideoBitmapDecoder;
-import com.bumptech.glide.load.resource.gif.ByteBufferGifDecoder;
-import com.bumptech.glide.load.resource.gif.GifDrawable;
-import com.bumptech.glide.load.resource.gif.GifDrawableTransformation;
-import com.bumptech.glide.load.resource.gif.StreamGifDecoder;
+//import com.bumptech.glide.load.resource.gif.ByteBufferGifDecoder;
```
TWiStErRob: forgot?

library/src/test/java/com/bumptech/glide/GlideTest.java
```diff
@@ -402,29 +400,29 @@ public void testNonDefaultModelWithRegisteredFactoryDoesNotThrow() {
...
-    String fakeUri = "content://fake";
-    InputStream testGifData = openResource("test.gif");
-    mockUri(Uri.parse(fakeUri), testGifData);
-
-    requestManager.as(byte[].class).apply(decodeTypeOf(GifDrawable.class)).load(fakeUri)
-        .into(target);
-
-    verify(target).onResourceReady(isA(byte[].class), isA(Transition.class));
-  }
+//
```
TWiStErRob: forgot?

library/src/test/java/com/bumptech/glide/load/resource/bitmap/BitmapDrawableTransformationTest.java
```diff
@@ -1,5 +1,6 @@
 package com.bumptech.glide.load.resource.bitmap;
 
+import static com.bumptech.glide.testlib.Util.cast;
```
TWiStErRob: this is static imported, but Mockito not in other classes?

library/src/test/java/com/bumptech/glide/load/resource/drawable/DrawableResourceTest.java
```diff
@@ -50,15 +47,15 @@ public void testDoesNotReturnOriginalDrawableOnGet() {
...
-    GifDrawable expected = mock(GifDrawable.class);
-    Drawable.ConstantState constantState = mock(Drawable.ConstantState.class);
-    when(constantState.newDrawable()).thenReturn(expected);
-    when(drawable.getConstantState()).thenReturn(constantState);
-
-    assertEquals(expected, resource.get());
-
-    verify(drawable).getConstantState();
-    verify(constantState).newDrawable();
+//    GifDrawable expected = mock(GifDrawable.class);
```
TWiStErRob: forgot?

testlib/src/main/java/com/bumptech/glide/testlib/Util.java
```diff
@@ -0,0 +1,124 @@
...
+import org.mockito.Mockito;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
+import org.robolectric.util.ReflectionHelpers;
+
+// FIXME move to testutil module
+/**
+ * Generic test utilities.
+ */
+public class Util {
```
TWiStErRob: missing utility class no-constructor

testlib/src/main/java/com/bumptech/glide/testlib/Util.java
```diff
@@ -0,0 +1,124 @@
...
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.security.MessageDigest;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Mockito;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
+import org.robolectric.util.ReflectionHelpers;
+
+// FIXME move to testutil module
```
TWiStErRob: I think you just did this.

sjudd:
Yeah the tests here need to be cleaned up (see my earlier comment).

GlideApp is part of the generated API. The generated API is currently available only for Applications (it's only generated when we find a RootGlideModule).  I can't just use com.bumptech.glide.Glide for the generated API because it would break all non-application projects. Even if the generated API were available for libraries, you'd have a name collision if both the library and the application generated the same class in the same package (ie com.bumptech.glide.Glide). 

The current thought on the generated API is:
1. Applications use the generated API by including a RootGlideModule and the name is different for the generated API to avoid confusion (defaults to GlideApp, but applications can change the name. For example the Google Photos version is called PhotosGlide).
2. Libraries (and Applications that don't want to deal with the complexity) use the existing Glide class and the fixed API.

Other solutions include:
1. Generating a Glide class for every library and/or application but base the Glide class on the package name (so you'd have com.google.android.apps.photos.Glide, com.bumptech.glide.integration.okhttp3.Glide etc). 
2. Generate com.bumptech.glide.Glide for applications and force libraries to use a renamed GenericGlide/BaseGlide etc.
3. Generate com.bumptech.glide.Glide for applications and generate some other named class for libraries.

The two basic limitations are:
1. We need an annotated class to generate an API and especially to generate an API in a custom package (ie in the library or application's package). Right now, the best I can think of is to generate APIs for applications/libraries with Root or Child GlideModules.
2. We need a coherent way for libraries and applications to migrate. We could take a hard line approach, force everyone to use at least one GlideModule (Root or Child) and only expose Glide and similar APIs via generated code. However, doing so would break a bunch of snapshot users (and Google apps), so we'd probably need to fork and abandon the current master as a branch to accomplish this.

Thoughts definitely welcome. The main reason I put this up so early was to hopefully get comments on the generated API and how it can work with libraries, especially those that extend the API.

TWiStErRob:
Wow, nice writeup. I'll take a look on generated code, it's kind of unrelated to this issue. I think I still have your apt PR in my queue :)

sjudd:
I think all the comments here are addressed.

PR #1800, opened 2017-03-20
Add a Glide.with method for View subclasses.

sjudd:
The implementation, especially for non-support 
Fragments, relies on implementation details of
FragmentManagerImpl. It’s not ideal, but it does
resolve a longstanding problem where View 
subclasses end up using the wrong RequestManager.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/Glide.java
+++ b/library/src/main/java/com/bumptech/glide/Glide.java
@@ -16,6 +16,7 @@ import android.support.annotation.VisibleForTesting;
 import android.support.v4.app.Fragment;
 import android.support.v4.app.FragmentActivity;
 import android.util.Log;
+import android.view.View;
 import com.bumptech.glide.gifdecoder.GifDecoder;
 import com.bumptech.glide.load.DecodeFormat;
 import com.bumptech.glide.load.data.InputStreamRewinder;
@@ -69,6 +70,7 @@ import com.bumptech.glide.module.ManifestParser;
 import com.bumptech.glide.request.RequestOptions;
 import com.bumptech.glide.request.target.ImageViewTargetFactory;
 import com.bumptech.glide.request.target.Target;
+import com.bumptech.glide.util.Preconditions;
 import com.bumptech.glide.util.Util;
 import java.io.File;
 import java.io.InputStream;
@@ -499,12 +501,10 @@ public class Glide implements ComponentCallbacks2 {
   private static RequestManagerRetriever getRetriever(@Nullable Context context) {
     // Context could be null for other reasons (ie the user passes in null), but in practice it will
     // only occur due to errors with the Fragment lifecycle.
-    if (context == null) {
-      throw new IllegalArgumentException(
-          "You cannot start a load on a Fragment where getActivity() returns null (which usually"
-              + " occurs when getActivity() is called before the Fragment is attached or after the"
-              + " Fragment is destroyed).");
-    }
+    Preconditions.checkNotNull(
+        "You cannot start a load on a not yet attached View or a  Fragment where getActivity() "
+            + "returns null (which usually occurs when getActivity() is called before the Fragment "
+            + "is attached or after the Fragment is destroyed).");
     return Glide.get(context).getRequestManagerRetriever();
   }
 
@@ -578,6 +578,30 @@ public class Glide implements ComponentCallbacks2 {
    */
   public static RequestManager with(Fragment fragment) {
     return getRetriever(fragment.getActivity()).get(fragment);
+  }
+
+  /**
+   * Begin a load with Glide that will be tied to the lifecycle of the {@link Fragment},
+   * {@link android.app.Fragment}, or {@link Activity} that contains the View.
+   *
+   * <p>A {@link Fragment} or {@link android.app.Fragment} is assumed to contain a View if the View
+   * is a child of the View returned by the {@link Fragment#getView()}} method.
+   *
+   * <p>This method will not work if the View is not attached. Prefer the Activity and Fragment
+   * variants unless you're loading in a View subclass.
+   *
+   * <p>This method may be inefficient for large hierarchies. Consider memoizing the result after
+   * the View is attached.
+   *
+   * <p>When used in Applications that use the non-support {@link android.app.Fragment} classes,
+   * calling this method will produce noisy logs from {@link android.app.FragmentManager}. Consider
+   * avoiding entirely or using the {@link Fragment}s from the support library instead.
+   *
+   * @param view The view to search for a containing Fragment or Activity from.
+   * @return A RequestManager that can be used to start a load.
+   */
+  public static RequestManager with(View view) {
+    return getRetriever(view.getContext()).get(view);
   }
 
   public Registry getRegistry() {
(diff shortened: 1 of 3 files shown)


library/src/main/java/com/bumptech/glide/manager/RequestManagerRetriever.java
```diff
@@ -60,6 +71,11 @@
    */
   private final Handler handler;
 
+  // Objects used to find Fragments and Activities containing views.
+  private final ArrayMap<View, Fragment> tempViewToSupportFragment = new ArrayMap<>();
+  private final ArrayMap<View, android.app.Fragment> tempViewToFragment = new ArrayMap<>();
```
TWiStErRob: Will this work on version not having `android.app.Fragment`? My guess is yes, due to erasure.

library/src/main/java/com/bumptech/glide/Glide.java
```diff
@@ -559,6 +560,31 @@ public static RequestManager with(Fragment fragment) {
...
+   * the View is attached.
+   *
+   * <p>When used in Applications that use the non-support {@link android.app.Fragment} classes,
+   * calling this method will produce noisy logs from {@link android.app.FragmentManager}. Consider
+   * avoiding entirely or using the {@link Fragment}s from the support library instead.
+   *
+   * @param view The view to search for a containing Fragment or Activity from.
+   * @return A RequestManager that can be used to start a load.
+   */
+  public static RequestManager with(View view) {
```
TWiStErRob: That's a lot of warnings, well documented!

library/src/main/java/com/bumptech/glide/manager/RequestManagerRetriever.java
```diff
@@ -133,6 +149,136 @@ public RequestManager get(Fragment fragment) {
     }
   }
 
+  public RequestManager get(View view) {
+    if (view.getContext() == null) {
+      throw new IllegalArgumentException("Unable to obtain a request manager for a view without"
```
TWiStErRob:
```java
throw new IllegalArgumentException(
    "Unable to obtain a request manager for a view without a Context");
```

would help readabilty, and full-text-search for the message; and it should still fit within margin.

library/src/main/java/com/bumptech/glide/manager/RequestManagerRetriever.java
```diff
@@ -133,6 +149,136 @@ public RequestManager get(Fragment fragment) {
...
+    if (activity instanceof FragmentActivity) {
+      Fragment fragment = findSupportFragment(view, (FragmentActivity) activity);
+      if (fragment == null) {
+        return get(activity);
+      }
+      return get(fragment);
+    }
+
+    // Standard Fragments.
+    android.app.Fragment fragment = findFragment(view, activity);
```
TWiStErRob: There's no help of erasure here. This should crash with a ClassNotFoundException on older devices if the activity is not a `FragmentActivity`.

library/src/main/java/com/bumptech/glide/manager/RequestManagerRetriever.java
```diff
@@ -133,6 +149,136 @@ public RequestManager get(Fragment fragment) {
...
+
+  @Nullable
+  private Fragment findSupportFragment(View target, FragmentActivity activity) {
+    tempViewToSupportFragment.clear();
+    findAllSupportFragmentsWithViews(
+        activity.getSupportFragmentManager().getFragments(), tempViewToSupportFragment);
+    Fragment result = null;
+    View activityRoot = activity.findViewById(android.R.id.content);
+    View current = target;
+    while (!current.equals(activityRoot)) {
```
TWiStErRob: Think about: What happens if users passes in DecorView?

library/src/main/java/com/bumptech/glide/manager/RequestManagerRetriever.java
```diff
@@ -133,6 +149,136 @@ public RequestManager get(Fragment fragment) {
...
+    }
+    tempViewToFragment.clear();
+    return result;
+  }
+
+  @RequiresApi(api = VERSION_CODES.HONEYCOMB)
+  private void findAllFragmentsWithViews(
+      android.app.FragmentManager fragmentManager,
+      ArrayMap<View, android.app.Fragment> result) {
+    String key = "i";
```
TWiStErRob: Did you want to use `FRAGMENT_INDEX_KEY` instead of `key` local?

library/src/main/java/com/bumptech/glide/manager/RequestManagerRetriever.java
```diff
@@ -36,6 +44,9 @@
   private static final int ID_REMOVE_FRAGMENT_MANAGER = 1;
   private static final int ID_REMOVE_SUPPORT_FRAGMENT_MANAGER = 2;
 
+  // This can be literally any value given the way FragmentManagerImpl is implemented.
+  private static final String FRAGMENT_INDEX_KEY = "key";
```
TWiStErRob: This could use a bit more explanation, I think.

sjudd:
Thanks for looking. 

Is this too much of a hack to submit? Even though the support library has a method to obtain a list of all fragments in a given fragment manager, the method's documentation indicates that it's not meant for external use. The non-support fragment version of this will spam error logs, even if it works.

I go back and forth on this.

TWiStErRob:
I don't oppose it. Do you actually have use case for this, or it was just an idea that popped into your head? I don't see where it indicates the internal use.

I wonder if adding an accessor class for `FragmentManagerImpl.mActive` to `android.app` package would be less of a hack.

sjudd:
The use case is if you have a custom view inside of a Fragment. In the current API if you try to start a request inside of the custom view, all you can do is pass in the View's Context, which will give you the Activity's RequestManager, not the Fragment's.

library/src/main/java/com/bumptech/glide/manager/RequestManagerRetriever.java
sjudd: Added a bit more, it's weird either way.

library/src/main/java/com/bumptech/glide/manager/RequestManagerRetriever.java
sjudd: I'm not sure, but I bumped the min sdk version to 14, so this should be a non-issue.

library/src/main/java/com/bumptech/glide/manager/RequestManagerRetriever.java
sjudd: Done.

library/src/main/java/com/bumptech/glide/manager/RequestManagerRetriever.java
sjudd: Good catch, fixed by the min sdk version change.

library/src/main/java/com/bumptech/glide/manager/RequestManagerRetriever.java
sjudd: Seems like it's ok? At some point the View's parent will be null or a non-view, in which case the loop breaks below.

library/src/main/java/com/bumptech/glide/manager/RequestManagerRetriever.java
sjudd: Done thanks.

sjudd:
I've taken the time to test this method, so I'll use it for now. I'd be open to the accessor class as well if anyone is interested in trying that out as an alternative to the hacks here.

PR #1810, opened 2017-03-22
Set min sdk version for Glide to 14.

sjudd:
Practically speaking this is already required by interfaces implemented in several classes. Given that ~1% of devices are still using Gingerbread, less than 1% are using honeycomb and it’s hard for us to avoid using newer APIs even in our sample apps, it’s probably time to move the sdk version forward.


library/src/main/java/com/bumptech/glide/load/engine/cache/MemorySizeCalculator.java
```diff
@@ -109,13 +108,9 @@ private String toMb(int bytes) {
...
   }
 
-  @TargetApi(Build.VERSION_CODES.KITKAT)
   private static boolean isLowMemoryDevice(ActivityManager activityManager) {
-    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
-      return activityManager.isLowRamDevice();
-    } else {
-      return Build.VERSION.SDK_INT < Build.VERSION_CODES.HONEYCOMB;
-    }
+    return Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT
```
TWiStErRob: Didn't we have some problems with shortcutting vs Build.VERSION before?

sjudd: Oh yeah... I have some vague recollection. I think you're right, I'll change this.

sjudd:
The interface is ComponentCallbacks2. Yes that only applies to Glide v4. v3 should still work back to gingerbread.

PR #1813, opened 2017-03-24
Fix wrong interpretation of the Netscape iteration count in an animated GIF

### LINKED ISSUE #1724: glide load gif never loop with  LOOP_INTRINSIC mode
here i find a bug, when glide load git with LOOP_INTRINSIC mode, a forever loop gif will never loop at all

=====code====
in third party parse  
https://gist.githubusercontent.com/devunwired/4479231/raw/df2725be4ae0f12f5265deaf0a769936ea94950b/GifDecoder.java
**protected int loopCount = 1; // iterations; 0 = repeat forever**

    /**
     * Reads Netscape extension to obtain iteration count.
     */
    private void readNetscapeExt() {
        do {
            readBlock();
            if (block[0] == 1) {
                // Loop count sub-block.
                int b1 = ((int) block[1]) & 0xff;
                int b2 = ((int) block[2]) & 0xff;
                header.loopCount = (b2 << 8) | b1;
            }
        } while ((blockSize > 0) && !err());
    }

**while in glide loopCount = 0 means never loop** 

    @Override
    public void setLoopCount(int loopCount) {
        if (loopCount <= 0 && loopCount != LOOP_FOREVER && loopCount != LOOP_INTRINSIC) {
            throw new IllegalArgumentException("Loop count must be greater than 0, or equal to "
                    + "GlideDrawable.LOOP_FOREVER, or equal to GlideDrawable.LOOP_INTRINSIC");
        }

        if (loopCount == LOOP_INTRINSIC) {
            maxLoopCount = decoder.getLoopCount();
        } else {
            maxLoopCount = loopCount;
        }
    }
hope u will fix it soon or maybe i could create a pr

toyama-sumio:
This is a pull request into not `master` but `3.0` branch.

If Glide shows an animated GIF with LOOP_INTRINSIC option as follows, the GIF animation loops incorrect number of times.

```java
ImageView target = (ImageView) findViewById(R.id.imageView);
Glide.with(this)
        .load(Uri.parse("<animated GIF URL>"))
        .into(new GlideDrawableImageViewTarget(target, GlideDrawable.LOOP_INTRINSIC));
```

This PR makes a GIF animation loop correct number of times as follows, and also fixes #1724:

| Netscape iteration count in animated GIF | Before merged | After merged |
|------------------------------------------|--------|-------|
| Doesn't exist                            | Only the first frame is shown.<br>Not animated.<br>(due to #1724) | Loops once |
| 0 (forever)                              | Only the first frame is shown.<br>Not animated.<br>(due to #1724) | Loops forever |
| 1                                        | Loops once | Loops twice |
| 2                                        | Loops twice | Loops 3 times |
| `n`                                      | Loops `n` times | Loops `n + 1` times |

Note that the fixed behavior of looping is the same as one of most browsers (IE, Edge, Safari and Chrome),
although an interpretation of the Netscape iteration count in animaged GIFs is ambiguous.  
(See the discussion of [Chromium Issue 592735 comment 5](https://bugs.chromium.org/p/chromium/issues/detail?id=592735#c5) for more information.)


sjudd:
Neat! Thanks for doing this.

If you'd like to try to cherry pick it into the master branch as well, I'd be grateful.

toyama-sumio:
@sjudd OK. I'll try it!
I'll create new Pull Request when it's finished.

library/src/main/java/com/bumptech/glide/load/resource/gif/GifDrawable.java
```diff
@@ -309,7 +309,23 @@ public void setLoopCount(int loopCount) {
         }
 
         if (loopCount == LOOP_INTRINSIC) {
-            maxLoopCount = decoder.getLoopCount();
+            int netscapeIterationCount = decoder.getNetscapeIterationCount();
```
TWiStErRob: This method should have been part of the decoder, people sometimes use the decoder directly, so any time they need to get the real loop count, they would have to copy this code from GifDrawable into wherever they need it.

TWiStErRob: This is still in SNAPSHOT so it's possible to move it to `getNetscapeIterationCount` without breaking, what do you think? @sjudd @toyama-sumio

toyama-sumio: @TWiStErRob @sjudd
Although I was not sure the difference between the responsibilities that `GifDecoder` and `GifDrawable` have, I have sorted out my thoughts as follows:

- `GifDecoder` reflects the _raw_ data structure of the GIF file.
- `GifDrawable` interprets the data structure exposed by `GifDecoder` so that Glide can handle it easily.

If the above thoughts are different from yours, I can move the code snippet to `GifDecoder`.
In that case, I'm thinking to implement it with the following manner. Any comments are welcome!

- Leave `getNetscapeIterationCount()` which gets _raw_ netscape iteration count
  so that other interpretations of the Netscape iteration count (like Firefox) can be used.
- Create a following method in `GifDecoder`.  
  (Is `Resulting` an appropriate naming?)
    
    ```java
    public int getResultingLoopCount() {
        if (header.loopCount == GifHeader.NETSCAPE_ITERATION_COUNT_DOES_NOT_EXIST) {
            return 1;
        }
        return header.loopCount + 1;
    }
    ```
- In this place, the code snippet becomes as follows.
  Note that a conditional branch can't be eliminated completely because the value of `FOREVER` is different between `GifDecoder` and `GifDrawable`. 
    
    ```java
    ...
    int resultingLoopCount = decoder.getResultingLoopCount();
    if (resultingLoopCount == GifHeader.NETSCAPE_ITERATION_COUNT_FOREVER) {
        maxLoopCount = LOOP_FOREVER;
    } else {
        maxLoopCount = resultingLoopCount;
    }
    ...
    ```

TWiStErRob: As far as I see `GifHeader` is the raw, `GifDecoder` is the interpretation and `GifDrawable` is the display of the interpretation. As I said, you can use `GifDecoder` directly to decode frames without using the Glide pipeline.

I would name them `getNetscapeLoopCount()` (to get raw header data) and `getTotalIterationCount()` (to get useful data) so to distinguish between "loop" and "iteration": "iteration" being one repeat of frames, "loop" being a repeated iteration after the first iteration. This should reflect the "loop+1" interpretation in the Chrome issue. You could also emphasize this with some JavaDoc on these methods. I think the mapping between 0 (`NETSCAPE_FOREVER`) and GlideDrawable's one is OK to stay there.

toyama-sumio: @TWiStErRob 
Thanks for your clarification! I understand.
Your suggested naming also helps me. I follow it.

I'll create a new pull request for this fix.

PR #1860, opened 2017-04-17
Replace the unnecessary continue syntax.

christmasjason:
Replace the unnecessary continue syntax.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/Glide.java
+++ b/library/src/main/java/com/bumptech/glide/Glide.java
@@ -185,13 +185,12 @@ public class Glide implements ComponentCallbacks2 {
           annotationGeneratedModule.getExcludedModuleClasses();
       for (Iterator<GlideModule> iterator = manifestModules.iterator(); iterator.hasNext();) {
         GlideModule current = iterator.next();
-        if (!excludedModuleClasses.contains(current.getClass())) {
-          continue;
+        if (excludedModuleClasses.contains(current.getClass())) {
+          if (Log.isLoggable(TAG, Log.DEBUG)) {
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/bumptech/glide/Glide.java
```diff
@@ -185,13 +185,12 @@ private static void initGlide(Context context) {
           annotationGeneratedModule.getExcludedModuleClasses();
       for (Iterator<GlideModule> iterator = manifestModules.iterator(); iterator.hasNext();) {
         GlideModule current = iterator.next();
-        if (!excludedModuleClasses.contains(current.getClass())) {
-          continue;
+        if (excludedModuleClasses.contains(current.getClass())) {
+          if (Log.isLoggable(TAG, Log.DEBUG)) {
+            Log.d(TAG, "RootGlideModule excludes manifest GlideModule: " + current.getClass());
```
TWiStErRob: Any particular reason to reduce the amount of info logged? Changing to `current.getClass()` loses information in the logs. Consider for example if for some reason there are two modules exists of the same type, seeing the `hashCode` may be most helpful to realize if it's the same object or there are two instances for some reason.

TWiStErRob:
Wow, actually it's more than my personal preference, it's a PMD error, the build failure is because of this: https://pmd.github.io/pmd-5.4.0/pmd-java/rules/java/design.html#AvoidDeeplyNestedIfStmts

I suggest we close and abandon this change.

christmasjason:
Wow, i see. So i think it is must be the continue sytax exist reason. Thanks.

PR #1946, opened 2017-05-19
Add vector drawable support via appcompat

Teovald:
## Description

use the support lib in order to inflate resources.  
This allows you to use the ids of VectorDrawables for placeholder, error or fallback calls.

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -20,7 +20,7 @@ dependencies {
     compile project(':third_party:disklrucache')
     compile project(':annotation')
     compile "com.android.support:support-v4:${SUPPORT_V4_VERSION}"
-
+    provided "com.android.support:appcompat-v7:${SUPPORT_V7_VERSION}"
     testCompile project(':testutil')
     testCompile 'com.google.guava:guava-testlib:18.0'
     testCompile "com.google.truth:truth:${TRUTH_VERSION}"
--- a/library/src/main/java/com/bumptech/glide/request/SingleRequest.java
+++ b/library/src/main/java/com/bumptech/glide/request/SingleRequest.java
@@ -2,8 +2,10 @@ package com.bumptech.glide.request;
 
 import android.content.res.Resources;
 import android.graphics.drawable.Drawable;
+import android.support.annotation.DrawableRes;
 import android.support.v4.content.res.ResourcesCompat;
 import android.support.v4.util.Pools;
+import android.support.v7.content.res.AppCompatResources;
 import android.util.Log;
 import com.bumptech.glide.GlideContext;
 import com.bumptech.glide.Priority;
@@ -102,6 +104,7 @@ public final class SingleRequest<R> implements Request,
   private Drawable fallbackDrawable;
   private int width;
   private int height;
+  private static boolean shouldCallAppCompatResources = true;
 
   public static <R> SingleRequest<R> obtain(
       GlideContext glideContext,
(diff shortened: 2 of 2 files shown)


TWiStErRob:
I don't think that's SVG, there's no SVG in Android as far as I know, only vector drawables:
https://developer.android.com/reference/android/graphics/drawable/VectorDrawable.html

The [path descriptor mini-language](https://www.w3.org/TR/SVG/paths.html#PathData) is the only thing that's borrowed from SVG.

Anyway, would it be possible to make this optional? Without adding a hard dependency on v7?

Also, @sjudd, we should probably revise the v4 dep and use the more fine-grained exploded libraries (core, fragment, etc).

library/build.gradle
```diff
@@ -20,7 +20,7 @@ dependencies {
     compile project(':third_party:disklrucache')
     compile project(':annotation')
     compile "com.android.support:support-v4:${SUPPORT_V4_VERSION}"
-
+    compile "com.android.support:appcompat-v7:${SUPPORT_V7_VERSION}"
```
TWiStErRob: Would it be possible to make this optional? Without adding a hard dependency on v7?

Also, @sjudd, we should probably revise the v4 dep and use the more fine-grained exploded libraries (core, fragment, etc).

samples/svg/src/main/res/drawable/image_loading.xml
```diff
@@ -1,12 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <layer-list xmlns:android="http://schemas.android.com/apk/res/android">
-    <item>
-        <bitmap
-            android:src="@android:drawable/sym_def_app_icon"
-            android:gravity="center" />
```
TWiStErRob: According to the official docs this resource is available in O. It would be a breaking change for them to remove it without deprecation first. Maybe the preview is missing it (in which case report an AOSP bug), I would wait until a final release before taking action.

Teovald: ok, removed

TWiStErRob: I'm confused, it looks like you squashed it into the commit, not reverted.

Teovald: my bad, pushed too soon, fixed.

Teovald:
and yes, svg is an abuse of language. 
Vector drawable is more like a subset of the svg standard mixed in with android constructs

library/build.gradle
sjudd: Yeah I'm not 100% sure it makes sense to add a hard dependency on v7 either. Maybe this should be an integration library? You can probably override the setPlaceholder methods with a GlideExtension, though it's a bit weird.

As an alternative Apps can also pass in Drawables directly, so those that are using v7 can call this method themselves.

Teovald: I would prefer to be able to pass an drawable id.  
- Terser calls (no need to initialise the drawable yourself, just pass the id)
- No risk of a crash because you passed the id of a VD to glide (right now it does crash)
- If you refactor a raster to VD, no need to check all the glide calls


Alternatively,

I have not toyed too much with reflection in this kind of use case but I think it might be possible to check for the presence of the class & method from the support lib and use them if they are here.  
If it is not there, fallback to the previous method.

I would also need to add a proguard rule for this class in the library's rules. Although I don't know yet whether I can add a rule for a class that might not be present in the classpath.

Teovald: Would that be an acceptable implementation ?

TWiStErRob: I think a better way would be to use a provided scope and `try { loadVectorWithV7 } catch(ClassNotFoundException ex) { loadVectorWithV7 }` (could cache failure so the weight of the exception only happens once). This way there's no need for proguard.

TWiStErRob: (I support the PR, the more formats we support the better)

Teovald: Just read about provided dependencies
Indeed it looks like a good way to do this.

Teovald:
It should be ok now :-) 
I have tested this commit on both pre and post L devices.

sjudd:
From the failing travis build:
/home/travis/build/bumptech/glide/library/src/main/java/com/bumptech/glide/request/SingleRequest.java:359: error: Line is longer than 100 characters (found 101).

Otherwise looks good to me. Worth noting for the peanut gallery that this still won't let you call ``load()`` with a resource id that's a vector drawable. It will let you use vector drawables for placeholder, error, and fallback.

Teovald:
My bad, I should not have added a javadoc at the last minute without running check afterwards.
Should be ok now

sjudd:
Thanks!

PR #1989, opened 2017-05-31
Fix broken Proguard rules

saschpe:
AppGlideModule resides in the 'module' sub-package and the GeneratedAppGlideModuleImpl needs to be kept as well so that Class.forName("com.bumptech.glide.GeneratedAppGlideModuleImpl") inside Glide.getAnnotationGeneratedGlideModules() won't fail.

People's obfuscated release builds will break otherwise.

### DIFF
--- a/_posts/2017-03-14-configuration.md
+++ b/_posts/2017-03-14-configuration.md
@@ -61,7 +61,8 @@ annotationProcessor 'com.github.bumptech.glide:compiler:1.0.0-SNAPSHOT'
 
 Finally, you should keep AppGlideModule implementations in your ``proguard.cfg``:
 ```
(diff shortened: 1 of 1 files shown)


TWiStErRob:
Are you sure it breaks?
> **Does ProGuard handle Class.forName calls?**
> Yes. ProGuard automatically handles constructs like Class.forName("SomeClass") and SomeClass.class. The referenced classes are preserved in the shrinking phase, and the string arguments are properly replaced in the obfuscation phase. -- https://www.guardsquare.com/en/proguard/faq

saschpe:
Yes. This was the important part. I'm using the beta version of the upcoming Android Gradle plugin. Not sure which Proguard version it uses internally.

sjudd:
Thanks for fixing this!

Issue #735, opened 2015-11-12
Image encryption in DiskCache


anotherdev:
I want to encrypt image files that are stored in Glide DiskCache using Facebook Conceal library. What would be the best approach?

TWiStErRob:
I think you'll need to create an `encoder` and `cacheDecoder` to wrap the built-in ones.

It is kind of like having a custom image format, see #718.

anotherdev:
Thanks for quick answer! I'll try to implement it.

Quick follow-up question: I'm migrating from Fresco (because of image encryption feature). Fresco clears the memory cache when app is backgrounded, does Glide support this also? If not, is there a way to do it?

TWiStErRob:
There's a `Glide.clearMemory()` method but, it's quite pointless to do so IMO. The JVM running your application won't be terminated when in background, so the heap consumption for the app (OS-level) should stay the same regardless of having no data in it or full of potentially useful bitmaps. If the OS needs space it'll call `onLowMemory` where Glide should automatically call that method.

It sounds like when resuming a Fresco-based app where `onResume` reloads all ImageViews (e.g. by reloading the data from DB) takes a long time to actually see images. NB I didn't try Fresco yet.

anotherdev:
Yeah, I agree. Clearing memory thing is more on security side (a client request). We have to do it this way or else we must disable memory cache.

anotherdev:
The encoder part seem to be straightforward.

I'm doing the decoder part. The starting point is `GifBitmapWrapperStreamResourceDecoder` and `GifBitmapWrapperResourceDecoder`. Which class & config should I use for the following?

1. `ResourceDecoder<ImageVideoWrapper, Bitmap>`
2. `ResourceDecoder<InputStream, GifDrawable>`

TWiStErRob:
Sorry, but I don't understand your question fully, the answer may be (by looking at implementations of the `ResourceDecoder` interface):
1. `ImageVideoBitmapDecoder`
2. `GifResourceDecoder`

TWiStErRob:
To make things easier you might want to try `.asBitmap()` which disables animated GIF support. But if that's needed, then your best bet is to just try to match the correct generic arguments.

sjudd:
You might actually be able to do this more easily in a DiskCache implementation. That way you can just encrypt all data as it goes in and comes out.

TWiStErRob:
[...]

PR #1994, opened 2017-06-01
Failed code of using Facebook Conceal to encrypt and decrypt the image across the whole app in Glide 4.0

TonyTangAndroid:
I am trying to integrate Facebook Conceal with Glide 4.0 to encrypt the cache of image globally across the whole app and decrypt the file before decoding the file into bitmap again.

I did not find nice document in how to achieve the purpose. But in according to the discussion [here](https://github.com/bumptech/glide/issues/735), it is possible by defining a new ResourceDecoder and ResourceEncoder, which is exactly what I had tried to do in the code base based on the sample app of Flickr. But it did not success.  

Any help or suggestion will be highly appreciated.


TonyTangAndroid:
Wrong pull request sent. I will close it.

PR #2020, opened 2017-06-09
Proguard - Keep AppGlideModule

minas90:


TWiStErRob:
README.md is deprecated I think, the full docs for v4 is at http://bumptech.github.io/glide/doc/download-setup.html#proguard I'm working on cleaning this up.

minas90:
ok great 👍  on the main page of the project all the info is updated other than proguard I think.

sjudd:
Yeah I'd be happy to replace most of the readme with links to the docs site in the long run, but the information should be accurate until we decide what to do. Thanks for sending this!

PR #2047, opened 2017-06-16
Updated README.md

Salad7:
GlideApp.with(this).load("http://goo.gl/gEgYUd").into(imageView); didnt work with my version of Android, Glide.with(this).load("http://goo.gl/gEgYUd").into(imageView); worked fine however

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ Simple use cases with Glide's [generated API][21] will look something like this:
   ...
   ImageView imageView = (ImageView) findViewById(R.id.my_image_view);
 
-  GlideApp.with(this).load("http://goo.gl/gEgYUd").into(imageView);
+  Glide.with(this).load("http://goo.gl/gEgYUd").into(imageView);
 }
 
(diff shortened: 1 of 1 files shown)


TWiStErRob:
@Salad7 it depends on your Glide version and configuration. v4 (master) is `Glide` or `GlideApp`, v3 is `Glide`. GlideApp needs the annotation processor to be enabled. I guess for the most basic use case as that line demonstrates it should be without apt, @sjudd?

sjudd:
That section of the readme mentions that it's using the generated API and points to the documentation page that explains what the generated API is and how to use it.

You could argue that it's better to avoid the generated API in the readme for simplicity, but if you want to do that you'd need to update the entire readme to not use the generated API, including the link and the code immediately below this.

I'm inclined to push the generated API in the long run because it's much nicer to use.

sjudd:
Let's leave this as is for now. Thanks for taking the time, great thought.

PR #2057, opened 2017-06-21
Rename getLifecycle into getGlideLifecycle.

svasilinets:
## Motivation and Context
"getLifecycle()" will be added to fragments from support library in 26*,
to avoid name conflicts getLifecycle in requestManagerFragment is
renamed.


sjudd:
Great thanks!

PR #2061, opened 2017-06-21
Prevent NullPointerException by not setting the static instance

Codeintheory:
until registration is complete

## Description
The Glide instance is not completely thread safe as another thread may get an instance which have not had the components registered

## Motivation and Context
With the current code, Thread 2 may get a glide instance because this.glide != null, but registration in Thread 1 has not completed yet

Thread 1- (Does initialization)
GlideApp.with(context).load(<url>).into(imageView);

Thread 2-
Loader loader = Glide.get(context).getLoaderFactory().buildModelLoader(modelClass, resourceClass);
Glide.with(mAppContext)
                .using(modelLoader, OrbResource.class)
                .from(OrbModel.class)

06-21 14:54:44.301: E/AndroidRuntime(29121): java.lang.NullPointerException: ModelLoader must not be null
06-21 14:54:44.301: E/AndroidRuntime(29121): 	at com.bumptech.glide.provider.FixedLoadProvider.<init>(FixedLoadProvider.java:28)
06-21 14:54:44.301: E/AndroidRuntime(29121): 	at com.bumptech.glide.GenericTranscodeRequest.build(GenericTranscodeRequest.java:42)
06-21 14:54:44.301: E/AndroidRuntime(29121): 	at com.bumptech.glide.GenericTranscodeRequest.<init>(GenericTranscodeRequest.java:60)
06-21 14:54:44.301: E/AndroidRuntime(29121): 	at com.bumptech.glide.RequestManager$GenericModelRequest$GenericTypeRequest.as(RequestManager.java:768)
etc...

### DIFF
--- a/library/src/main/java/com/bumptech/glide/Glide.java
+++ b/library/src/main/java/com/bumptech/glide/Glide.java
@@ -167,10 +167,11 @@ public class Glide {
                     for (GlideModule module : modules) {
                         module.applyOptions(applicationContext, builder);
                     }
-                    glide = builder.createGlide();
+                    Glide glide = builder.createGlide();
                     for (GlideModule module : modules) {
                         module.registerComponents(applicationContext, glide);
                     }
+                    Glide.glide = glide;
                 }
             }
         }


Codeintheory:
Completed the license agreement

Codeintheory:
The same bug is in the 4.0 branch as well as I see two, btw

sjudd:
Nice catch, thanks!

We'll either have to change the interface in Glide v4 or do something more complicated since the Glide object isn't passed in to the Module in v4.

sjudd:
Also license agreement still seems to be pending.

TWiStErRob:
@sjudd same fix would work, all we need is `glide = initGlide(context);` so the read/write/sync is in the same block and not a side effect in the middle of a method. Or you would expect (=allow?!) `registerComponents` to call `Glide.get(context)`?

Codeintheory:
I changed commit to include email as per the requirements of the CLA. Waiting for pending license agreement. Thx

sjudd:
@Codeintheory can you double check the cla? Not sure what's up with the bot, but when I look for the email address in the commit or your username it doesn't find anything. Sorry for the inconvenience...

sjudd:
I'd love to get this in, but still waiting on the cla, @Codeintheory is there some other email address that I should check?

Codeintheory:
@sjudd I've updated my google CLA to match the email with github account. Please check again for codeinthoery@yahoo.com. Thx

Issue #2086, opened 2017-06-28
4.0.0-RC1 - Combined transformations


realdeadfish:
I wrote a custom transformation that uses a `BitmapShader` to "center-crop" an image (an animated GIF) to the outside's width / height of the container, while applying rounded corners to it. This works all nicely now, except that I need another transformation on the same thing, namely a blur transformation.

The issue here is now that I cannot apply the blur transformation before I center-crop the image in the other transformation, because the center-cropping might upscale the image resulting in blur / scaling artifacts.

Due to my transformation's corner rounding however I also cannot apply the blur transformation after the center-crop transformation, because this would naturally blur my just cut-out edges. So I thought, well, maybe I need another pass, even if it's ugly, to again cut-out my image after blurring, so I basically did this:

    CenterCropRoundedCornersTransformation roundCornerTransformation =
              new CenterCropRoundedCornersTransformation(...);
    RequestOptions options = new RequestOptions();
    options.transform(roundCornerTransformation);
    options.transform(new BlurTransformation(...));
    options.transform(roundCornerTransformation);
    Glide.with(context)
              .load(file)
              .apply(options)
              .into(imageView);

Now what struck me was that the `BlurTransformation` code does not seem to get executed at all, i.e. during debugging I never actually got into this transformation's `transform(...)` method. So I dug deeper into the code and figured that for some (to me unknown) reason transformations are registed in a map by their drawable type, where they overwrite each other without further notice. This however is contrary to the fact again that _two_ transformations (one corner and one blur) seem to somehow work flawlessly, but when adding the third the thing falls apart.

I guess you're saying now "it's inefficient to apply more than one transformation" and I guess you're right and if the solution is really to combine cropping, rounding and blurring into one transformation, I'll certainly try to accomplish that, but for now I thought of code re-use, i.e. being able to mix and match single transformations easily together, but that seems to have unexpected boundaries.

realdeadfish:
OK, my fault, it really does not work with two transformations, I just see that my blur transformation actually replaces the original transformation - the (test) images just happen to be scaled properly so not much center-cropping took actually place. So, is there a way to chain transformations then at all?

realdeadfish:
I came up with a `CombinedTransformation` class that I paste here for reference. This just executes any number of transformations one after another:

```
class CombinedTransformation(vararg val transformations: Transformation<Bitmap>) 
        : Transformation<Bitmap> {
    override fun transform(context: Context, resource: Resource<Bitmap>, 
[...]

PR #2138, opened 2017-07-13
Add a shortcut to apply multiple transformations on a single request.

realdeadfish:
## Description
As per discussion here:
https://github.com/bumptech/glide/issues/2086

## Motivation and Context
Makes the possibility to apply multiple transformations easier discoverable via API.

I'd like to have broaden the test context a bit more, but this would have meant to introduce `@VisibleForTesting` getters to internals in a couple of places which I neglected to do for now.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/request/RequestOptions.java
+++ b/library/src/main/java/com/bumptech/glide/request/RequestOptions.java
@@ -9,6 +9,7 @@ import android.support.annotation.Nullable;
 import com.bumptech.glide.Priority;
 import com.bumptech.glide.load.DecodeFormat;
 import com.bumptech.glide.load.Key;
+import com.bumptech.glide.load.MultiTransformation;
 import com.bumptech.glide.load.Option;
 import com.bumptech.glide.load.Options;
 import com.bumptech.glide.load.Transformation;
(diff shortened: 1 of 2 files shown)


realdeadfish:
Signed the CLA.

library/src/main/java/com/bumptech/glide/request/RequestOptions.java
```diff
@@ -877,11 +878,36 @@ public RequestOptions transform(@NonNull Transformation<Bitmap> transformation)
...
+   * and throws an exception if asked to transform an unknown type.
+   * <p>
+   * <p>This will override previous calls to {@link #dontTransform()}.
+   *
+   * @param transformations One or more {@link Transformation}s for
+   *                        {@link Bitmap}s.
+   * @see #optionalTransform(Transformation)
+   * @see #optionalTransform(Class, Transformation)
+   */
+  public RequestOptions multiTransform(@NonNull Transformation<Bitmap>... transformations) {
```
TWiStErRob: I think `transforms` would be a better name (alphabetically closer to transform). Is there any particular reason why not just use an overload to the existing `transform` method and introduce a new name?

library/src/main/java/com/bumptech/glide/request/RequestOptions.java
```diff
@@ -186,8 +189,8 @@ public static RequestOptions signatureOf(@NonNull Key signature) {
   public static RequestOptions fitCenterTransform() {
     if (fitCenterOptions == null) {
       fitCenterOptions = new RequestOptions()
-          .fitCenter()
-          .autoClone();
+              .fitCenter()
+              .autoClone();
```
TWiStErRob: Looks like you may have a different formatter that the project. Glide uses 2-space indents and continuation is usually double indent, so 2*2=4, 8 is weird in this code base.

library/src/main/java/com/bumptech/glide/request/RequestOptions.java
```diff
@@ -804,7 +805,7 @@ public RequestOptions centerInside() {
 
   /**
    * Applies {@link CircleCrop} to all default types, and ignores unknown types.
-   *
+   * <p>
```
TWiStErRob: Paragraph is already there in next line, these shouldn't be added.

sjudd:
Thanks for doing this!

If it helps Glide uses the google style guide: https://google.github.io/styleguide/javaguide.html

I believe you can apply the style guide in intellij/android studio using this: https://github.com/google/styleguide/blob/gh-pages/intellij-java-google-style.xml.

If I can remember exactly how I did it, I'll update our contributing page on the docs site and link that here.

sjudd:
Added steps to import the style guide into Android Studio here (it should update shortly): http://bumptech.github.io/glide/dev/contributing.html

library/src/main/java/com/bumptech/glide/request/RequestOptions.java
realdeadfish: Well, it closer resembled the name of the class that actually does the transformation, but I have no strong obligations there, so I will rename it.

realdeadfish:
Reformatted and cleaned.

library/src/test/java/com/bumptech/glide/request/RequestOptionsTest.java
```diff
@@ -1,9 +1,12 @@
 package com.bumptech.glide.request;
 
-import static com.google.common.truth.Truth.assertThat;
```
sjudd: You'll need to revert this import removal.

realdeadfish:
Sorry, was on vacation, that's why it took a bit longer.

sjudd:
Great thanks for doing this, I appreciate it!

PR #2153, opened 2017-07-20
Update Fetcher documentation

Teovald:
The Fetcher documentation needs to be updated for the new Glide 4 API


Teovald:
Unrelated : should glide use @WorkerThread @MainThread annotations ?
That would make its API simpler to grok

TWiStErRob:
Yes, I support threading annotations, it helps a lot to know if you can do I/O and other stuff safely. All the components of the flow should be annotated. e.g. I never know if Transcoder is on UI or not. Annotations are more prevalent than the `* <p> This will always be called on` comments. Adding UI thread annotation on the listener/target interface will break a lot of people's builds (lint), in a good way!

sjudd:
Great thanks!

PR #2203, opened 2017-08-03
Annotate getDataClass and getDataSource from DataFetcher as NonNull

Tolriq:
## Description

Annotate getDataClass and getDataSource from DataFetcher as NonNull
Returning null from those will trigger obscure crash in MultiClassKey, annotation helps preventing that.

## Motivation and Context

Help devs to avoid pitfalls :)

### DIFF
--- a/integration/okhttp/src/main/java/com/bumptech/glide/integration/okhttp/OkHttpStreamFetcher.java
+++ b/integration/okhttp/src/main/java/com/bumptech/glide/integration/okhttp/OkHttpStreamFetcher.java
@@ -1,5 +1,6 @@
 package com.bumptech.glide.integration.okhttp;
 
+import android.support.annotation.NonNull;
 import android.util.Log;
 import com.bumptech.glide.Priority;
 import com.bumptech.glide.load.DataSource;
@@ -89,11 +90,13 @@ public class OkHttpStreamFetcher implements DataFetcher<InputStream> {
     // TODO: call cancel on the client when this method is called on a background thread. See #257
   }
 
+  @NonNull
   @Override
   public Class<InputStream> getDataClass() {
     return InputStream.class;
   }
 
+  @NonNull
   @Override
   public DataSource getDataSource() {
     return DataSource.REMOTE;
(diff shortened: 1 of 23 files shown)


integration/volley/src/main/java/com/bumptech/glide/integration/volley/VolleyStreamFetcher.java
```diff
@@ -67,11 +67,13 @@ public void cancel() {
     }
   }
 
+  @android.support.annotation.NonNull
```
sjudd: Can you import NonNull instead of referencing it via the package name?

integration/okhttp3/src/main/java/com/bumptech/glide/integration/okhttp3/OkHttpStreamFetcher.java
```diff
@@ -87,11 +87,13 @@ public void cancel() {
     }
   }
 
+  @android.support.annotation.NonNull
```
sjudd: Same here, can you import NonNull?

integration/okhttp/src/main/java/com/bumptech/glide/integration/okhttp/OkHttpStreamFetcher.java
```diff
@@ -89,11 +89,13 @@ public void cancel() {
     // TODO: call cancel on the client when this method is called on a background thread. See #257
   }
 
+  @android.support.annotation.NonNull
```
sjudd: And here too, can you import NonNull?

Tolriq: @sjudd Actually this is strange that the compilation and test passed, the refactoring was made by AS.
Integration libraries does not import Support Annotation libraries.
To import the NonNull I need to add the dependency is it ok for you or drop the annotation in those ?

sjudd: Yeah it's ok to add the dependency, the main library requires the annotations anyway so it's not a new dependency for applications.

Tolriq:
Pr updated.

sjudd:
Great thanks!

Issue #2204, opened 2017-08-03
Infinite error on fast scroll in recycler view.


Tolriq:
**Glide Version**:
4.0.0

**Integration libraries**:
No

**Device/Android Version**:
Android 7 Samsung.

**Issue details / Repro steps / Use case background**: 
Trying v4 before migrating. So very basic use case.
GlideApp with a basic configuration and a custom fetcher migrated from v3.

**Glide load line / `GlideModule` (if any) / list Adapter code (if any)**:
```java
  RequestOptions requestOptions = new RequestOptions();
                requestOptions.placeholder(placeHolder);
GlideApp.with(appContext)
                        .load(getGlideImageRequest(url))
                        .apply(requestOptions)
.into(holder.fanart);
```
The getGlideImageRequest just builds a specific object for internal use.

All works perfectly, but when scrolling very fast with image downloading, I will trigger infinite stack trace shown later. Making it quite hard to find the root cause as everything is spammed to death and no real clue about possible cause :(

**Layout XML**:
```xml
<FrameLayout xmlns:android="...
```

**Stack trace / LogCat**:
```ruby
Request threw uncaught throwable
                                                                                    java.lang.NullPointerException: Attempt to invoke interface method 'int java.lang.Comparable.compareTo(java.lang.Object)' on a null object reference
                                                                                        at java.util.concurrent.PriorityBlockingQueue.siftDownComparable(PriorityBlockingQueue.java:374)
                                                                                        at java.util.concurrent.PriorityBlockingQueue.dequeue(PriorityBlockingQueue.java:303)
                                                                                        at java.util.concurrent.PriorityBlockingQueue.take(PriorityBlockingQueue.java:518)
                                                                                        at java.util.concurrent.ThreadPoolExecutor.getTask(ThreadPoolExecutor.java:1058)
                                                                                        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1118)
                                                                                        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:607)
                                                                                        at java.lang.Thread.run(Thread.java:762)
                                                                                        at com.bumptech.glide.load.engine.executor.GlideExecutor$DefaultThreadFactory$1.run(GlideExecutor.java:347)
```

When scrolling slowly I sometimes get following error and image does not display even if the loader did load it correctly.
```ruby

Request threw uncaught throwable
                                                                                     java.lang.IllegalStateException: Already released
[...]

PR #2208, opened 2017-08-04
Prevent crash due to late calls to DataCallback by clients

Tolriq:
## Description

Prevent crash due to late calls to DataCallback by clients
(https://github.com/bumptech/glide/issues/2204)
## Motivation and Context

Errors generated by too late calls are really hard to find.
This solution prevent the problem by securing the calls.

Another approach may be to throw a proper error to help devs find the root cause, but as crashing on those points triggers really strange things afterwards it's not really something I can address for now.


sjudd:
Useful context, but I'd prefer concrete error messages when subcomponents do unexpected things, closing for now, we can follow up in the issue.

PR #2277, opened 2017-08-21
Add a clear note about model needing to implement hashCode and equals.

Tolriq:
See title.

I'm not native English, so can update phrasing according to your needs.


sjudd:
Looks reasonable to me, thanks!

PR #2293, opened 2017-08-26
Update model on failed request update.

### LINKED ISSUE #2270: Follow up of: https://github.com/bumptech/glide/pull/2261
@sjudd Ok so there's one last side effect remaining after https://github.com/bumptech/glide/issues/2245

Opening a new issue as it's a special case, that will maybe needs something on my side to workaround.

As explained on previous issues, I do have a special model that handles some connexion states internaly.
And those states as they do not change the actual picture data, are not handled in the model hashcode/equals so that memory cache can properly be used.

The problem is that the restart of the query after your patch still use the previous model and not the new one with the updated internal flags so does not work.

1) The restart can maybe be updated to use the new model, even if for Glide it's the same as model could have internal states.
2) Adding a way to indicate a cacheKey for the model. To allow different hashcode/equals to indicate when the model is actually different, but still allow memory cache to use the same data as the picture is the same
3) Something else?

Tolriq:
## Description

Ensure that restarted request use the updated model when failed/cancelled as it can contains data to help the request to succeed.

## Motivation and Context

Mostly fixes #2270

Since I had to push a beta for my app to fix some unseen Android O issues and wanted to start to get feedback about Glide 4, I went with this temporary solution to have all working. I know you have not made any decisions yet, but I PR this as is to have a reference since it's now distributed to end users.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/RequestBuilder.java
+++ b/library/src/main/java/com/bumptech/glide/RequestBuilder.java
@@ -359,11 +359,15 @@ public class RequestBuilder<TranscodeType> implements Cloneable {
     Request request = buildRequest(target);
 
     Request previous = target.getRequest();
-    if (request.isEquivalentTo(previous)) {
+    // When request was failed or cancelled, be sure to use the updated model as it can contains
+    // unexposed data that could help the request to succeed on restart.
+    // See https://github.com/bumptech/glide/issues/2270
+    if (request.isEquivalentTo(previous)
+      && (Preconditions.checkNotNull(previous).isComplete()
+         || Preconditions.checkNotNull(previous).isRunning())) {
       request.recycle();
       // If the request is completed, beginning again will ensure the result is re-delivered,
-      // triggering RequestListeners and Targets. If the request is failed, beginning again will
-      // restart the request, giving it another chance to complete. If the request is already
+      // triggering RequestListeners and Targets. If the request is already
       // running, we can let it continue running without interruption.
       if (!Preconditions.checkNotNull(previous).isRunning()) {
         previous.begin();


library/src/main/java/com/bumptech/glide/RequestBuilder.java
```diff
@@ -359,11 +359,12 @@ protected RequestOptions getMutableOptions() {
     Request request = buildRequest(target);
 
     Request previous = target.getRequest();
-    if (request.isEquivalentTo(previous)) {
+    if (request.isEquivalentTo(previous)
```
sjudd: Add a comment that references #2270 as to why this is necessary.

sjudd:
I'm willing to merge this. It's something that might go away in the future if it becomes burdensome, but I can't think of anything that would conflict with this behavior right now at least.

We should probably also try to follow up on the flashing when you manually clear() and restart requests separately. Maybe there's something else we can do to mitigate that case as well.

Tolriq:
Comment added.

To be honest fixing the flashing with clear would be cool as certainly can fix other things, but having devs to always call clear when their model have private fields that can't be exposed due to hashCode() seems way more hackish than this :(

Maybe another way would be to have an optional interface on the model to test for equity and call it if the model have it instead of equals for the request restart.
Kinda the same reasoning for your usage of isEquivalentTo and not equals. And have your isEquivalentTo call that one when available.

sjudd:
Great thanks!

Can you file an issue with some details on your views that might help me track down why the flashing is occurring for your use case when you do clear? Is it in a RecyclerView? If so, are the views dimensions variable?

An optional interface is definitely worth more thought. Let's try to at least understand the flashing a bit better and go from there.

PR #2339, opened 2017-09-06
dont override the placeholder from xml if user didnt set it in Glide …

richardissuperman:
``
@Override
    public void onLoadStarted(Drawable placeholder) {
        if(placeholder != null){
        view.setImageDrawable(placeholder);
        }
    }
``

## Description
prevent glide for overriding placeholder when user didnt set the placeholder by using
``
placeholder(int res);
``
if user didnt set placeholder in Glide code, it will end up making imageview's drawable to null which leads the imageview to have 0 width and 0 height(if wrap_contnet)

## Motivation and Context

if a user has a imageview which set to wrap_content for both width and height(with a placeholder in xml) , and he didnt add placeholder() method , he will end up having a imageview which is 0 width and 0 height.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/request/target/ImageViewTarget.java
+++ b/library/src/main/java/com/bumptech/glide/request/target/ImageViewTarget.java
@@ -45,7 +45,7 @@ public abstract class ImageViewTarget<Z> extends ViewTarget<ImageView, Z> implem
      */
     @Override
     public void onLoadStarted(Drawable placeholder) {
-        if(placeholder != null){
+        if(placeholder != null && view.getDrawable() != null){
         view.setImageDrawable(placeholder);
         }
     }
--- a/library/src/main/java/com/bumptech/glide/request/target/ImageViewTarget.java
+++ b/library/src/main/java/com/bumptech/glide/request/target/ImageViewTarget.java
@@ -45,11 +45,11 @@ public abstract class ImageViewTarget<Z> extends ViewTarget<ImageView, Z> implem
      */
     @Override
     public void onLoadStarted(Drawable placeholder) {
-        if(placeholder != null && view.getDrawable() != null){
-        view.setImageDrawable(placeholder);
+        if (placeholder == null && view.getDrawable() != null) {
+            return;
         }
+        view.setImageDrawable(placeholder);
     }
-
     /**
      * Sets the given {@link android.graphics.drawable.Drawable} on the view using
      * {@link android.widget.ImageView#setImageDrawable(android.graphics.drawable.Drawable)}.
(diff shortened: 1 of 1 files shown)


richardissuperman:
I signed it

sjudd:
This is a non-trivial change in behavior. How can we tell if the user intends for us to override any image in the View or not?

richardissuperman:
do you think if i can check if the imageview has a drawable then proceed to the check ? something like this:

``
@Override
    public void onLoadStarted(Drawable placeholder) {
        if(placeholder != null && view.getDrawable() != null){
        view.setImageDrawable(placeholder);
        }
    }
``

cause when user set src in xml they dont expect Glide to override it to null right? 
Thats my only concern, if you think its not worth it to change then you can close this pr 

Thank you so much 😸

library/src/main/java/com/bumptech/glide/request/target/ImageViewTarget.java
```diff
@@ -45,11 +45,11 @@ public void setDrawable(Drawable drawable) {
      */
     @Override
     public void onLoadStarted(Drawable placeholder) {
-        if(placeholder != null && view.getDrawable() != null){
-        view.setImageDrawable(placeholder);
+        if (placeholder == null && view.getDrawable() != null) {
```
richardissuperman: @sjudd if placeholder is null but imageview itself has drawable already, skip this method,thanks

library/src/main/java/com/bumptech/glide/request/target/ImageViewTarget.java
```diff
@@ -45,9 +45,11 @@ public void setDrawable(Drawable drawable) {
      */
     @Override
     public void onLoadStarted(Drawable placeholder) {
+        if (placeholder == null && view.getDrawable() != null) {
```
sjudd: After more thought, we can't make this change. A typical use case where this will occur and you won't want it is in a RecyclerView or ListView. The previous image may have been loaded into the ImageView. If you don't clear it out, even when the placeholder is null, you're going to end up showing the old image while the new one loads, which isn't right.

I'm not sure there's going to be a good way to generalize this, I think you're going to be stuck passing in the placeholder to Glide.

richardissuperman: @sjudd yeah i think you are right, have to clear it in recyclerview otherwise it will mess up, make sense!
We will just end up using placeholder() and thumbnail() together .
Thanks

sjudd: Thank you, I appreciate you putting this up anyway, it's super helpful to see exactly what you're trying to do.

TWiStErRob:
@richardissuperman there's also another problem by assuming user wants to see XML set image. Some people use `android:src` to set up the preview in AS, because they're likely not aware of the `tools:` namespace and the performance penalty not using it incurs.

richardissuperman:
thanks @TWiStErRob  it makes sense also, i guess i just need to use the placeholder() method ;)

PR #2349, opened 2017-09-07
Override default Encoders

MistaGreen:
## Description

Override default Encoders for existing data classes. I've added prepend methods to Encoder and ResourceEncoder classes to achieve this.  

## Motivation and Context

I want to encrypt images in Glide disk cache. To do that I need to be able to replace Encoders for existing data types like Bitmap, InputStream, Drawable, etc.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/Registry.java
+++ b/library/src/main/java/com/bumptech/glide/Registry.java
@@ -23,6 +23,8 @@ import com.bumptech.glide.provider.ModelToResourceClassCache;
 import com.bumptech.glide.provider.ResourceDecoderRegistry;
 import com.bumptech.glide.provider.ResourceEncoderRegistry;
 import com.bumptech.glide.util.pool.FactoryPools;
+
+import java.io.File;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


MistaGreen:
I signed it

library/src/main/java/com/bumptech/glide/Registry.java
```diff
@@ -69,7 +71,23 @@ public Registry() {
...
+    return this;
+  }
+
+  /**
+   * Prepends the given {@link Encoder} into the list of available {@link Encoder}s
+   * so that it is attempted before all later and default {@link Encoder}s for the given
+   * data class.
+   *
+   * <p>This method allows you to replace the default {@link Encoder} because it ensures
+   * the registered {@link Encoder} will run first. You can return false in
```
sjudd: Returning false will cause the load to fail. There isn't currently any logic that tries Encoders in order, it just assumes there's one per data type.

sjudd: https://github.com/bumptech/glide/blob/35c28c28ca2a030e90df5aac67b0356c769cd623/library/src/main/java/com/bumptech/glide/load/engine/SourceGenerator.java#L74

library/src/main/java/com/bumptech/glide/Registry.java
```diff
@@ -144,7 +162,27 @@ public Registry() {
...
+
+  /**
+   * Registers a new {@link com.bumptech.glide.load.data.DataRewinder.Factory} to handle a
+   * non-default data type that can be rewind to allow for efficient reads of file headers.
+   *
+   * Prepends the given {@link ResourceEncoder} into the list of available {@link ResourceEncoder}s
+   * so that it is attempted before all later and default {@link ResourceEncoder}s for the given
+   * data type.
+   *
+   * <p>This method allows you to replace the default {@link ResourceEncoder} because it ensures
```
sjudd: Same here: https://github.com/bumptech/glide/blob/35c28c28ca2a030e90df5aac67b0356c769cd623/library/src/main/java/com/bumptech/glide/load/engine/DecodeJob.java#L627

sjudd:
Other than the two minor comments I made, looks good. Thanks!

sjudd:
LGTM, thanks!

Issue #2307, opened 2017-08-30
[Feature request] Add a way to differentiate model equity from image equity.


Tolriq:
Follow up of many discussions :)

Last proposal was to add an optional interface to model that would implement an isEquivalentTo that would be used if present to determine if a request is different or not to restart it.

This allows models to point to same image cache but still have different internal data to for example get the actual image from different place or with different arguments.

sjudd:
Fixed by #2353 I believe.

PR #2353, opened 2017-09-07
Implement a new optional model interface ModelEquity.

Tolriq:
## Description

This allows models to handle equity for internal states that can be different from the target image result that is handled with equals()/hashCode()
For example to allow restart of image download from a new source but still using the same cache in Glide

## Motivation and Context

Addressing #2307

### DIFF
--- a/library/src/main/java/com/bumptech/glide/RequestBuilder.java
+++ b/library/src/main/java/com/bumptech/glide/RequestBuilder.java
@@ -359,15 +359,11 @@ public class RequestBuilder<TranscodeType> implements Cloneable {
     Request request = buildRequest(target);
 
     Request previous = target.getRequest();
-    // When request was failed or cancelled, be sure to use the updated model as it can contains
-    // unexposed data that could help the request to succeed on restart.
-    // See https://github.com/bumptech/glide/issues/2270
-    if (request.isEquivalentTo(previous)
-      && (Preconditions.checkNotNull(previous).isComplete()
-         || Preconditions.checkNotNull(previous).isRunning())) {
+    if (request.isEquivalentTo(previous)) {
       request.recycle();
       // If the request is completed, beginning again will ensure the result is re-delivered,
-      // triggering RequestListeners and Targets. If the request is already
+      // triggering RequestListeners and Targets. If the request is failed, beginning again will
+      // restart the request, giving it another chance to complete. If the request is already
       // running, we can let it continue running without interruption.
       if (!Preconditions.checkNotNull(previous).isRunning()) {
         previous.begin();
--- a/library/src/main/java/com/bumptech/glide/load/model/Model.java
+++ b/library/src/main/java/com/bumptech/glide/load/model/Model.java
@@ -1,0 +1,28 @@
+package com.bumptech.glide.load.model;
+
+/**
+ * An optional interface that models can implement to enhance control over Glide behaviors.
+ */
+public interface Model {
+
+    /**
+     * Returns {@code true} if this model produces the same image using the same mechanism
+     * (server, authentication, source etc) as the given model.
+     * <p>
+     * Models must also implement {@link #equals(Object other)} and {@link #hashCode()}
+     * to ensure that caching functions correctly.
+     * If this object returns {@code true} from this method for a given Model,
+     * it must also be equal to and have the same hash code as the given model.
+     * <p>
+     * However, this model may be equal to and have the same hash code as a given model
+     * but still return {@code false} from this method.
+     * This method optionally allows you to differentiate between Models that load the same image
+     * via multiple different means.
+     * For example one Model might load the image from server A and another model might load
+     * the same image from server B.
+     * The models must be equal to each other with the same hash code because they load
+     * the same image. However two requests made with the different models are not exactly the
+     * same because the way the image is loaded will differ.
+     */
+    boolean isEquivalentTo(Object other);
+}
--- a/library/src/main/java/com/bumptech/glide/util/Util.java
+++ b/library/src/main/java/com/bumptech/glide/util/Util.java
@@ -4,6 +4,8 @@ import android.annotation.TargetApi;
 import android.graphics.Bitmap;
 import android.os.Build;
 import android.os.Looper;
+
+import com.bumptech.glide.load.model.Model;
 import com.bumptech.glide.request.target.Target;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
@@ -188,6 +190,16 @@ public final class Util {
    */
   public static boolean bothNullOrEqual(Object a, Object b) {
     return a == null ? b == null : a.equals(b);
+  }
+
+  public static boolean bothModelsNullEquivalentOrEquals(Object a, Object b) {
+    if (a == null) {
+      return b == null;
+    }
+    if (a instanceof Model) {
+      return ((Model) a).isEquivalentTo(b);
+    }
+    return a.equals(b);
   }
 
   public static int hashCode(int value) {
--- a/library/src/main/java/com/bumptech/glide/request/SingleRequest.java
+++ b/library/src/main/java/com/bumptech/glide/request/SingleRequest.java
@@ -578,7 +578,7 @@ public final class SingleRequest<R> implements Request,
       SingleRequest that = (SingleRequest) o;
       return overrideWidth == that.overrideWidth
           && overrideHeight == that.overrideHeight
-          && Util.bothNullOrEqual(model, that.model)
+          && Util.bothModelsNullEquivalentOrEquals(model, that.model)
           && transcodeClass.equals(that.transcodeClass)
           && requestOptions.equals(that.requestOptions)
           && priority == that.priority;


Tolriq:
@sjudd Since I can no more reproduce the flickering I went ahead and added the model equity. (I start to think that it was caused by memory pressure due to another bug).

JavaDoc is minimal as I'm not English native so, if you have requirements / details for them please tell.

library/src/main/java/com/bumptech/glide/util/Util.java
```diff
@@ -190,6 +192,16 @@ public static boolean bothNullOrEqual(Object a, Object b) {
     return a == null ? b == null : a.equals(b);
   }
 
+  public static boolean bothModelsNullOrEqual(Object a, Object b) {
```
TWiStErRob: `bothNullOrEquivalent`

library/src/main/java/com/bumptech/glide/RequestBuilder.java
```diff
@@ -359,15 +359,11 @@ protected RequestOptions getMutableOptions() {
...
-    // unexposed data that could help the request to succeed on restart.
-    // See https://github.com/bumptech/glide/issues/2270
-    if (request.isEquivalentTo(previous)
-      && (Preconditions.checkNotNull(previous).isComplete()
-         || Preconditions.checkNotNull(previous).isRunning())) {
+    if (request.isEquivalentTo(previous)) {
       request.recycle();
       // If the request is completed, beginning again will ensure the result is re-delivered,
-      // triggering RequestListeners and Targets. If the request is already
+      // triggering RequestListeners and Targets. If the request is failed, beginning again will
```
sjudd: If the request has failed, calling into again will restart the request...

library/src/main/java/com/bumptech/glide/load/model/ModelEquity.java
```diff
@@ -0,0 +1,16 @@
+package com.bumptech.glide.load.model;
+
+/**
+ * An optional interface that models can implement to handle difference in equity
+ * between the model and the target.
+ */
+public interface ModelEquity {
```
sjudd: I might just call this Model. We reference Model all over the place in the documentation and it's plausible we might expand this interface or make it required in a future version. What do you think?

library/src/main/java/com/bumptech/glide/load/model/ModelEquity.java
```diff
@@ -0,0 +1,16 @@
...
+
+    /**
+     * Returns {@code true} if this Model is equivalent to the given Model.
+     *
+     * @param other the model to be compared to
```
sjudd: you can drop these two comments since you're describing them already in your summary statement above.

library/src/main/java/com/bumptech/glide/load/model/ModelEquity.java
```diff
@@ -0,0 +1,16 @@
+package com.bumptech.glide.load.model;
+
+/**
+ * An optional interface that models can implement to handle difference in equity
+ * between the model and the target.
+ */
+public interface ModelEquity {
+
+    /**
+     * Returns {@code true} if this Model is equivalent to the given Model.
```
sjudd: Might be worth an expanded comment on what equivalent to is. 

Maybe 
``` 
Returns {@code true} if this Model produces the same image as the equivalent Model to avoid unnecessarily restarting a Request
```

library/src/main/java/com/bumptech/glide/RequestBuilder.java
Tolriq: Actually I restored your previous comment I think you were talking about the begin call after the if.

library/src/main/java/com/bumptech/glide/load/model/ModelEquity.java
Tolriq: Yes Model is Ok if there's plan to expand.

library/src/main/java/com/bumptech/glide/load/model/ModelEquity.java
Tolriq: It's more complicated than that :) Even after all the explanations about the need :p 

Equals/HashCode represent the fact that the model produce the same image.
isEquivalentTo is a step further it represent the fact that the model use the same way to produce the same image.

But we need to be sure that users still implements correctly hashCode/Equals for the cache part and the equity of resulting image. Not sure how to properly word all this.

sjudd: One more stab:

----
Returns {@code true} if this model produces the same image using the same mechanism (server, authentication, source etc) as the given model.

<p>Models must also implement {@link #equals()} and {@link #hashCode()} to ensure that caching functions correctly.

<p>If this object returns {@code true} from this method for a given Model, it must also be equal to and have the same hash code as the given model. However, this model may be equal to and have the same hash code as a given model but still return {@code false} from this method.  

<p>This method optionally allows you to differentiate between Models that load the same image via multiple different means. For example one Model might load the image from server A and another model might load the same image from server B. The models must be equal to each other with the same hash code because they load the same image. However two requests made with the different models are not exactly the same because the way the image is loaded will differ.
-----

Just realized in writing this actually that this isn't entirely sufficient. Request deduplication in Engine.java is done based on equals/hashCode. If you start two different requests with Models that return true from equals but false from equivalentTo, the Engine will deduplicate the one that starts second into the first one and only allow the first one to run (the second will simply be notified when the first finishes or fails). 

It doesn't matter if you always cancel the first request before the second, but if you were to do something like start two requests, one for each equal but not equivalent model, simultaneously into two different targets, it wouldn't do quite what you expect.

Tolriq: I'm not sure Engine should be touched as the target image is the same. 
So if 2 different models points to the same target image most of the time it's better to not fire 2 network / cache retrieval when only 1 is enough.

Only edge case would be 2 different path for same images, called at the exact same time. But one  path is down and is started just before the other.
This leads to 2 fails instead of 1 fail 1 success.

Maybe this edge case should be addressed, but always using isEquivalentTo would IMO be worse as more request than necessary most of the time.

I don't know yet how engine works, but proper way to address would be to only make first call, then on failure if not Equivalent start second call. (Maybe not possible depending on what data is still available when you currently drop the second query).

Tolriq: I forget to tell, but PR updated without touching Engine, tell me if / what you want to do for that part. (Maybe as a second PR).

library/src/main/java/com/bumptech/glide/load/model/Model.java
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * An optional interface that models can implement to enhance control over Glide behaviors.
+ */
+public interface Model {
+
+    /**
+     * Returns {@code true} if this model produces the same image using the same mechanism
+     * (server, authentication, source etc) as the given model.
+     * <p/>
```
sjudd: <p>

library/src/main/java/com/bumptech/glide/load/model/Model.java
```diff
@@ -0,0 +1,28 @@
...
+     * Models must also implement {@link #equals(Object other)} and {@link #hashCode()}
+     * to ensure that caching functions correctly.
+     * If this object returns {@code true} from this method for a given Model,
+     * it must also be equal to and have the same hash code as the given model.
+     * <p/>
```
sjudd: <p>

library/src/main/java/com/bumptech/glide/util/Util.java
sjudd: +1

sjudd: Actually bothModelsNullOrEquivalent is probably even better if that's not with @TWiStErRob  meant.

Tolriq: I did not change as it's also Equals if Equivalent not implemented and put the 3 things sounded odd :( Or I add a javadoc to the function ?

library/src/main/java/com/bumptech/glide/load/model/Model.java
Tolriq: @sjudd There's actually no text with your comment so no idea what you want :)

sjudd: haha my bad, ``<p>``. I forgot that would just get converted into whitespace :)

library/src/main/java/com/bumptech/glide/util/Util.java
sjudd: bothNullEqualOrEquivalent?

Long name here is fine, it's kind of an unusual function. If you think it's best as is, I'm fine with that though, it's hard to explain this in a single method name.

Tolriq:
@sjudd PR updated, when the longest name as you said it's not a problem :)

Do you want that I open another issue to discuss about the request Request deduplication?
I tried to find where such thing would occurs but was not able to find by myself :(

sjudd:
Yeah might as well. The issue is here: https://github.com/bumptech/glide/blob/a4af922db56c928a56bcb8bfc7651e32fe10fd10/library/src/main/java/com/bumptech/glide/load/engine/Engine.java#L184.

Basically the rest of Glide uses equals/hashcode to compare models. I don't think I'd want to what we're doing here everywhere else. We may just have to live with it.

Tolriq:
Well equals/hashcode are OK for 99% of the time in other places as only the result matters.

Only issue would be 2 same target images with different path at the same time with the 1st one starting being a down path.

This leads to 2 failure instead of 1 fail / 1 success. Not a big deal for 99,99999% of use case I think.

And reading code, model is lost when reusing the job.
Implementing a fully working solution (Auto restart of second job if first one fails if model are not Equivalent) sounds overly complicated for this edge case. Let's keep things like that. And see if someone complains. The model equity already greatly improve controls for devs that needs it.

PR #2363, opened 2017-09-09
Repackage annotation compiler dependencies using org.anarres.jarjar

### LINKED ISSUE #2059: Conflicting dependencies in annotation processors
**Glide Version**: 4.0.0-RC1

**Integration libraries**: OkHttp3 

I'm trying migrate to v4 and got stuck with auto-generated GlideApp. 

```java
GlideApp.with...
```
**I'v got a Glide Module class:**
```java
@GlideModule
public final class MyGlideModule extends AppGlideModule {


  @Inject 
  @Named("GLIDE_CLIENT")
  protected OkHttpClient client;


    public MyGlideModule() {
        AccountComponent().inject(this);
    }

    @Override
        builder.setMemoryCache(new LruResourceCache(10 * 1024 * 1024));
    }

    @Override
    public void registerComponents(Context context, Registry glide) {
        glide.replace(GlideUrl.class, InputStream.class, new OkHttpUrlLoader.Factory(client));
        glide.register(SVG.class, PictureDrawable.class, new ImageLoadingUtils.SvgDrawableTranscoder())
                .append(InputStream.class, SVG.class, new ImageLoadingUtils.SvgDecoder());
    }

    @Override
    public boolean isManifestParsingEnabled() {
        return false;
    }
}
```
In my build.gradle file i have those dependencies
```groovy
    compile "com.github.bumptech.glide:glide:4.0.0-RC1"
    compile 'com.android.support:support-v4:25.3.1'
    annotationProcessor 'com.github.bumptech.glide:compiler:4.0.0-RC1'
    compile 'com.caverock:androidsvg:1.2.1'
```

**Stack trace / LogCat**:
```ruby
Error:(83, 4) error: cannot find symbol variable GlideApp
```

Also i cloned SVG's samples and there autogeneration works fine. Can Dagger2 conflict with Glide? Or any suggetions, how can i fix it?

TWiStErRob:
## Description

Use a gradle plugin to do the repcakaging. The drawback of this approach is that we need to change the source code. **I suggest no to merge this, added as historical evidence only; #2364 is a better approach**

## Motivation and Context

Fixes #2059


sjudd:
Thanks @TWiStErRob. I agree the other approach is better. In particular I'd have a hard time maintaining an internal copy of these libraries and dealing with the package name changes.

PR #2364, opened 2017-09-09
Repackage annotation compiler dependencies using jarjar directly

### LINKED ISSUE #2059: Conflicting dependencies in annotation processors

TWiStErRob:
## Description

Add two more steps to build:
 * repackage (change and fatjar packages)
 * ProGuard shrink (remove unused classes)

Sizes of `compiler-4.2.0-SNAPSHOT.jar`:

| Name | Size | Comment |
|---|---|---|
| Original | 72,239 | raw classes from this module |
| Repackaged | 2,648,476 | (incl. Guava, Javapoet, Auto) |
| Proguarded | 251,481 | dontobfuscate,dontoptimize,shrink; keep all module code |

The only drawback I see is the inclusion of `jarjar` JAR in the libs folder, only 1.3 is published on Maven Central.

## Motivation and Context

Fixes #2059


TWiStErRob:
Hmm, actually, the whole problem stems from autoservice pulling in Guava, is it really worth it to have this dependency? Note that replacing the `import`, `@AutoService` and `compile 'auto-service'` (3 lines + 2MB of dependencies) with a file named `src/main/resources/META-INF/services/javax.annotation.processing.Processor` containing the fully qualified class name (1 line) would solve the issue as well.

TWiStErRob:
**Warning**: I didn't test-run this change, only that the .jar file is built correctly and similar to the original.

sjudd:
Thanks @TWiStErRob, working on testing this.

sjudd:
For posterity, #2363 lets us use a gradle plugin, but requires modifying Glide's source to accomplish the same thing.

annotation/compiler/build.gradle
```diff
@@ -14,4 +26,73 @@ dependencies {
     testCompile 'com.google.testing.compile:compile-testing:0.10'
```
sjudd: Remove the two testCompile lines (they're totally unused anyway).

annotation/compiler/build.gradle
```diff
@@ -1,10 +1,22 @@
...
 dependencies {
-    compile 'com.squareup:javapoet:1.9.0'
-    compile 'com.google.auto.service:auto-service:1.0-rc3'
+    // from https://code.google.com/archive/p/jarjar/downloads
+    jarjar files('libs/jarjar-1.4.jar')
+
+    compileOnly 'com.squareup:javapoet:1.9.0'
+    compileOnly 'com.google.auto.service:auto-service:1.0-rc3'
+
     compile 'com.google.code.findbugs:jsr305:3.0.1'
```
sjudd: compileOnly here is fine since these are just annotations anyway. Doing so will resolve the findbugs conflict as well as those with javapoet and auto-service.

sjudd:
@TWiStErRob let me know if you want me to take this over, thank you for putting this up. Great idea! I really appreciate the help.

sjudd:
Merged here: a0e388e234919cb4c94ea391811b5314ac5a9654. Thanks!

PR #2367, opened 2017-09-10
Avoid redundant placeholder calls when replacing requests.

sjudd:
When replacing an old Request with a new Request for a particular View or Target, avoid calling Target#onLoadCleared(Drawable).

The benefit is twofold:
1. We avoid the wasted work of setting a placeholder drawable in the old request (the new request will do so in onLoadStarted anyway).
2. We avoid calling requestLayout on the View, which allows us to use its old size to load the new image, instead of waiting for a layout pass. In some cases this delay can cause flashing while we wait for the unnecessary layout pass to complete.

The problem is that Targets may not get onLoadCleared when they expect to. I'm not sure that anyone really cares about this, but it's definitely a behavior change and if you were relying on onLoadCleared it would be hard to understand why it's not called. This behavior change also only really makes sense for ViewTarget implementations.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/Glide.java
+++ b/library/src/main/java/com/bumptech/glide/Glide.java
@@ -635,10 +635,10 @@ public class Glide implements ComponentCallbacks2 {
     return registry;
   }
 
-  void removeFromManagers(Target<?> target) {
+  void removeFromManagers(Target<?> target, boolean setPlaceholder) {
     synchronized (managers) {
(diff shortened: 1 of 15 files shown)


sjudd:
I'd welcome comments here if anyone has any thoughts one way or another. If you're relying on ``onLoadCleared`` callbacks in ``Target``, definitely let me know.

sjudd:
After more thought, I think this is too unsafe for custom targets, closing for now.

PR #2400, opened 2017-09-16
Change {{ site.url }}/glide into {{ site.baseurl }}.

Muyangmin:
## Description
Hello @sjudd , I've changed all links that started with `{{ site.url}}/glide` to `{{ site.baseurl }}`.
**This change does not affect glide official documentation pages, but it solves a problem when migrating to other languages**. The problem detail is described as below.

## Motivation and Context
In current version of glide documentation, there are two way to link to javadoc:
1. [javadocs.md](https://raw.githubusercontent.com/bumptech/glide/gh-pages/_posts/2015-05-17-javadocs.md)
```Text
[1]:{{ site.baseurl }}{% link /javadocs/410/index.html %}
```
2. [download-setup.md](https://raw.githubusercontent.com/bumptech/glide/gh-pages/_posts/2015-05-15-download-setup.md) 
```Text
[6]: {{ site.url }}/glide/doc/generatedapi.html#kotlin
```

Both of them works well in sjudd.io and bumptech.io, since the doc site is happen to be the same as hardcoded `glide`.

Unfoturenately our Simplified Chinese project is named `glide-docs-cn`, and the gh-pages is not enabled on the fork, we'll get a **404** error while trying  to click the second type of links. But `site.baseurl` will work as expected.

In hack maybe I can copy a `javadocs` directory in my `glide` fork, or even change the Chinese doc repo to `glide`, but obviously that changes the semantic of fork repo and bringing new maintain costs. 

So I just change those links and have tried it in both local and github pages, and by the way suggest to use baseurl pattern in the future docs. 

If any trouble this PR will introduce, please figure out, thank you very much.

### DIFF
--- a/_posts/2015-05-15-download-setup.md
+++ b/_posts/2015-05-15-download-setup.md
@@ -102,4 +102,4 @@ See the [generated API][6] page for details.
 [3]: https://source.android.com/source/jack
 [4]: https://android-developers.googleblog.com/2017/03/future-of-java-8-language-feature.html
 [5]: https://developer.android.com/studio/write/java8-support.html
-[6]: {{ site.url }}/glide/doc/generatedapi.html#kotlin
+[6]: {{ site.baseurl }}/doc/generatedapi.html#kotlin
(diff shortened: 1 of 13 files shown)


sjudd:
Nice! Thanks for pointing that out. I didn't even notice the difference... This seems better in general, I'll use it moving forward.

PR #2405, opened 2017-09-18
Add link to Simplified Chinese Doc.

Muyangmin:
## Description

This is a simple PR to link to Simplified Chinese documentation site. Details are already discussed in #2377 .

PR #2409, opened 2017-09-19
Add multi language documentation link support.

Muyangmin:
Hi, I've add a link to Chinese doc in header.html, as is advised in #2405 . I think maybe this is the best position to put global links. 

Also for future *possible* other language translations, I treat the config as an array(as shown below):
```YAML
# in config.yml
# URL to documentation for other language(s), used in _includes/header.html
otherlang: [
  ['中文版', 'https://muyangmin.github.io/glide-docs-cn']
]
```
To add another language version, just add a row in this file is enough.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -17,6 +17,11 @@ navigation: 2
 # URL to source code, used in _includes/footer.html
 codeurl: 'https://github.com/bumptech/glide'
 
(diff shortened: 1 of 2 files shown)


sjudd:
Looks good, thanks! I appreciate the extra step for other languages.

Issue #2403, opened 2017-09-16
UncaughtThrowableStrategy API / extension 


stephanenicolas:
In our crash reporting tool, we can see this:

![image](https://user-images.githubusercontent.com/1832925/30516489-adaee8dc-9af5-11e7-85ed-8d481e56c4a9.png)

The threads created by GlideExecutor thread factory crash, mostly because of OOMs, but they actually crash the app and are caught by the crash reporting general uncaught exception handler.

It would be valuable for us to be able to pass a custom UncaughtExceptionStrategy to the glide executor, that would simply log the crash, without crashing the app. Our app can live with an image that could not be loaded.

Is it possible to create an API so that we can use a custom error handler here:
https://github.com/bumptech/glide/blob/master/library/src/main/java/com/bumptech/glide/load/engine/executor/GlideExecutor.java#L349

If you agree with this feature, we will gladly contribute a PR for it.

sjudd:
Yeah I think that's totally reasonable.

You could probably work around it with sufficient hackery in a default uncaught exception handler, but you'd probably have to do ugly things like assume the names of the threads Glide creates.

Go for it, thanks for offering.

PR #2411, opened 2017-09-19
issue #2403: create a UncaughtThrowableStrategy API

stephanenicolas:
Solves issue #2403.

The implementation could have taken other forms. I preserved the original API and added new constructors to the `GlideExecutor`. I find the result a bit bloated but it should do. Let me know if you have any feedback or changes you would like us to do. Maybe a better solution would have been to define a default strategy in the executor directly. Up to you !

### DIFF
--- a/library/src/main/java/com/bumptech/glide/GlideBuilder.java
+++ b/library/src/main/java/com/bumptech/glide/GlideBuilder.java
@@ -18,7 +18,6 @@ import com.bumptech.glide.load.engine.cache.LruResourceCache;
 import com.bumptech.glide.load.engine.cache.MemoryCache;
 import com.bumptech.glide.load.engine.cache.MemorySizeCalculator;
 import com.bumptech.glide.load.engine.executor.GlideExecutor;
-import com.bumptech.glide.load.engine.executor.GlideExecutor.UncaughtThrowableStrategy;
 import com.bumptech.glide.manager.ConnectivityMonitorFactory;
 import com.bumptech.glide.manager.DefaultConnectivityMonitorFactory;
 import com.bumptech.glide.manager.RequestManagerRetriever;
@@ -37,7 +36,6 @@ public final class GlideBuilder {
   private MemoryCache memoryCache;
   private GlideExecutor sourceExecutor;
   private GlideExecutor diskCacheExecutor;
-  private UncaughtThrowableStrategy uncaughtThrowableStrategy;
   private DiskCache.Factory diskCacheFactory;
   private MemorySizeCalculator memorySizeCalculator;
   private ConnectivityMonitorFactory connectivityMonitorFactory;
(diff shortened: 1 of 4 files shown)


library/src/main/java/com/bumptech/glide/load/engine/executor/GlideExecutor.java
```diff
@@ -274,29 +311,33 @@ public boolean accept(File file, String s) {
    * A strategy for handling unexpected and uncaught {@link Throwable}s thrown by futures run on the
    * pool.
    */
-  public enum UncaughtThrowableStrategy {
+  public abstract static class UncaughtThrowableStrategy {
```
sjudd: Probably just an interface is ok here?

library/src/main/java/com/bumptech/glide/GlideBuilder.java
```diff
@@ -292,10 +306,16 @@ GlideBuilder setEngine(Engine engine) {
...
+      if (uncaughtThrowableStrategy != null) {
+        sourceExecutor = GlideExecutor.newSourceExecutor(uncaughtThrowableStrategy);
+      }
       sourceExecutor = GlideExecutor.newSourceExecutor();
     }
 
     if (diskCacheExecutor == null) {
+      if (uncaughtThrowableStrategy != null) {
+        diskCacheExecutor = GlideExecutor.newDiskCacheExecutor(uncaughtThrowableStrategy);
+      }
```
sjudd: It looks like you're overriding this just below with newDIskCacheExecutor() always? same for newSourceExecutor above.

library/src/main/java/com/bumptech/glide/GlideBuilder.java
```diff
@@ -245,6 +247,18 @@ public GlideBuilder setConnectivityMonitorFactory(ConnectivityMonitorFactory fac
...
 
   /**
+   * Sets the {@link
+   * UncaughtThrowableStrategy} to use to
+   *                                  handle uncaught exceptions.
+   *
+   * @param uncaughtThrowableStrategy The strategy to use.
+   * @return This builder.
+   */
+  public void setUncaughtThrowableStrategy(UncaughtThrowableStrategy uncaughtThrowableStrategy) {
```
sjudd: Since you can already set the disk cache and source executors and I expect this to be relatively rare, you can probably remove these methods on GlideBuilder and let users pass in their own GlideExecutors instead if they want to set the UncaughtThroawableStrategy.

stephanenicolas:
@sjudd : all comments have been addressed. I am not 100% sure I got right what you said about the `setUncaughtThrowableStrategy` in the builder. I removed the method (I was confused as you talked about `these methods`, plural) and preserve the API, and kept the most powerful part where you can set the executors and not only the strategy. 

If you could review this PR again, it would be very nice. Thanks for being so responsive.
And finally, if you accept the PR, we would appreciate an ETA for the next release so we can plan to move to it on our side.

Thanks for glide, it's really very nice a lib !

stephanenicolas:
Just in case people look for a usage example, here is what we do in our app to use the new strategy API:

```java
final UncaughtThrowableStrategy crashReportingUncaughtThrowableStrategy = new ...
builder.setDiskCacheExecutor(newDiskCacheExecutor(crashReportingUncaughtThrowableStrategy));
builder.setResizeExecutor(newSourceExecutor(crashReportingUncaughtThrowableStrategy));
```

sjudd:
LGTM, thank you!

If you want to you can add a section on how to use this API to the docs page: http://bumptech.github.io/glide/doc/configuration.html#application-options. There's some documentation on how to do this here: http://bumptech.github.io/glide/dev/contributing.html#documentation. Don't feel obligated though!

Issue #2419, opened 2017-09-21
downloadOnly + onlyRetrieveFromCache = is in cache?


proft:
Hi!

I have simple application and use Glide 4.1.1. It hast two activity. In first activity I load a image via

```
GlideApp.with(activity.getApplicationContext()).downloadOnly().load(url);
```

In logs I see http request ``HostConnection::get() New Host Connection established 0x7f7e27c19380, tid 14993``.

I want to use the loaded image in second activity in ``InfoWindow``. As you know, ``InfoWindow`` wants all content available at time of rendring. So, I use following snippet

```
public View getInfoWindow(Marker marker) {
    ...
    GlideApp.with(activity.getApplicationContext())
        .load(url) // the same url 
        .apply(new RequestOptions().onlyRetrieveFromCache(true))
        .into(ivPhoto);
    ...
}
```

I see new http request ``HostConnection::get() New Host Connection established 0x7f7e27c19480, tid 14983`` and no image in ``ImageView``. So, Glide doesn't use cache and made a new request for image.

Is it possible to use downloaded image?

Thanks!

sjudd:
Start with: http://bumptech.github.io/glide/doc/debugging.html#unexpected-cache-misses. Probably the sizes you're using aren't the same in both cases (and maybe the transformation, depending on the scale type in your ImageView).

proft:
Thanks for response! But no result :(.

1. In first request I just use ``downloadOnly()`` in second just ``into()``
2. Checked, no ``scaleType`` for ``ImageView``.

I added ``RequestOptions``, but no result

```
        final RequestOptions glideOptions = new RequestOptions();
        glideOptions.placeholder(R.drawable.logo);
        glideOptions.dontAnimate();
        glideOptions.diskCacheStrategy(DiskCacheStrategy.ALL);
        glideOptions.signature(new IntegerVersionSignature(2017));
```

Logs:

```
[...]

PR #2424, opened 2017-09-23
Make non-started loads are visible via Android lint

TWiStErRob:
## Description

Add `@CheckResult` to request builder's non-terminal methods.

## Motivation and Context

Prevent issues like https://github.com/bumptech/glide/issues/2419#issuecomment-331282358 by giving a lint error if Glide is used incorrectly. The builder must be terminated in order to start the load, so any intermediate builder methods' result must be used.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/RequestBuilder.java
+++ b/library/src/main/java/com/bumptech/glide/RequestBuilder.java
@@ -85,6 +85,7 @@ public class RequestBuilder<TranscodeType> implements Cloneable {
    * @see RequestOptions#apply(RequestOptions)
    * @return This request builder.
    */
+  @CheckResult
   public RequestBuilder<TranscodeType> apply(@NonNull RequestOptions requestOptions) {
     Preconditions.checkNotNull(requestOptions);
     this.requestOptions = getMutableOptions().apply(requestOptions);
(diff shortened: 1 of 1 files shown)


TWiStErRob:
Do we want to add the annotation on terminal methods that return `FutureTarget`? As far as I know `submit()`/`submit(int,int)`/`into(int,int)` needs to be cancelled (as per class JavaDoc), this would only ensure that at least `.get()` was called, even if not `cancel`.

sjudd:
You'll want to remove the check result special casing here too: https://github.com/bumptech/glide/blob/7457b9d82d2596c8f6be38e66cc7978d06568627/annotation/compiler/src/main/java/com/bumptech/glide/annotation/compiler/RequestBuilderGenerator.java#L312.

I'm ok with adding ``@CheckResult`` to submit if we make ``preload()`` handle being called on a background thread for users who just want to trigger a load on a background thread but don't want to do anything with a future. ``cancel()`` isn't really a requirement though, it can be an optimization if you're loading a lot of the same image type.

TWiStErRob:
I think this is good enough without `submit()`, especially if you say `cancel()` is not required.

Re APT: not sure I understand what that check does, didn't have time to good deep into that new code.

sjudd:
Merged here: 739cb35b637e8d4dd277d1133dbaeb2e2dc7d4ad. Thanks!

PR #2427, opened 2017-09-25
Update README.md

dalinaum:
## Description

`GlideApp` -> `Glide`.
Use `apply` method instead of `centerCrop`.

## Motivation and Context

According to `Glide v4 doucment`, I updated `README.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ Simple use cases with Glide's [generated API][21] will look something like this:
   ...
   ImageView imageView = (ImageView) findViewById(R.id.my_image_view);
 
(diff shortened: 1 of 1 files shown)


TWiStErRob:
Duplicate of #2047

dalinaum:
Thanks @TWiStErRob 

Is `centerCrop` generated API too? Otherwise, I think I would make a PR for `apply(centerCropTransform())`.

TWiStErRob:
@dalinaum yes it is, every non-static method from http://bumptech.github.io/glide/javadocs/430/com/bumptech/glide/request/RequestOptions.html are copied over to GlideApp's builder.

PR #2437, opened 2017-09-28
Fix typos in documentation.

Muyangmin:
## Description

Hello @sjudd , today when I was comparing and proofreading new docs I found a few typos, and fixed it now. 

By the way tell you a good news that I've write a simple script to compare differences in my Chinese doc and your English version, thus I'll get an issue when doc changes. This should make it easier to get our docs get synchronized.


sjudd:
Thanks again. Reasons why I should always run spell check++ :/.

Glad to hear about the script, neat idea.

PR #2441, opened 2017-09-30
rm test html in _posts.

Muyangmin:
I've deleted a html in _post dir for the reason described below.

## Description

I'm maintaining the Glide documentation of Chinese version, and as mentioned in #2437 , I use a robot to compare my repo's `mirror` path and `glide/_posts` path，to automatically pick up file changes in order to get the docs to be more consistent. So I must keep files in `mirror` to be exactly same as `glide/_post`.

But currently there is a `download-setup.html` in posts, which already caused Muyangmin/glide-docs-cn#10 . I've copied it to my repo to temporarily solve that issue, but I think this is just a historical test file, and should be deleted.

Please fell free to point out any problems, thanks!


sjudd:
Wups, thanks for fixing this!

PR #2447, opened 2017-10-03
Update proguard rules

ytRino:
## Description
It seems that was wrong class name and warning says
```
Note: the configuration refers to the unknown class 'com.bumptech.glide.AppGlideModule'
      Maybe you meant the fully qualified name 'com.bumptech.glide.module.AppGlideModule'?
```


sjudd:
Thanks!

Issue #2304, opened 2017-08-30
Dependencies can NOT be found in Jcenter


googolmo:
I can't download glide from jcenter, and there is no glide version 4 in jcenter.
Follow this link https://jcenter.bintray.com/com/github/bumptech/glide/glide/

sjudd:
I've never actually uploaded anything to jcenter. I've only ever uploaded to maven central. I was assuming jcenter was copying new versions over from maven central, but either that's stopped or I was incorrect.

It's kind of a pain to change the way uploads work. Is there some reason why you don't want to download from maven central?

googolmo:
@sjudd Sometimes maven central's server is very slow for my network.
In my option you can upload it to bintray and it can be sync to maven central.

sjudd:
Duplicate of #544

Issue #544, opened 2015-07-16
Missing files in jcenter


TWiStErRob:
While researching #532 I found that some files are missing from the `jcenter()` repo.
Take a look at the version folders at https://jcenter.bintray.com/com/github/bumptech/glide/okhttp-integration/ and you'll notice that many of them is missing a file or two.

Some are missing .aar some just .asc files. The most complete release seem to be 1.2.1, everything else is missing some .asc files, or the whole `*.aar*` bunch. Weirdly

``` gradle
compile 'com.github.bumptech.glide:okhttp-integration:1.3.1@aar'
```

works and merges the manifest correctly... which means that [okhttp-integration-1.3.1.aar](https://jcenter.bintray.com/com/github/bumptech/glide/okhttp-integration/1.3.1/okhttp-integration-1.3.1.aar) is there. So it may be just a jcenter display issue.

TWiStErRob:
I had a deja vu and look into it a little. I found a previous support ticket for another lib which was missing a whole version. So I contacted the bintray support again: *29907 Invisible files*

TWiStErRob:
> After investigating this issue and checking the logs for upload requests for the file ‘okhttp-integration-1.3.1.aar’. We think the file ‘okhttp-integration-1.3.1.aar’ might not have been uploaded to Bintray which could be the reason why we are not seeing it in the Bintray and Jcenter. The file is being downloaded from a remote repository when giving the filename in the path. Could you please try uploading this file to the repository in Bintray and see if it is showing up in the Bintray and JCenter.
If this is not the case and the file was uploaded to Bintray at some point of time, could you please let us know the approximate date and time when it was uploaded and we will investigate the file.

So it's likely pulling from [Maven Central](http://search.maven.org/#artifactdetails%7Ccom.github.bumptech.glide%7Cokhttp-integration%7C1.3.1%7Cjar) transparently.

sjudd:
I'm currently doing all uploads via Maven Central and I believe they just mirror everything they find on Maven. I'd consider switching to just jcenter, but I don't want to do both...

TWiStErRob:
Aah, so it sounds like you never uploaded anything there. They must be streaming files from Maven Central transparently, and if a file gets too many hits they physically mirror it; that's why only some files appear to be uploaded, but in fact all are available.

TWiStErRob:
They confirmed that something like this happens, but from repo.jfrog.org not Maven Central. Feel free to close this if you want to keep the current deployment process.

[...]

PR #2426, opened 2017-09-23
Build improvements

TWiStErRob:
(squash-merge this)

## Description

Update some build-related version numbers, and multiple minor code cleanups. Separated each improvement into commits, so we can drop some if you don't like it.

## Motivation and Context

Try to reduce the amount of noise at build-time, by fixing a few warnings. Also some cleanup in build.gradle files where IDEA was giving warnings and make it easier to edit code in IDEA.

PR #2448, opened 2017-10-04
Add side by side script to upload to bintray.

sjudd:
This allows us to test uploading to bintray without eliminating our
ability to upload to sonatype and maven central directly. If uploading
to bintray seems to work for a version or two and we figure out syncing
to maven central and snapshots, we can later remove the
maven_central.gradle script entirely.

Progress towards #2304 / #544.

At the moment the big piece that seems to be missing is some way to upload snapshots from travis. It sort of seems like jfrog/bintray have something, I just haven't figured out how to do it yet.


library/build.gradle
```diff
@@ -143,14 +143,3 @@ afterEvaluate {
...
 apply from: "${rootProject.projectDir}/scripts/upload.gradle"
-// exclude <dependency> tag for android support-v4 library from :glide's pom
-// this will ensure that this warning will not prevent the build from completing:
-// Module 'com.github.bumptech.glide:glide:4.0.0-SNAPSHOT' depends on one or more Android Libraries
-// but is a jar. Most users will need to override support-v4 version anyway if a newer version is
-// available
-// TODO make support-v4 a <scope>runtime</scope> dependency in pom.xml
-afterEvaluate {
-    uploadArchives.repositories.mavenDeployer.pom.whenConfigured { p ->
-        p.dependencies = p.dependencies.findAll { dep -> dep.artifactId != "support-v4" }
```
TWiStErRob: Where did this go?

scripts/artifacts.gradle
```diff
@@ -0,0 +1,85 @@
...
+    if (isAndroidProject) {
+        def releaseVariants = project.android.libraryVariants.findAll {
+            it.buildType.name.equalsIgnoreCase('release')
+        }
+
+        def getAndroidCompileSdkVersion = project.android.compileSdkVersion
+
+        def getAndroidSdkDirectory = project.android.sdkDirectory
+
+        def getAndroidJar = "${getAndroidSdkDirectory}/platforms/${getAndroidCompileSdkVersion}/android.jar"
```
TWiStErRob: note: `glide/build.gradle` may duplicate some this code

scripts/upload.gradle
```diff
@@ -1,198 +1,7 @@
-/*
```
TWiStErRob: hmm, this split may conflict with #2426, I'll resolve the changes requested there this week!

library/build.gradle
sjudd: Just deleted, we don't depend on v4 and we already use provided for our support library dependencies.

scripts/upload.gradle
sjudd: Resolved I think.

TWiStErRob: Looks good.

PR #2449, opened 2017-10-04
Add ExternalCache that fallsback to InternalCache if not available.

ANPez:
I've added an ExternalCacheFactory that tries to use external cache, and falls back to internal if not available.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/load/engine/cache/ExternalCacheDiskCacheFactory.java
+++ b/library/src/main/java/com/bumptech/glide/load/engine/cache/ExternalCacheDiskCacheFactory.java
@@ -9,17 +9,27 @@ import java.io.File;
  *
  * <p><b>Images can be read by everyone when using external disk cache.</b>
  */
+@Deprecated
 public final class ExternalCacheDiskCacheFactory extends DiskLruCacheFactory {
 
+  /**
+   * @deprecated Use {@link #ExternalPreferredCacheDiskCacheFactory(Context)}.
+   */
   public ExternalCacheDiskCacheFactory(Context context) {
     this(context, DiskCache.Factory.DEFAULT_DISK_CACHE_DIR,
         DiskCache.Factory.DEFAULT_DISK_CACHE_SIZE);
   }
 
+  /**
+   * @deprecated Use {@link #ExternalPreferredCacheDiskCacheFactory(Context, int)}.
+   */
   public ExternalCacheDiskCacheFactory(Context context, int diskCacheSize) {
     this(context, DiskCache.Factory.DEFAULT_DISK_CACHE_DIR, diskCacheSize);
   }
 
+  /**
+   * @deprecated Use {@link #ExternalPreferredCacheDiskCacheFactory(Context, int, String)}.
+   */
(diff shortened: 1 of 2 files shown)


ANPez:
Signed CLA

sjudd:
Do you have any context for why this is useful? Are there some devices where getExternalCacheDir always returns null? Are there any cases where the external cache directory is transiently null and then becomes valid later on?

ANPez:
For example, I have a Galaxy Nexus for testing purposes, where getExternalCacheDir returns null is there is no sd card present on the device.
It might aswell happen that SD card is temporarily unavailable, because of a temporary error, or by user interaction.

sjudd:
Would this make sense as a replacement for the existing external disk cache factory?

My only concern with transient failures is that you might end up swapping back and forth between two different cache directories each time the app is opened which would cause a  lot of additional cache misses and some unexpected storage usage. 

Maybe we should always try the external one and if it ever goes away, permanently fall back to the internal directory? That way we'd at least avoid the churn problem? Or do you think transient failures are sufficiently unlikely that we shouldn't worry about them?

ANPez:
In my experience, in old Android 2.x days, those kind of transient errors where kinda usual. Nowadays, fortunately, things have changed as many phones come with enough space as users don't need cards.

But yes, that is a worry that you could end up with two copies of everything. In that case, I'd proceed as you suggested and, I'd got any error, always fall back to the internal one. I don't think it's worth it implementing some kind of a "merged cache provider" for internal and external files.

Finally, I've been using Glide for quite a long time and, when used external storage, I've always followed this pattern of trying external, and falling back to internal storage on error. Maybe there is a use case where you want to force external, but I can't think of any.

sjudd:
Ok, maybe deprecate the old external disk cache factory and add a comment suggesting using this one instead? https://github.com/bumptech/glide/blob/9018ae6ae935a28a86452bb93c49e8bdcebb3ff5/library/src/main/java/com/bumptech/glide/load/engine/cache/ExternalCacheDiskCacheFactory.java#L12. 
That way in a future version we can change the behavior while allowing some time to comment. 

The only thing I can think of that would let you detect a previous failure would be to read/write a value to shared preferences that you'd check each time the disk cache was opened?

TWiStErRob:
Might be a silly idea, but the existence of the internal cache folder could be the trigger for using the permanent fallback. If it once fell back to internal cache, then that folder was created, so subsequent times it's enough to check the existence of that, i.e.:
```java
File internal = ...;
if (internal.exists()) return internal; // fallback was triggered in the past
FIle external = ...;
if (available(external)) return external; // preferred behavior
return internal; // fallback
```

sjudd:
Not silly at all! that's actually much better than shared prefs, good thought

ANPez:
Good idea! I'll work on it and update this pull request :+1:

sjudd:
Merged here: 56a427573bae591696d6986da4209757c36db9a3. Thanks!

TWiStErRob:
@sjudd Why are you commenting where you merged? Do you know about the "squash and merge" feature of GitHub?

sjudd:
@TWiStErRob I made a few minor code style tweaks here and I added support for the generated API to your PR, so the commits aren't exactly the same.

Issue #1464, opened 2016-09-12
Error:Failed to resolve: glide-parent.third_party:gif decoder:unspecified


jakingting:
**Glide Version**: 4.0.0-SNAPSHOT

**Integration libraries**: doesn't matter

**Device/Android Version**: doesn't matter, build error

**Issue details / Repro steps / Use case background**: 
Since today our gradle build started giving errors.

**Glide load line / `GlideModule` (if any) / list Adapter code (if any)**:
```gradle
compile 'com.github.bumptech.glide:glide:4.0.0-SNAPSHOT'
```

**Stack trace / LogCat**:
```
Error:A problem occurred configuring project ':app'.
> Could not resolve all dependencies for configuration ':app:_P003DebugCompile'.
   > Could not find glide-parent.third_party:gif_decoder:unspecified.
     Searched in the following locations:
         file:/D:/jakingting/android/Android Studio/gradle/m2repository/glide-parent/third_party/gif_decoder/unspecified/gif_decoder-unspecified.pom
         file:/D:/jakingting/android/Android Studio/gradle/m2repository/glide-parent/third_party/gif_decoder/unspecified/gif_decoder-unspecified.jar
         https://jcenter.bintray.com/glide-parent/third_party/gif_decoder/unspecified/gif_decoder-unspecified.pom
         https://jcenter.bintray.com/glide-parent/third_party/gif_decoder/unspecified/gif_decoder-unspecified.jar
         https://jitpack.io/glide-parent/third_party/gif_decoder/unspecified/gif_decoder-unspecified.pom
         https://jitpack.io/glide-parent/third_party/gif_decoder/unspecified/gif_decoder-unspecified.jar
         http://oss.sonatype.org/content/repositories/snapshots/glide-parent/third_party/gif_decoder/unspecified/gif_decoder-unspecified.pom
         http://oss.sonatype.org/content/repositories/snapshots/glide-parent/third_party/gif_decoder/unspecified/gif_decoder-unspecified.jar
         file:/D:/jakingting/android/sdk/extras/android/m2repository/glide-parent/third_party/gif_decoder/unspecified/gif_decoder-unspecified.pom
         file:/D:/jakingting/android/sdk/extras/android/m2repository/glide-parent/third_party/gif_decoder/unspecified/gif_decoder-unspecified.jar
         file:/D:/jakingting/android/sdk/extras/google/m2repository/glide-parent/third_party/gif_decoder/unspecified/gif_decoder-unspecified.pom
[...]

PR #2508, opened 2017-10-24
Fix crash because of wrong executor being created

stephanenicolas:
We are having issues using the new API to set an UncaughtThrowableStrategy. There was a mistake in my latest PR.

Sorry about it. Can you give us an ETA to release the fix on central? We might have to reverse the change on our side if it's not published soon.

The crash is pretty simple: no images load (on Oreo devices ??? strange, it should always crash..). Anyway, yes, no images are loaded because we use an executor that can't go to the network.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/load/engine/executor/GlideExecutor.java
+++ b/library/src/main/java/com/bumptech/glide/load/engine/executor/GlideExecutor.java
@@ -140,7 +140,7 @@ public final class GlideExecutor extends ThreadPoolExecutor {
    */
   public static GlideExecutor newSourceExecutor(
           UncaughtThrowableStrategy uncaughtThrowableStrategy) {
-    return newDiskCacheExecutor(DEFAULT_DISK_CACHE_EXECUTOR_THREADS,
+    return newSourceExecutor(DEFAULT_DISK_CACHE_EXECUTOR_THREADS,
                                 DEFAULT_DISK_CACHE_EXECUTOR_NAME, uncaughtThrowableStrategy);
   }
 


sjudd:
Ugh sorry I should have caught that too. Thanks for noticing and sending this.

I'm curious about why it only doesn't work on O+ though. I believe I added (and tested) the network restriction before O went out, but I could be wrong.

This would go out in our next monthly release, 4.3, which I'm aiming to get out by the end of October. You can follow that milestone for changes to the date.

LGTM, just waiting to confirm the build passes.

sjudd:
Firebase failure is a configuration issue. Thanks!

stephanenicolas:
Thx for merging it. For us, end of October is quite long, we probably gonna
reverse it as we can't afford the app to bug on O+ for 2 weeks.

In case you could do a release only for this bug fix, it would be
appreciated for sure. But I don't wanna break your monthly release effort,
it's already amazing to keep up this pace on the long run.

Thx for glide and sorry again for adding this bug. Damn it ! ;)

S.

stephanenicolas:
@sjudd finally it's all good, we could find a workaround until you release 4.3. Thx !

sjudd:
@stephanenicolas Great glad you have a solution. Thanks for contributing and following up when you found the bug. It's incredibly helpful!

In case you hadn't considered it, you can always try a snapshot (or just downloading the jar from Sonatype to avoid it changing out from under you): http://bumptech.github.io/glide/dev/snapshots.html

TWiStErRob:
@sjudd You can specifically reference a snapshot version, without it being dynamic, see for example https://github.com/bumptech/glide/issues/1464#issuecomment-246280003

sjudd:
... I had no idea. Or I'd forgotten that I'd once had an idea :). I'll see if I can update our documentation.

sjudd:
Added here: b93defadcaf680fd39825b23f9f31d9b266c08b9

Issue #2271, opened 2017-08-19
GIF overdraw


arkon:
**Glide Version**: 4.0.0

**Integration libraries**: N/A

**Device/Android Version**: Nexus 7 2013 (7.1.2), OnePlus One (7.1.2)

**Issue details / Repro steps / Use case background**: 

Loading a local GIF (a copy of [this GIF](https://listen.moe/files/images/kanna.gif)) displays overdraw artifacts.

Within the activity ([link to code](https://github.com/arkon/listen-moe-android/blob/a389e9db8df3020f30f221b38cf166f6acfffa96/app/src/main/java/me/echeung/moemoekyun/ui/activities/SearchActivity.java#L51)):
```java
Glide.with(this)
        .load(R.drawable.kanna_dancing)
        .into(binding.searchPlaceholder);
```

**Layout XML** ([link to code](https://github.com/arkon/listen-moe-android/blob/1b8756aa00b5bbe2d9bbfcee5048afe10e267329/app/src/main/res/layout/activity_search.xml#L67)):
```xml
<ImageView
    android:id="@+id/search_placeholder"
    android:layout_width="156dp"
    android:layout_height="200dp"
    android:layout_gravity="center"
    android:layout_marginBottom="16dp" />
```

**Stack trace / LogCat**:
```ruby
E/GlideExecutor: Request threw uncaught throwable
    java.lang.NullPointerException: Attempt to invoke interface method 'void com.bumptech.glide.load.data.DataFetcher.cleanup()' on a null object reference
        at com.bumptech.glide.load.engine.DecodeJob.run(DecodeJob.java:238)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1133)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:607)
        at java.lang.Thread.run(Thread.java:761)
        at com.bumptech.glide.lo    GlideExecutor.java:347)
```

**GIF of what it looks like**:
<details>
<summary>Expand to see GIF</summary>

![baconnjh47deugene08192017111452](https://user-images.githubusercontent.com/4098258/29487952-a9e123aa-84cf-11e7-829f-85d57c89c7c9.gif)

</details>

sjudd:
[...]

PR #2521, opened 2017-10-26
clear all pixels when dispose is 3 but previousImage is null

michaeimm:
## Description
Fix Issues #2271 and part of #1094

## Motivation and Context
Dispose in frame should not be DISPOSAL_PREVIOUS when there did not save previous image. GIF like #2271 is wrong because the dispose is DISPOSAL_PREVIOUS in first frame.

### DIFF
--- a/third_party/gif_decoder/src/main/java/com/bumptech/glide/gifdecoder/StandardGifDecoder.java
+++ b/third_party/gif_decoder/src/main/java/com/bumptech/glide/gifdecoder/StandardGifDecoder.java
@@ -418,9 +418,17 @@ public class StandardGifDecoder implements GifDecoder {
     // Final location of blended pixels.
     final int[] dest = mainScratch;
 
-    // clear all pixels when meet first frame or dispose is 3 but previousImage is null
-    if (previousFrame == null || (previousFrame.dispose == DISPOSAL_PREVIOUS && previousImage == null)) {
+    // clear all pixels when meet first frame and drop prev image from last loop
+    if (previousFrame == null) {
       previousImage = null;
+      Arrays.fill(dest, COLOR_TRANSPARENT_BLACK);
+    }
+
+    // clear all pixels when dispose is 3 but previousImage is null.
+    // When DISPOSAL_PREVIOUS and previousImage didn't be set, new frame should draw on
+    // a empty image
+    if (previousFrame != null && previousFrame.dispose == DISPOSAL_PREVIOUS
+            && previousImage == null) {
       Arrays.fill(dest, COLOR_TRANSPARENT_BLACK);
     }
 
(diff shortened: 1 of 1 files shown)


michaeimm:
I signed it!

TWiStErRob:
Please run `gradlew check` locally, there seems to be some violation; probably a long line.

I also suggest you run this example with the new code to see if there's a regression:
https://github.com/TWiStErRob/glide-support/blob/master/src/glide3/java/com/bumptech/glide/supportapp/github/_1094_gif_disposal/TestFragment.java

third_party/gif_decoder/src/main/java/com/bumptech/glide/gifdecoder/StandardGifDecoder.java
```diff
@@ -418,8 +418,9 @@ private Bitmap setPixels(GifFrame currentFrame, GifFrame previousFrame) {
     // Final location of blended pixels.
     final int[] dest = mainScratch;
 
-    // clear all pixels when meet first frame
-    if (previousFrame == null) {
+    // clear all pixels when meet first frame or dispose is 3 but previousImage is null
```
TWiStErRob: "dispose is 3 but previousImage is null" part doesn't explain much, it's exactly the same as the code below it, you should probably explain what it means, or why we need to drop the prev image/fill to black.

michaeimm: OK. I think conditions should be separate. We need to drop the prev image when meet first frame like we do clear all pixels. If prev image exist in second frame, prev image which saved in prev loop will be fill back in new loop and cause overdraw.

sjudd:
:third_party:gif_decoder:checkstyle[ant:checkstyle] /home/travis/build/bumptech/glide/third_party/gif_decoder/src/main/java/com/bumptech/glide/gifdecoder/StandardGifDecoder.java:422: error: Line is longer than 100 characters (found 105).

https://travis-ci.org/bumptech/glide/jobs/293267074

michaeimm:
Thank you for your reply.
I have modified something.
Please review the latest commit.

sjudd:
The two travis failures are setup related and can be ignored.

third_party/gif_decoder/src/main/java/com/bumptech/glide/gifdecoder/StandardGifDecoder.java
```diff
@@ -418,8 +418,17 @@ private Bitmap setPixels(GifFrame currentFrame, GifFrame previousFrame) {
     // Final location of blended pixels.
     final int[] dest = mainScratch;
 
-    // clear all pixels when meet first frame
+    // clear all pixels when meet first frame and drop prev image from last loop
     if (previousFrame == null) {
+      previousImage = null;
```
sjudd: Here you probably should be releasing previousImage so the Bitmap can be re-used, see the four lines in ``clear()``: https://github.com/bumptech/glide/blob/d447e582397cc7cf3f95a9871eb18d81a917544f/third_party/gif_decoder/src/main/java/com/bumptech/glide/gifdecoder/StandardGifDecoder.java#L327

sjudd:
LGTM. Thank you for putting this up and your patience. I really appreciate the help!

Issue #2022, opened 2017-06-10
What is the maximum cache size that I can set?? Can I set more than INTEGER_MAX value?


bhargavms:
HI,
I would like to know if it would be possible to set a cache size more than the that allowed by the `int` data type? since `builder.setDiskCache(new InternalCacheDiskCacheFactory(context, diskCacheSizeBytes));` takes `int` as the discCacheSizeBytes.

sjudd:
Not currently, you're welcome to send a pull request to change these to longs.

2GB is a lot of cache space...

bhargavms:
Alright Ill be sure to get on it first thing once I finish the project at hand :)

sjudd:
Thanks! I'll look forward to it.

sjudd:
Fixed in #2558

PR #2537, opened 2017-10-31
Use long value so that cache sizes greater than 2Gigabytes can be supported

bhargavms:
## Description
I want to be able to set cache sizes greater than 2 gigabytes which is the max value supported by the int primitive type, my approact to achieving this is to change the type used for maintaing cache size to long, so that cache with greater than 2 gigabyte can be used since these days phones are getting bulkier I think its fair to also let developers use more of the phone to improve User experience.

Issue Reference:
https://github.com/bumptech/glide/issues/2022

## Motivation and Context
Bigger cache sizes equals better user experiences!

### DIFF
--- a/third_party/gif_decoder/src/main/java/com/bumptech/glide/gifdecoder/StandardGifDecoder.java
+++ b/third_party/gif_decoder/src/main/java/com/bumptech/glide/gifdecoder/StandardGifDecoder.java
@@ -418,8 +418,20 @@ public class StandardGifDecoder implements GifDecoder {
     // Final location of blended pixels.
     final int[] dest = mainScratch;
 
-    // clear all pixels when meet first frame
+    // clear all pixels when meet first frame and drop prev image from last loop
     if (previousFrame == null) {
+      if (previousImage != null) {
+        bitmapProvider.release(previousImage);
+      }
+      previousImage = null;
+      Arrays.fill(dest, COLOR_TRANSPARENT_BLACK);
+    }
+
+    // clear all pixels when dispose is 3 but previousImage is null.
+    // When DISPOSAL_PREVIOUS and previousImage didn't be set, new frame should draw on
+    // a empty image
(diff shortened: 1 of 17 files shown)


bhargavms:
CLA signed

sjudd:
The changes you've made are only to the classes that hold data in memory. You want memory caches > 2GB?

bhargavms:
@sjudd oh yes, forgot to change diskcache too, will update PR, yes user should have option to set whatever cache size he wants is my opinion.

bhargavms:
CLA signed

bhargavms:
@sjudd if you don't like using `long` for memory cache size, I can revert commit and squash, I'm not terribly passionate about that part.

third_party/gif_decoder/src/main/java/com/bumptech/glide/gifdecoder/StandardGifDecoder.java
```diff
@@ -418,8 +418,20 @@ private Bitmap setPixels(GifFrame currentFrame, GifFrame previousFrame) {
     // Final location of blended pixels.
     final int[] dest = mainScratch;
 
-    // clear all pixels when meet first frame
+    // clear all pixels when meet first frame and drop prev image from last loop
```
sjudd: I think this means you need to rebase.

sjudd:
I don't object to using longs for memory cache/bitmap pool. It's a little overkill now, but not problematic. LGTM. 

It may be another day or two before I merge this, I think there's going to be a dot release of 4.3 which I'd like to get out first.

third_party/gif_decoder/src/main/java/com/bumptech/glide/gifdecoder/StandardGifDecoder.java
bhargavms: @sjudd I will close this PR and open another one by rebasing correctly

Issue #2549, opened 2017-11-02
Issue cleanup and maintenance


sjudd:
We have 464 (about to be 465) open issues. Almost all of these are stale. Most of them are questions where either the poster didn't close the issue after receiving a response or the poster never answered some set of questions or suggestions.

For newer questions we're reluctant to close issues when providing responses because closing issues seems to discourage people from posting responses. Only contributors seem to have the power to re-open issues as well, which might be part of the problem. Unfortunately our reluctance to close issues is just adding to the number of stale issues over time.

For now I've been mostly able to keep track of important bugs and feature requests with a combination of Milestones and Labels. However I do occasionally miss a response to an issue where there wasn't initially enough information to triage it.

I believe we can make this better by:
1. Closing issues that do not fill out the issue template with a template suggestion to file a new issue or add a comment with the template filled out.
2. Adding a bot (https://github.com/probot/stale) to automatically close stale issues after N days (probably 7) if they're not labeled as a bug or an enhancement. 

We also need to deal with the backlog at some point. The protobot/stale seems to actually run through old issues for you, though only 30 an hour. That's mostly fine, but we're going to spam watchers (and contributors) with 400+ notifications while the bot makes its way through the backlog.

I'm also open to other ideas.

sjudd:
@TWiStErRob this is mostly for you, let me know if you have thoughts. I'm leaning pretty strongly towards at least enabling a bot to close stale issues.

PR #2556, opened 2017-11-03
Add probot stale configuration.

sjudd:
Progress towards #2549. This will comment on and automatically close stale issues that don't have a whitelisted set of tags.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,21 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 7
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - bug
+  - enhancement
+  - feature
+  - good first issue
+  - help wanted
+# Label to use when marking an issue as stale
+staleLabel: stale
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  activity in the last seven days. It will be closed if no further activity 
+  occurs within the next seven days. Thank you for your contributions.
+# Comment to post when closing a stale issue. Set to `false` to disable
+closeComment: false
+unmarkComment: false


TWiStErRob:
Wait a bit with the merge, I have some holidays coming up in the next weeks, I could do some manual triaging.

sjudd:
You will still be able to triage, this is mostly going to target unlabeled issues or question labeled issues that no ones touched in a while. We can also re-open issues if the bot closes them. 

Are you worried we have some older useful issues that aren't labeled?

TWiStErRob:
It's possible that some issues haven't been responded to, in the past year I haven't had much time, so there may be issues which are waiting for response from us. Those shouldn't be said to be stale.

Also unlabeled issues should be labeled as either questions or needs info and only then start the staling process. I can commit a day to do some triaging. Bot closing stuff and then we reopening is not ideal, because I assume that anything in a Closed has been processed by a human on the maintainer side at some point. If it starts closing issues I won't be able to have that distinction (i.e. don't know which issues need to be looked at).

Can you set a minimum issue number/date for the bot to start processing from? All I'm asking is to postpone this for a few days. (Sat, Sun, Mon), from Tue on I won't be available.

sjudd:
I can definitely wait until next week, there's no rush here. I'm trying to make sure I understand what your concerns are so that whatever solution we come up with addresses them.

My view is generally that if an issue is substantially old, regardless of who it's waiting on a response from, it's probably not worth keeping open. If the reporter hasn't followed up and no other users have commented or touched the issue, it's probably not hugely important. That said, I think most of the old issues are already labeled and won't be closed by this bot. 

Mostly I'm aiming at all of the questions we get. I'd like to avoid closing them immediately when I answer them, but I don't have the patience to go back through and manually close them after N days if there's no follow up after we do provide an answer.

We do also have a pretty massive backlog that I don't want to go through manually either. The bot will comment on those and then give another 7 days for anyone in the issue to react before closing the issue, so hopefully it won't even get to close any issue someone wants to keep open, it will just prompt them to say something.

PR #2558, opened 2017-11-03
Use long value so that cache sizes greater than 2Gigabytes can be supported.

bhargavms:
## Description
I want to be able to set cache sizes greater than 2 gigabytes which is the max value supported by the int primitive type, my approact to achieving this is to change the type used for maintaing cache size to long, so that cache with greater than 2 gigabyte can be used since these days phones are getting bulkier I think its fair to also let developers use more of the phone to improve User experience.

Issue Reference:
https://github.com/bumptech/glide/issues/2022

## Motivation and Context
Bigger cache sizes equals better user experiences!


bhargavms:
@sjudd resigned commits and rebased with Master

sjudd:
Thanks!

PR #2641, opened 2017-11-24
External storage exists but not writable

ANPez:
The Huawei P9 lite is returning the external storage path as if it were available, but it is not writable, therefore, writes fail. This commit fixes the issue.


sjudd:
LGTM, thanks!

PR #2666, opened 2017-12-01
Update dependencies and fix contributing issues

SUPERCILEX:
Prep work for an upcoming PR


.gitignore
```diff
@@ -23,24 +21,7 @@ docs/**/*
 # sed
 *.bak
 
-# Intellij 
-*.ipr
+# Intellij
 *.iml
-*.iws
```
SUPERCILEX: These are unnecessary and don't cover all the cases.

instrumentation/build.gradle
```diff
@@ -31,7 +31,6 @@ dependencies {
 
 android {
     compileSdkVersion COMPILE_SDK_VERSION as int
-    buildToolsVersion BUILD_TOOLS_VERSION as String
```
SUPERCILEX: Build tools declarations are no longer required and Tor said they should be removed.

SUPERCILEX:
@sjudd shouldn't there be a Travis build?

sjudd:
Yup there should be a travis build I'll look in to it.

I'm going to try to release 4.4 before accepting, but I'll try and look through everything sooner so that it's all set to go.

Thanks for sending these!

SUPERCILEX:
@sjudd Thanks for the info, SGTM!

PS: I have way more PRs to come, so feel free to ignore them until 4.4 is out. 😄

sjudd:
So travis ran: https://travis-ci.org/bumptech/glide/builds/309889803. It just didn't update the status here.

.travis.yml
```diff
@@ -2,16 +2,16 @@ language: android
 sudo: required
 dist: trusty
 
+before_install:
```
sjudd: What's the motivation for this change?

SUPERCILEX: Travis sucks. 😆 Nah, just kidding. Seriously though, Travis has been unreliable in my experience when it comes to installing dependencies properly, hence trying to minimize the use of their scripts. For a concrete reason though, their `licenses` block [failed on one of my builds](https://travis-ci.org/SUPERCILEX/glide/jobs/309892702#L1204). If you can find another reliable solution that's less ugly than this one, awesome! Otherwise, I'm not sure what else to do. 😄

sjudd: Huh, I haven't seen that failure before. Maybe it's worse with pull requests? Right now my travis flakiness is all emulator related.

SUPERCILEX: I've also seen it with FirebaseUI: https://github.com/firebase/FirebaseUI-Android/pull/991#discussion_r146145121. It seems to happen when the Gradle plugin tries to install dependencies: I think Travis uses an `echo "y"` whereas Google actually looks for the files.

SUPERCILEX:
@sjudd Also see this last minute fix: f26b53c. All my personal Travis builds were failing because I don't have access to your Firebase stuff, but I'm also not on a PR. That commit just makes sure Travis is on your repo.

.travis.yml
sjudd: Got it, I'm ok with it, thanks.

.travis.yml
```diff
@@ -2,16 +2,16 @@ language: android
...
+  - echo -e "\n8933bad161af4178b1185d1a37fbf41ea5269c55\nd56f5187479451eabf01fb78af6dfcb131a6481e" > "$ANDROID_HOME/licenses/android-sdk-license"
+  - echo -e "\n84831b9409646a918e30573bab4c9c91346d8abd\n504667f4c0de7af1a06de9f4b1727b84351f2910" > "$ANDROID_HOME/licenses/android-sdk-preview-license"
+
 android:
   components:
-  - tools
-  - platform-tools
-  - tools
-  - build-tools-26.0.2
-  - android-26
```
sjudd: Are these removed because they're the defaults?

SUPERCILEX: No, but the Android Gradle Plugin downloads these for you (hence the licensing stuff). You can see it start here: https://travis-ci.org/bumptech/glide/jobs/310330628#L1017.

SUPERCILEX:
@sjudd Alrighty, I've rebased that last commit away so we should be good for a merge whenever you're ready! 😄

sjudd:
Actually this should be safe for 4.4. Thanks!

PR #2671, opened 2017-12-03
Cleanup target api annotations

SUPERCILEX:
Any reason to use `TargetApi`? Dunno why the support lib folks haven't just deprecated that annotation, but `RequiresApi` is the better one. Also, this PR removes target annotations for api 14 which is the min of this library anyway.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/Glide.java
+++ b/library/src/main/java/com/bumptech/glide/Glide.java
@@ -1,6 +1,5 @@
 package com.bumptech.glide;
 
-import android.annotation.TargetApi;
 import android.app.Activity;
 import android.content.ComponentCallbacks2;
 import android.content.Context;
@@ -10,7 +9,6 @@ import android.graphics.Bitmap;
 import android.graphics.drawable.BitmapDrawable;
 import android.graphics.drawable.Drawable;
 import android.net.Uri;
-import android.os.Build;
 import android.os.ParcelFileDescriptor;
 import android.support.annotation.Nullable;
 import android.support.annotation.VisibleForTesting;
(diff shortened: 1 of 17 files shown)


sjudd:
I think most of these are actually not correct.

``@RequiresAPI`` means it's up to the caller to only call the method/class if they're on that API level or higher. 
``@TargetAPI`` means that the method/class is certifying that it's safe for it to be used on any API level, as long as the internal methods use APIs less than or equal to the specified target API.

In most of these cases the methods themselves are checking the SDK version, so we should be using ``@TargetApi``.

SUPERCILEX:
Ohhh, I see what you're saying. Then is there a reason to have them at all? Every method could be annotated going by that philosophy... I'm fine either way so if you want to keep them, I'll update this PR to just remove the ones using API 14.

sjudd:
``@TargetApi`` is basically ``@SuppressWarnings`` for API checks, except that it's limited to a maximum sdk version so that it doesn't silently suppress warnings for a newer API level if you modify the method later on to use a new sdk level that you're not checking against.

For example if we use a method that requires android O, check it properly with an if/else, suppress the warning with @SuppressWarnings, and then go back and add a method that requires O MR1, we won't see a lint or ide warning about the unsafe O MR1 call. TargetApi fixes that by warning for all methods past whatever level you give it.

So using @TargetApi makes sense whenever @SuppressWarnings would make sense, it's just a bit safer.

sjudd:
... Also ``@SuppressWarnings`` is a user on GitHub, wups...

SUPERCILEX:
@sjudd Ohhhhhhh, wow, that's super cool! Ok, I really get it now, yeah, that sounds awesome. Now I have to go through all my projects and remove my `@SuppressLint("NewApi")`! 😆

Anyway, I'm going to change this PR to just remove the outdated suppressions.

SUPERCILEX:
Ok, I think I've got everything done right. I removed some checks that just seemed unnecessary, removed the API 14 ones, and changed some of the `TargetApi`s to `RequiresApi`s where relevant.

PR #2672, opened 2017-12-03
Fix wrong document

tranleduy2000:


sjudd:
See http://bumptech.github.io/glide/doc/generatedapi.html#getting-started

PR #2681, opened 2017-12-05
Add nullability annotations to library module (1/n)

SUPERCILEX:
Man this is painful! 😄 Here's what I got done before I ran out of podcasts.

### DIFF
--- a/annotation/compiler/test/src/test/resources/EmptyAppGlideModuleTest/GlideApp.java
+++ b/annotation/compiler/test/src/test/resources/EmptyAppGlideModuleTest/GlideApp.java
@@ -4,6 +4,7 @@ import android.annotation.SuppressLint;
 import android.app.Activity;
 import android.app.Fragment;
 import android.content.Context;
+import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.support.annotation.VisibleForTesting;
 import android.support.v4.app.FragmentActivity;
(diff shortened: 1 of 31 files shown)


SUPERCILEX:
@sjudd I couldn't find docs on updating the generated API tests. Could you help me fix those? Thanks! 😄

sjudd:
I need to write a script for them. Right now the best way I've found to do it is to run the tests in Android Studio, then click on the failure comparison. That will open a pretty useful window that shows the complete files and highlights the diff. If the diff looks reasonable you can copy/paste the new version into the old file.

It's a pain :/. There's probably a better way of doing this...

sjudd:
Oh I should also note that the files that we check against are stored in resources, here: https://github.com/bumptech/glide/tree/dc12b60c8f0a21eb9b3ab2164471c93d2e20ae25/annotation/compiler/test/src/test/resources.

Either the tests will have an identically named package under resources with their test classes (if their output differs from the default) or they default to the files in either https://github.com/bumptech/glide/tree/dc12b60c8f0a21eb9b3ab2164471c93d2e20ae25/annotation/compiler/test/src/test/resources/EmptyAppGlideModuleTest or https://github.com/bumptech/glide/tree/dc12b60c8f0a21eb9b3ab2164471c93d2e20ae25/annotation/compiler/test/src/test/resources/EmptyLibraryGlideModuleTest.

The helper class defines methods that are used to determine the actual path:
https://github.com/bumptech/glide/blob/dc12b60c8f0a21eb9b3ab2164471c93d2e20ae25/annotation/compiler/test/src/test/java/com/bumptech/glide/annotation/compiler/test/Util.java#L15

SUPERCILEX:
@sjudd Ok! After trying to figure it out off and on, I think I finally got it. See #2703 😄

sjudd:
Awesome thanks, sorry for the delay!

SUPERCILEX:
No problemo, I haven't had time to work on the next batch anyway.

Issue #2637, opened 2017-11-24
[Feature request] Improve ViewPropertyTransition.Animator


Tolriq:
**Glide Version**:
4.3.1

When trying to understand some speed problems on low end devices, I found out that my custom animations where too slow sometimes and left the imageview in transient state too long. 
Meaning that recyclerview did not recycle the viewholder and created new viewholders everytime.

Fix is easy cancel animations on recycle :) But then trying to do that with current Glide API becomes complicated.

Workaround is to store a reference to viewproperty animator in a tag in the image then call the cancel in the adapter but this is quite ugly :)

I would suggest that ViewPropertyTransition.Animator is extended to have a cancel() or release() interface and have this interface called when clear is called on the target or view recycled

sjudd:
Thanks for the idea, I'd welcome a pull request if there's something that would make this easier.

Tolriq:
Well if you draw me a map about where to plug for the calls and wanted names I can try to find some times.

But usually I start to be lost at all possible cases when going so deep in Glide :(

sjudd:
I haven't looked at this in detail, so I might be missing something, but probably we should:

1. Modify [``Transition``][1] to add a cancellation method.
2. Update [``SingleRequest``][2] to retain a reference to the ``Transition`` and cancel it when a new load is started or when ``clear()`` or ``pause()`` are called.

We could also make calling the new ``cancel()`` method the responsibility of ``Target`` implementations, but that requires modifying a lot more classes and is more error prone because subclasses can always accidentally not call super. It's probably fine in that there's no good reason to keep a transition running after the corresponding request is cleared.

That combined with the recently added [``clearOnDetach``][3] method would probably work pretty well for ``RecyclerView``.

[1]: https://github.com/bumptech/glide/blob/5580e517d05d178946369e178eb16022224dd362/library/src/main/java/com/bumptech/glide/request/transition/Transition.java
[2]: https://github.com/bumptech/glide/blob/5580e517d05d178946369e178eb16022224dd362/library/src/main/java/com/bumptech/glide/request/SingleRequest.java#L575
[3]: http://bumptech.github.io/glide/javadocs/440/com/bumptech/glide/request/target/ViewTarget.html#clearOnDetach--

Tolriq:
@sjudd I tried to give it a quick shot before week end, but seems like handling it in SingleRequest is complicated. 

To do everything correctly for all cases, the cancel in Transition should pass the ViewAdapter but there's no access to it from there.

[...]

PR #2701, opened 2017-12-08
Expand glide transition API to allow cancellation.

Tolriq:
## Description

Expand glide transition API to allow cancellation.

## Motivation and Context

See #2637

### DIFF
--- a/library/src/main/java/com/bumptech/glide/request/transition/DrawableCrossFadeTransition.java
+++ b/library/src/main/java/com/bumptech/glide/request/transition/DrawableCrossFadeTransition.java
@@ -15,6 +15,7 @@ import android.graphics.drawable.TransitionDrawable;
 public class DrawableCrossFadeTransition implements Transition<Drawable> {
   private final int duration;
   private final boolean isCrossFadeEnabled;
+  private ViewAdapter viewAdapter;
 
   /**
    * @param duration The duration that the cross fade animation should run if there is something to
@@ -50,11 +51,23 @@ public class DrawableCrossFadeTransition implements Transition<Drawable> {
     if (previous == null) {
       previous = new ColorDrawable(Color.TRANSPARENT);
     }
+    viewAdapter = adapter;
     TransitionDrawable transitionDrawable =
         new TransitionDrawable(new Drawable[] {previous, current});
     transitionDrawable.setCrossFadeEnabled(isCrossFadeEnabled);
     transitionDrawable.startTransition(duration);
     adapter.setDrawable(transitionDrawable);
     return true;
+  }
+
+  @Override
+  public void cancel() {
+    if (viewAdapter != null) {
+      TransitionDrawable transitionDrawable = (TransitionDrawable) viewAdapter.getCurrentDrawable();
+      if (transitionDrawable != null) {
+        transitionDrawable.resetTransition();
+        viewAdapter.setDrawable(transitionDrawable.getDrawable(1));
+      }
+    }
   }
 }
--- a/library/src/main/java/com/bumptech/glide/request/transition/ViewPropertyTransition.java
+++ b/library/src/main/java/com/bumptech/glide/request/transition/ViewPropertyTransition.java
@@ -12,6 +12,7 @@ import android.view.View;
 public class ViewPropertyTransition<R> implements Transition<R> {
 
   private final Animator animator;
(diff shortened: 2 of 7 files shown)


Tolriq:
@sjudd first proposal for #2637.

Not fully tested for the moment but prefers to validate concepts early and eventually before Week End :)

And needs helps / discussion about DrawableCrossFadeTransition as I don't really know how they works and how to properly cancel them without generating issues.

library/src/main/java/com/bumptech/glide/request/transition/ViewPropertyTransition.java
```diff
@@ -12,6 +13,7 @@
 public class ViewPropertyTransition<R> implements Transition<R> {
 
   private final Animator animator;
+  private WeakReference<ViewAdapter> viewAdapter;
```
sjudd: I don't think there's a need for a weak reference here. If there is, can you add a comment explaining why it's necessary?

library/src/main/java/com/bumptech/glide/request/transition/DrawableCrossFadeTransition.java
```diff
@@ -57,4 +57,9 @@ public boolean transition(Drawable current, ViewAdapter adapter) {
     adapter.setDrawable(transitionDrawable);
     return true;
   }
+
+  @Override
+  public void cancel() {
+    // TODO
```
sjudd: https://developer.android.com/reference/android/graphics/drawable/TransitionDrawable.html#resetTransition()

Then swap the positions of the Drawables? I think?

Tolriq: Well I don't know what would be expected as a cancel don't really know on cancel what image should be shown. But if you say I should show the last I'll do that :)

library/src/main/java/com/bumptech/glide/request/transition/ViewPropertyTransition.java
Tolriq: As said in the issue discussion I'm not sure to fully understand all internals of Glide (Actually I'm sure I don't :p) so I chose the safe way to avoid hard references as I don't know if there's case where the transition can last too long.

sjudd: :) A hard reference here should be fine then, especially since you're adding an API to cancel and end the transition in clear anyway.

Tolriq:
Updated but probably won't have time to fully test before January now :(

sjudd:
Wups sorry I looked through this and it seems reasonable. Let me know when you get the chance to test it.

Tolriq:
@sjudd Will try to do the tests before end of year next week. Do you have a date for next release and do you think this could be merged for it?

Or no hurry and I can wait some more.

sjudd:
@Tolriq I'm probably aiming for the first week of January after the holidays for 4.5.0

Tolriq:
@sjudd 
Ok so tested and it works as expected and match great with clearOnDetach.

But this does not covers all the cases and I'm facing another issue that I'll describe in previous issue: #2637 
(Not tested the transitiondrawable part)

sjudd:
Ok I'll take a look at merging this, but after 4.5.0 goes out so we have some time for testing before it's released.

sjudd:
Looks reasonable to me. I still need to test the transition drawable state.

The onAnimationEnd issue sounds like it might be a bug? Maybe in the framework? It's documented to call listeners: https://developer.android.com/reference/android/view/animation/Animation.html#cancel().

Tolriq:
@sjudd Yes Cancel does call the listeners normally.

The issue seems more like Glide sometimes release resources before the call kick in. Maybe it's due to the fact that I set transition to null. But I'm not sure if it's better to be conservative and null things or risk a leak.

Seeing all different Android issues since 1.5 I tends to be very very very conservative :)

Tolriq:
@sjudd time for Yearly ping :) Happy New Year.

I've just saw that I still have my workaround applied but it means clearing transition in onFailedToRecycleView and it's not as efficient as doing it earlier :(

Any chance we can finally finish this?

Tolriq:
Re re re re based :)

Juansbulbisful (approved):

PR #2711, opened 2017-12-12
Replaced explicit null-checks by a declarative approach

denis-zhdanov:
Hello,

The idea is to replace explicit *null*-checks by *NotNull* annotation on target method parameters. Actual checks are generated by [Traute](http://traute.oss.harmonysoft.tech/) *javac* plugin during compilation.  

Example - *RequestBuilder*  

**Before**  

```java
public RequestBuilder<TranscodeType> apply(RequestOptions requestOptions) {
    Preconditions.checkNotNull(requestOptions);
    this.requestOptions = getMutableOptions().apply(requestOptions);
    return this;
}
```  

**After**  

```java
public RequestBuilder<TranscodeType> apply(@NonNull RequestOptions requestOptions) {
    this.requestOptions = getMutableOptions().apply(requestOptions);
    return this;
}
```

The check is present in the resulting bytecode:  

```
javap -c ./library/build/intermediates/classes/release/com/bumptech/glide/RequestBuilder.class
...
public com.bumptech.glide.RequestBuilder<TranscodeType> apply(com.bumptech.glide.request.RequestOptions);
    Code:
       0: aload_1
       1: ifnonnull     14
       4: new           #18                 // class java/lang/NullPointerException
       7: dup
       8: ldc           #19                 // String requestOptions must not be null
      10: invokespecial #20                 // Method java/lang/NullPointerException."<init>":(Ljava/lang/String;)V
      13: athrow
```

Benefits of using this approach:
* the source code is cleaner
* the code is better documented as it's clear where *null* is unexpected
* IDEs highlight possible *null* problems during development when method parameters are marked by *NonNull* annotations

TESTED: ./gradlew build

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -1,3 +1,6 @@
+plugins {
+    id "tech.harmonysoft.oss.traute" version "1.0.5"
+}
 apply plugin: 'com.android.library'
 apply plugin: 'maven'
 apply plugin: 'findbugs'
@@ -50,6 +53,11 @@ android {
         sourceCompatibility JavaVersion.VERSION_1_7
         targetCompatibility JavaVersion.VERSION_1_7
     }
+}
+
+traute {
+    javacPluginVersion = "1.0.10"
+    exceptionTexts = [ 'parameter' : '${PARAMETER_NAME} must not be null' ]
 }
 
 afterEvaluate {
--- a/library/findbugs-exclude.xml
+++ b/library/findbugs-exclude.xml
@@ -48,4 +48,25 @@
       <Bug pattern="EI_EXPOSE_REP" />
     </Match>
 
+    <!-- Null-checks are generated automatically, so, suppress them for construcors with super()/this() calls -->
+    <Match>
+        <Class name="com.bumptech.glide.GlideContext" />
+        <Bug pattern="RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE" />
+    </Match>
+
+    <Match>
+        <Class name="com.bumptech.glide.load.resource.transcode.BitmapDrawableTranscoder" />
+        <Bug pattern="RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE" />
+    </Match>
+
+    <Match>
+        <Class name="com.bumptech.glide.RequestManager" />
+        <Bug pattern="RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE" />
+    </Match>
+
+    <Match>
+        <Class name="com.bumptech.glide.RequestManager$ClearTarget" />
+        <Bug pattern="RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE" />
+    </Match>
+
 </FindBugsFilter>
--- a/library/src/main/java/com/bumptech/glide/RequestBuilder.java
+++ b/library/src/main/java/com/bumptech/glide/RequestBuilder.java
@@ -102,7 +102,6 @@ public class RequestBuilder<TranscodeType> implements Cloneable,
   @NonNull
   @CheckResult
   public RequestBuilder<TranscodeType> apply(@NonNull RequestOptions requestOptions) {
-    Preconditions.checkNotNull(requestOptions);
     this.requestOptions = getMutableOptions().apply(requestOptions);
     return this;
   }
@@ -126,7 +125,7 @@ public class RequestBuilder<TranscodeType> implements Cloneable,
   @CheckResult
   public RequestBuilder<TranscodeType> transition(
       @NonNull TransitionOptions<?, ? super TranscodeType> transitionOptions) {
-    this.transitionOptions = Preconditions.checkNotNull(transitionOptions);
+    this.transitionOptions = transitionOptions;
     isDefaultTransitionOptionsSet = false;
     return this;
   }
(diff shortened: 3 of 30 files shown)


denis-zhdanov:
Thanks, I've signed it.

sjudd:
Thanks for sending this, it's a good idea. 

A couple things I have to consider:
1. How would this function in Glide's internal build where traute is probably not available?
2. Is there a performance overhead to checking for null? Typically I'd think the answer is no, but some methods may be called frequently enough or the number of null checks might be sufficient to introduce some overhead.

sjudd:
Two more things to think about before I forget:

1. How would this compare to nullability analysis (a build time check)? 
2. How many of Glide's bugs have been related to nullability issues?

denis-zhdanov:
Hello,

Thanks for the feedback!
Regarding the questions:
*> How would this function in Glide's internal build where traute is probably not available?*  
What do you mean by the *Glide's internal build*?
*> Is there a performance overhead to checking for null? Typically I'd think the answer is no, but some methods may be called frequently enough or the number of null checks might be sufficient to introduce some overhead.*  
It doesn't add any overhead because it just replaces explicit calls to *Preconditions.checkNotNull()* to generated in-place checks. E.g. consider [Option.java](https://github.com/bumptech/glide/pull/2711/files#diff-410c36293bfb0f7727e7391368e8fd59):  

**Source code before**  

```java
private Option(String key, T defaultValue, CacheKeyUpdater<T> cacheKeyUpdater) {
    this.key = Preconditions.checkNotEmpty(key);
    this.defaultValue = defaultValue;
    this.cacheKeyUpdater = Preconditions.checkNotNull(cacheKeyUpdater);
}
```  

**Source code after**  

```java
private Option(String key, T defaultValue, @NonNull CacheKeyUpdater<T> cacheKeyUpdater) {
    this.key = Preconditions.checkNotEmpty(key);
    this.defaultValue = defaultValue;
    this.cacheKeyUpdater = cacheKeyUpdater;
}
```  

**Resulting bytecode looks like if it's compiled from the sources below**  

```java
private Option(String key, T defaultValue, @NonNull CacheKeyUpdater<T> cacheKeyUpdater) {
    if (cacheKeyUpdater == null) {
        throw new NullPointerException("cacheKeyUpdater must not be null");
    }
    this.key = Preconditions.checkNotEmpty(key);
    this.defaultValue = defaultValue;
    this.cacheKeyUpdater = cacheKeyUpdater;
}
``` 
I.e. the resulting code is negligibly faster because the *null*-check is inlined.

*> How would this compare to nullability analysis (a build time check)?*  
This approach helps with built-time check because now target method parameters are marked by *NotNull* annotations, so, IDEs can highlight possible *NPE* - [Android Studio documentation](https://developer.android.com/studio/write/annotations.html#adding-nullness)  
*> How many of Glide's bugs have been related to nullability issues?*  
I'm afraid I do not have that information

denis-zhdanov:
Hey @sjudd 

Could you please share your thoughts on the idea of adding *Traute* to your project?

Regards, Denis

sjudd:
> > How would this function in Glide's internal build where traute is probably not available?

> What do you mean by the Glide's internal build?

Glide is built by Bazel internally at Google. 

> > Is there a performance overhead to checking for null? Typically I'd think the answer is no, but some methods may be called frequently enough or the number of null checks might be sufficient to introduce some overhead.

> It doesn't add any overhead because it just replaces explicit calls to Preconditions.checkNotNull() to generated in-place checks.

It will if it adds an explicit null check everywhere the ``@NonNull`` annotation is used. Not all places that have ``@NonNull`` have explicit checks. Or at least it will add whatever overhead there is to checking non-null. In some cases it's likely these checks will be redundant because there will be a call on the non-null object in the method. The explicit preconditions checks are most useful in constructors or initialization methods where the object will be called in some other method later on.

> > How would this compare to nullability analysis (a build time check)?

> This approach helps with built-time check because now target method parameters are marked by NotNull annotations, so, IDEs can highlight possible NPE - Android Studio documentation

``@NonNull`` gives you this without the plugin. The plugin avoids explicit null checks in the code and/or provides some additional run time guarantees depending on how the ``@NonNull`` annotation is used right?

Nullability analysis theoretically provides you compile time checks so you can discover problems before you run the code in question. I'm not sure how robust or accurate these are. I think Uber has one: https://github.com/uber/NullAway.

One other thought - have you tested this to see what happens to line numbers in stack traces? Are they altered to account for the added null check so that line numbers are offset? Or are the original line numbers retained?

denis-zhdanov:
Hello,

1. I didn't find a way to apply javac plugins to *Bazel* builds. [Asked on SO](https://stackoverflow.com/questions/47920495/is-it-possible-to-put-an-external-dependency-to-javac-classpath-in-bazel) in case I missed something obvious. Just curious - why this project is configured to use *Gradle* instead of *Bazel* then?
2. not-*null* check is a single [ifnonnull](https://cs.au.dk/~mis/dOvs/jvmspec/ref-ifnonnul.html) bytecode instruction. I didn't perform exact measurement, but it just doesn't have a chance to have any performance impact
3. Sure, line numbers are preserved. There are a number of tests for that like [this one](https://github.com/denis-zhdanov/traute/blob/master/core/test/src/test/java/tech/harmonysoft/oss/traute/test/suite/MethodParameterTest.java#L109) or [this](https://github.com/denis-zhdanov/traute/blob/master/core/test/src/test/java/tech/harmonysoft/oss/traute/test/suite/MethodParameterTest.java#L135)

sjudd:
Gradle and the non-google build came first, but Gradle isn't used by most Android apps at Google (most use a variant of Bazel), so we needed a solution for that. 

At some point in the future it may make sense to switch to Bazel for the external build. At the moment there's a lot of Gradle based infrastructure in Glide's build and Gradle is more clearly tied to Android externally, so I'm not sure it makes sense right now. 

I'm also working on a change that will integrate Uber's nullability checker, it might make for an interesting test case. I don't think either precludes the other.

TWiStErRob:
Another thing to consider: I'm assuming this plugin works by adding an `annotationProcessor` behind the scenes, so we can say goodbye to incremental builds on `:library`.

PR #2712, opened 2017-12-12
Add nullability annotations to gif modules

SUPERCILEX:
More progress on nullability.

I've also noticed https://github.com/bumptech/glide/pull/2711... looks like it would combine nicely with my work. 😄


sjudd:
Thanks!

PR #2720, opened 2017-12-14
Add more models affected by #738

strooooke:
Added more models requiring bitmap lock; see #738 

Directly observed on XT1032 API level 22 and XT1068 API level 23; other models added as variants of those as listed by wikipedia.


sjudd:
Great thanks!

PR #2721, opened 2017-12-14
Fix all annotation testing issues on Windows

### LINKED ISSUE #2709: Get annotation processor unit tests running on windows without local changes
See #2703 for context.

SUPERCILEX:
Fixes #2709. It's a tad uglier than I would have liked, but running `./gradlew :annotation:compiler:test:test` is all green now! 😄


sjudd:
In the long run maybe we could either define a truth subject or move the conversion into helper methods to, but this is great, thanks!

SUPERCILEX:
Alrighty, SGTM.

PR #2722, opened 2017-12-14
Cleanup

SUPERCILEX:
Some cleanup I've noticed from working on the nullability annotation stuff:
- Use Java 6 foreach loops
- Add a set of forgotten nullability annotations
- Miscellaneous

### DIFF
--- a/library/src/main/java/com/bumptech/glide/load/ImageHeaderParserUtils.java
+++ b/library/src/main/java/com/bumptech/glide/load/ImageHeaderParserUtils.java
@@ -31,7 +31,9 @@ public final class ImageHeaderParserUtils {
     }
 
     is.mark(MARK_POSITION);
-    for (ImageHeaderParser parser : parsers) {
+    //noinspection ForLoopReplaceableByForEach to improve perf
+    for (int i = 0, size = parsers.size(); i < size; i++) {
+      ImageHeaderParser parser = parsers.get(i);
       try {
         ImageType type = parser.getType(is);
         if (type != ImageType.UNKNOWN) {
@@ -52,7 +54,9 @@ public final class ImageHeaderParserUtils {
       return ImageType.UNKNOWN;
     }
 
-    for (ImageHeaderParser parser : parsers) {
+    //noinspection ForLoopReplaceableByForEach to improve perf
+    for (int i = 0, size = parsers.size(); i < size; i++) {
+      ImageHeaderParser parser = parsers.get(i);
       ImageType type = parser.getType(buffer);
       if (type != ImageType.UNKNOWN) {
         return type;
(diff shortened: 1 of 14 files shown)


library/src/main/java/com/bumptech/glide/load/ImageHeaderParserUtils.java
```diff
@@ -31,9 +31,7 @@ public static ImageType getType(List<ImageHeaderParser> parsers, @Nullable Input
     }
 
     is.mark(MARK_POSITION);
-    int size = parsers.size();
-    for (int i = 0; i < size; i++) {
-      ImageHeaderParser parser = parsers.get(i);
+    for (ImageHeaderParser parser : parsers) {
```
sjudd: These are deliberate. Using for each allocates an iterator, which we avoid by writing the loop ourselves. Same with the rest of the loops unfortunately.

library/src/main/java/com/bumptech/glide/module/AppliesOptions.java
```diff
@@ -4,7 +4,7 @@
 import com.bumptech.glide.GlideBuilder;
 
 /**
- * An internal interface, to be removed when {@link GlideModule}s are removed.
```
sjudd: This was correct before, we want to refer to the deprecated GlideModule class.

library/src/main/java/com/bumptech/glide/load/ImageHeaderParserUtils.java
SUPERCILEX: Awwww, really? Is that perf improvement necessary for _every_ other for loop or just the ones in this class? Also, do you mind me adding a comment explaining the perf improvement to each loop so other people don't make the same mistake?

library/src/main/java/com/bumptech/glide/module/AppliesOptions.java
SUPERCILEX: Oops, SGTM.

library/src/main/java/com/bumptech/glide/load/ImageHeaderParserUtils.java
sjudd: I usually default to starting with for each, then switch when I see the allocations in the allocation tracker during scrolling. 

On ART devices this kind of optimization is probably irrelevant, but avoiding object allocations can make a pretty big difference on dalvik devices. 

You're more than welcome to add a comment.

SUPERCILEX: Oh, got it. I'm happy as long as you start with the most readable version first. 😄

SUPERCILEX: Last question: is there any reason to cache `size` and not just call it repeatedly in the if statement as is the standard? It's just a field in `ArrayList` so calling it should be dirt cheap.

sjudd: It keeps the performance the same in both cases, at least theoretically. I don't have much evidence that it matters, and in most cases in Glide it's probably not called often enough to matter at all, but it's easy to write and not particularly less readable.

Effective java mentions that for each might be slightly more performant than a standard loop because it memoizes the size: https://stackoverflow.com/a/256861/800716.

We've also seen some interesting behavior in ART in particular where keeping local copies of variables around can make a huge difference in some versions of Android in some loops: e7a494253187f14bc73a909f522e4da2ce093920

SUPERCILEX: @sjudd Alrighty, thanks for answering all my questions—that's some really cool stuff! 😄

sjudd:
Thanks!

PR #2729, opened 2017-12-17
Util nullability

SUPERCILEX:
The entire `util` package is done!!! 😄

### DIFF
--- a/library/src/main/java/com/bumptech/glide/util/ContentLengthInputStream.java
+++ b/library/src/main/java/com/bumptech/glide/util/ContentLengthInputStream.java
@@ -1,6 +1,7 @@
 package com.bumptech.glide.util;
 
 import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
 import android.text.TextUtils;
 import android.util.Log;
 import java.io.FilterInputStream;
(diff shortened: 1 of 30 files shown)


library/src/main/java/com/bumptech/glide/util/ContentLengthInputStream.java
```diff
@@ -58,12 +62,12 @@ public synchronized int read() throws IOException {
   }
 
   @Override
-  public int read(@NonNull byte[] buffer) throws IOException {
```
SUPERCILEX: I prefer to not add nullability annotations to overridden methods who don't already have them, what do you think? It's making assumptions about third party codebases... (I know this is the java lang so it's safe, but as a general rule I prefer not to)

library/src/main/java/com/bumptech/glide/util/LruCache.java
```diff
@@ -119,17 +119,16 @@ public synchronized Y get(T key) {
    * @param key  The key to add the item at.
    * @param item The item to add.
    */
-  public synchronized Y put(T key, @Nullable Y item) {
+  @Nullable
+  public synchronized Y put(@NonNull T key, @NonNull Y item) {
```
SUPERCILEX: The `item` nullability was incorrect: `getSize` would blow on null so it can't be null (also didn't notice anyone in our codebase using null).

SUPERCILEX: Hmmm, the tests have told me that's wrong. Is 63a73de correct?

SUPERCILEX:
BTW, the build failed because the emulator is out of date (which is unrelated to this PR).

library/src/main/java/com/bumptech/glide/util/ContentLengthInputStream.java
sjudd: Doesn't matter to me, I think it's ok either way.

SUPERCILEX: Sweet!

sjudd:
Thanks!

PR #2734, opened 2017-12-19
Change library dependencies configuration from api to compileOnly

### LINKED ISSUE #2730: The 4.4.0 version is conflict with appcompat-v7:26.1.0
**Glide Version**:4.4.0

**Integration libraries**:No

**Device/Android Version**: Google Pixel

The activity has a textview & imageview & imageview is using glide to load image, when the activity launches, it'll crash.

**Glide load line / `GlideModule` (if any) / list Adapter code (if any)**:GlideModule
```java
GlideApp.with(imageView.getContext())
                .load(url)
                .placeholder(placeholder)
                .error(error)
                .into(imageView);
```

**Stack trace / LogCat**:
```ruby
java.lang.NoSuchMethodError: No static method getFont(Landroid/content/Context;ILandroid/util/TypedValue;ILandroid/widget/TextView;)Landroid/graphics/Typeface; in class Landroid/support/v4/content/res/ResourcesCompat; or its super classes (declaration of 'android.support.v4.content.res.ResourcesCompat' 
at android.support.v7.widget.TintTypedArray.getFont(TintTypedArray.java:119)
```

rivendell1984:
## Description

Library's dependencies' version is 27.0.2, this change can fix the conflict with version 26.1.0 support dependency in user's code.

fixed #2730

### DIFF
--- a/integration/gifencoder/build.gradle
+++ b/integration/gifencoder/build.gradle
@@ -2,6 +2,7 @@ apply plugin: 'com.android.library'
 
 dependencies {
     implementation project(':library')
+    api "com.android.support:support-annotations:${ANDROID_SUPPORT_VERSION}"
 
     testImplementation project(":testutil")
     testImplementation "com.google.truth:truth:${TRUTH_VERSION}"
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -9,7 +9,7 @@ dependencies {
     api project(':third_party:gif_decoder')
     api project(':third_party:disklrucache')
     api project(':annotation')
-    api "com.android.support:support-fragment:${ANDROID_SUPPORT_VERSION}"
+    compileOnly "com.android.support:support-fragment:${ANDROID_SUPPORT_VERSION}"
     compileOnly "com.android.support:appcompat-v7:${ANDROID_SUPPORT_VERSION}"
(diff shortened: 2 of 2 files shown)


rivendell1984:
I signed it!

2017-12-19 16:22 GMT+08:00 googlebot <notifications@github.com>:

sjudd:
I don't think this makes sense. With the PR, you'll get a compilation error unless you add a dependency on the support library at the same time that you add a dependency on Glide, which isn't how Gradle is supposed to work.

Using Gradle's existing tools to handle conflicting dependencies if you don't want Glide's version of the support library or don't like how Gradle is resolving your dependencies seems like the right answer.

Let me know if you think that's incorrect.

rivendell1984:
I have added support-annotations in gifencoder like other third-party library did(okhttp,gifdecoder), and there is no complication fail again.(travis check failed is same as issue #2738)
I agree to use gradle configuration resolving the dependency conflict, but I think it could be better if this library upgrade transparently for glide users.

sjudd:
We can't be transparent about which version we use. Using compileOnly breaks users who don't depend on the support library already.

rivendell1984:
Agree, I'll close this PR

PR #2746, opened 2017-12-22
Finish adding nullability annotations to Registry class and its dependencies

SUPERCILEX:
The entire root package is done! 😄


sjudd:
Thanks!

PR #2750, opened 2017-12-22
Add nullability annotations to RequestOptions

SUPERCILEX:
It's just a whole latta return `@NonNull` annotations.


sjudd:
Thanks!

PR #2751, opened 2017-12-22
fix the emulator -no-window option issue

### LINKED ISSUE #2738: API 16, 17, 18 emulators are failing on travis.
Log attached: [failing_emulator.txt](https://github.com/bumptech/glide/files/1573838/failing_emulator.txt)

I've filed an issue on travis' GitHub after finding that a previously successful commit failed when I retried it: https://github.com/travis-ci/travis-ci/issues/8946. 

It seems like something's changed in the last day or so that started making these emulator tests fail.

Since I haven't found a way to fix this so far, I'll disable these tests for the time being.

rivendell1984:
## Description

Remove the -no-window option for android-16,17,18, and start a xvfb for them. 

## Motivation and Context

Fix #2738

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,14 +20,17 @@ env:
   matrix:
     - COMPONENT=unit
     - COMPONENT=firebase
-      # TODO(#2738): Re-enable these when we find a fix.
-      #    - COMPONENT=instrumentation ANDROID_TARGET=16
-      #    - COMPONENT=instrumentation ANDROID_TARGET=17
-      #    - COMPONENT=instrumentation ANDROID_TARGET=18
+    - COMPONENT=instrumentation ANDROID_TARGET=16
+    - COMPONENT=instrumentation ANDROID_TARGET=17
+    - COMPONENT=instrumentation ANDROID_TARGET=18
     - COMPONENT=instrumentation ANDROID_TARGET=19
     - COMPONENT=instrumentation ANDROID_TARGET=21
     - COMPONENT=instrumentation ANDROID_TARGET=22
     - COMPONENT=samples
+
+before_install:
+  - sudo apt-get -qq update
+  - sudo apt-get install -y xvfb
 
 before_script:
   - ./scripts/travis_before_script.sh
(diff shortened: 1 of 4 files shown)


sjudd:
Interesting! Thanks for looking in to this!

I actually already attached what I think is a fix for the issue: https://github.com/sjudd/glide/commit/0e3172a86719029b0642677c59b442f0e3bdd687.

It looks like -no-window only doesn't work for the Nexus 5X? Or I got lucky and it does work for Nexus 4?

.travis.yml
```diff
@@ -20,15 +20,18 @@ env:
...
-      #    - COMPONENT=instrumentation ANDROID_TARGET=18
+    - COMPONENT=instrumentation ANDROID_TARGET=16
+    - COMPONENT=instrumentation ANDROID_TARGET=17
+    - COMPONENT=instrumentation ANDROID_TARGET=18
     - COMPONENT=instrumentation ANDROID_TARGET=19
     - COMPONENT=instrumentation ANDROID_TARGET=21
     - COMPONENT=instrumentation ANDROID_TARGET=22
     - COMPONENT=samples
 
+before_install:
```
sjudd: You can probably just do this in the travis-create-emulator script, that way you do'nt have to add the sdkmanager update line and you'll avoid installing these on the matrices where they're not used (unit, samples, firebase)

sjudd:
So I submitted the --device Nexus 4 change. It seems like a smaller unknown than running the emulators with a window and it looks like it's about a minute faster.

That said, I haven't spent a lot of time trying to get emulators working or trying to optimize the boot times. If you think your approach is better in the long run or have any other tips for running emulator tests quickly and reliably on travis please let me know.

Thanks again.

rivendell1984:
Thanks for your feedback. I think your solution is better, it's simple & can reduce the building time in Travis CI.

Below is just comment for this issue, and I think it's emulator's issue which not worth you spent a lot of time on it:
I have made some testing, if the emulator's resolution is bigger than 768x1280(which is Nexus 4's resolution, you're lucky!), for example Nexus 5X/Nexus 6/Pixel, starting emulator with -no-window option & android-16 image will be break by "Display subsystem failed", and many properties haven't created, for example init.svc.bootanim, which terminate the building.

sjudd:
Sure thank you for the detail and investigation. I've filed a bug internally and will try to track down the right person to take a look, I'm sure this detail will be very helpful.

rivendell1984:
OK, thank you & I'll close this PR.

PR #2755, opened 2017-12-23
Upgrade to API 27

### LINKED ISSUE #2730: The 4.4.0 version is conflict with appcompat-v7:26.1.0

SUPERCILEX:
@sjudd The API 27 sources were just released so we can close #2730.


gradle.properties
```diff
@@ -37,8 +37,8 @@ PMD_VERSION=5.4.0
 FINDBUGS_VERSION=3.0.0
 JACOCO_VERSION=0.7.1.201405082137
 
-COMPILE_SDK_VERSION=26
-TARGET_SDK_VERSION=26
+COMPILE_SDK_VERSION=27
+TARGET_SDK_VERSION=27
```
SUPERCILEX: I don't believe there were any API changes that would affect Glide... just making sure you're aware of the target change.

sjudd: So the support library change brings up a common and difficult issue, which is that it's not uncommon for developers to target or compile against older versions for a long period of time. Target API changes can be particularly difficult to manage. For example there's a change in the way non-compliant notifications are handled for apps that target API 27 that causes a crash that's not present for people targeting API 26.

Often when we try to bump versions early, we get complaints or people avoid updating to the latest version of Glide. The same issue applies with the Java version only more so.

I don't know when the right time to update is frankly. Maybe we should have some policy that says we will update after N weeks? I'm not really sure what N should be.

SUPERCILEX: Oh I see, well that's unfortunate. Yeah, not sure what N should be either. Does the source release mean we can still update the compile sdk, or do we have to wait to update anything?

sjudd: Compile sdk is probably fine. It may still require an upgrade for users, it's just the easiest. 

I'll think about it a bit more and maybe check some other Android libraries and see if there's any kind of consensus.

sjudd: All other libraries I've checked are still using API 26. That said, Glide is likely to continue to want to use recently released APIs and as a result we're probably much less tolerant of old target sdk versions.

Let's merge this, the previous version of Glide targeted 27 anyway. In the long run we might consider some kind of compat library (like Glide-next) that allows access to newer APIs without forcing everyone who uses Glide to upgrade at once.

SUPERCILEX: SGTM, I've merged to force a build.

sjudd:
Thanks!

PR #2762, opened 2017-12-26
 Clean up synthetic accessor method rot

TWiStErRob:
## Description

Since PMD 5.5.4 there's a [`AccessorMethodGeneration`](https://pmd.github.io/pmd-5.8.1/pmd-java/rules/java/design.html#AccessorMethodGeneration) rule in `design` category thanks to [Jake's request](https://sourceforge.net/p/pmd/bugs/1496/ ). The PR removes 20 synthetic methods and prevents adding any new ones.

In this PR:
 * simply updating PMD to 5.5.4+ enabled the rule.
 * didn't update to 6.0.0, only to 5.8.1 to not have to deal with deprecations in this PR
 * fixed all synthetic warnings, and annotated the members
 * changed the message of PMD to refer to our `@Synthetic` annotation for future ci build failures
 * work around https://github.com/pmd/pmd/issues/808 and suppress other false positives
 * fix a few unrelated warnings in the touched files
 * re-enabled `AvoidDuplicateLiterals` and simplified `UncommentedEmptyConstructor`
 * fixed/ignore tests to run on Windows
 * Upgrade error-prone to include fix for https://github.com/google/error-prone/issues/812 (what's your take on this?); also fixing errorprone version should minimize `maven-metadata.xml` downloads.

In some files (`ActiveResources`, `DecodeJob`, `ViewTarget.OnAttachStateChangeListener`, ) I took a different approach than making everything visible. It looks like in those classes the inner classes had some field envy to the outer class, so I moved the method to the outer class so there's only one visible `@Synthetic` entry needed. In some places I also protected the published methods with `@RestrictTo(Scope.LIBRARY)` to flag abuse.

## Motivation and Context

Followup to https://github.com/bumptech/glide/pull/1556#issuecomment-257479358

After this PR only `disklrucache` synthetic accessors (22) remain.

To check count, run
```
gradlew :glide:jar && dex-method-list glide/build/libs/glide-full-4.5.0-SNAPSHOT.jar  | grep "access"
```
---

@sjudd Can you please add `!.idea/inspectionProfiles/Project_Default.xml` to `.gitignore` and commit that file so everyone sees the same warnings in AS? I wanted to add `Java/J2ME issues/Synthetic accessor call` inspection to the config, but I noticed that file is not version controlled.


library/src/main/java/com/bumptech/glide/RequestBuilder.java
```diff
@@ -571,8 +574,9 @@ protected RequestOptions getMutableOptions() {
     return into(target, /*targetListener=*/ null);
   }
 
+  @RestrictTo(Scope.LIBRARY)
```
sjudd: I'm ok not using this for non-public methods. I assume anything that's non-public can be broken and even some of the public APIs we have are breakable (those really should be marked with ``@RestrictTo`` or some equivalent.

Fine to keep it, just as a note for the future (or a discussion point).

library/src/main/java/com/bumptech/glide/load/engine/DecodeJob.java
```diff
@@ -41,26 +41,22 @@
     Poolable {
   private static final String TAG = "DecodeJob";
 
-  @SuppressWarnings("WeakerAccess")
-  @Synthetic
-  final DecodeHelper<R> decodeHelper = new DecodeHelper<>();
+  private final DecodeHelper<R> decodeHelper = new DecodeHelper<>();
```
sjudd: Nice! Thanks!

library/src/main/java/com/bumptech/glide/load/engine/bitmap_recycle/SizeStrategy.java
```diff
@@ -111,7 +111,7 @@ static String getBitmapString(int size) {
   static class KeyPool extends BaseKeyPool<Key> {
 
     public Key get(int size) {
-      Key result = get();
+      Key result = super.get();
```
sjudd: Out of curiosity, what does the super do?

TWiStErRob: Open the file on `master`, there's a warning on it. It clarifies which `get` is being called: both `BaseKeyPool` and `SizeStrategy` have a `get` name in scope, so the super clears it up which one. Actually let me check that again, because it's a static inner class.

library/src/main/java/com/bumptech/glide/RequestBuilder.java
TWiStErRob: Lot of negation there :)
You're right: `default` methods don't need this, if someone uses "internal" (name/keyword for default visibility in other languages) API, they can fix it if we break something. `protected` on public classes may need this, and `public` methods that are only public because of packages, and not because it's public API also need this. I'll remove from `default` ones.

library/src/main/java/com/bumptech/glide/load/engine/bitmap_recycle/SizeStrategy.java
TWiStErRob: image
so the circumstances I described above stand, but it may be an inspection false positive as the inner class is `static` and the `get` signature is different in the outer class. I guess it still plays nicer with the `super.`.

TWiStErRob:
@sjudd merged master into this, resolved conflict, removed `@RestrictTo`, answered questions.
I suggest a "Squash and Merge" for this PR, and then let's revisit PMD 6 (you have a lot of excludes).

library/src/main/java/com/bumptech/glide/load/engine/bitmap_recycle/SizeStrategy.java
TWiStErRob: https://youtrack.jetbrains.com/issue/IDEA-184421

TWiStErRob:
Also bump on https://github.com/TWiStErRob/android-realistic/blob/master/.gitignore#L30-L36

sjudd:
Thanks! I'll add the files you suggest to the gitignore, no particular reason to exclude them I don't think.

PR #2789, opened 2018-01-07
Ensure all generated classes are fully nullabilitily annotated

SUPERCILEX:
And voila, every generated class now has nullability annotations everywhere! 😄

PS: fae9444e85857bc524511f9b78063765bbcc760d is amazing! When I ran the build, I was like "why are the tests passing!?" Anyway, thanks! ❤️


library/src/main/java/com/bumptech/glide/module/AppliesOptions.java
```diff
@@ -1,11 +1,13 @@
 package com.bumptech.glide.module;
 
 import android.content.Context;
+import android.support.annotation.NonNull;
 import com.bumptech.glide.GlideBuilder;
 
 /**
  * An internal interface, to be removed when {@link GlideModule}s are removed.
  */
+@Deprecated
```
SUPERCILEX: These changes got buried at the end, but I want to make sure you see them. I've deprecated `AppliesOptions`, `RegistersComponents`, and `ManifestParser` since they're all linked to the old `GlideModule`. Was there a reason for not deprecating them in the first place?

sjudd: Nope no reason I don't think. It adds a couple of extra IDE warnings, but nowhere important.

library/src/main/java/com/bumptech/glide/RequestManager.java
```diff
@@ -118,6 +118,7 @@ public RequestManager(
     glide.registerRequestManager(this);
   }
 
+  @NonNull
```
sjudd: This is a void method, it shouldn't need NonNull?

SUPERCILEX: Oh lol, Intellij didn't even complain. 😕

SUPERCILEX:
@sjudd No problem! And yeah, really sorry about the build breakages, I should have warned you that Kotlin strictly enforces type inference. Take all the time you need! 😊

sjudd:
Thanks!

PR #2797, opened 2018-01-10
Start adding nullability annotations to load package

SUPERCILEX:


SUPERCILEX:
@sjudd Any updates? 😊

library/src/main/java/com/bumptech/glide/load/resource/bitmap/DrawableTransformation.java
```diff
@@ -58,6 +58,7 @@ public DrawableTransformation(Transformation<Bitmap> wrapped, boolean isRequired
         return resource;
       }
     }
+    // TODO bitmapResourceToTransform can be null, what to do?
```
sjudd: It will be non-null if you reach this point, I believe you can remove this todo.

SUPERCILEX: @sjudd Oh you're right, thanks!

SUPERCILEX:
@sjudd No problem! (And we have a green build!)

sjudd:
Thanks!

SUPERCILEX:
Awesome!

PR #2873, opened 2018-02-05
Fix param mistake

### LINKED ISSUE #2798: MemorySizeCalculator.setMemoryCacheScreens()  logic mistake
When I read glide source code, I found there is a logic mistake in MemorySizeCalculator.setMemoryCacheScreens().

```
 /**
 * Sets the number of device screens worth of pixels the
 * {@link com.bumptech.glide.load.engine.cache.MemoryCache} should be able to hold and
 * returns this Builder.
 */
public Builder setMemoryCacheScreens(float memoryCacheScreens) {
  Preconditions.checkArgument(bitmapPoolScreens >= 0,
      "Memory cache screens must be greater than or equal to 0");
  this.memoryCacheScreens = memoryCacheScreens;
  return this;
}
```
I think `Preconditions.checkArgument()`'s first param should be `memoryCacheScreens >= 0`. is it? Tks

tuesda:
## Description

Fix a little mistake about param.
Close #2798


sjudd:
Thanks!

Issue #2810, opened 2018-01-13
Glide's `trimMemory()` implementation doesn't trim when foregrounded


ygnessin:
I am profiling memory on my app usage and was looking at the built-in `tremMemory(int level)` methods in `LruResourceCache` and `LruBitmapPool` to see what their behavior is when I invoke them from my application's `onTrimMemory` method. The trim memory levels are documented [here](https://developer.android.com/reference/android/content/ComponentCallbacks2.html).

Here is Glide's implementation of this, for both of the aforementioned classes (and probably others):
```java
    public void trimMemory(int level) {
        if (level >= android.content.ComponentCallbacks2.TRIM_MEMORY_MODERATE) {
            // Nearing middle of list of cached background apps
            // Evict our entire bitmap cache
            clearMemory();
        } else if (level >= android.content.ComponentCallbacks2.TRIM_MEMORY_BACKGROUND) {
            // Entering list of cached background apps
            // Evict oldest half of our bitmap cache
            trimToSize(getMaxSize() / 2);
        }
    }
```
The problem with this is that Glide is treating the `level` as a sequential number. It assumes that higher is worse. But that is not actually the case, if you look at the constant value of the levels in the linked documentation.

In particular, notice that unless `level` is at least 40, Glide won't trim any memory. And all levels 40 and above are only triggered if the app is backgrounded. But if the device is running critically low on memory while the app is foregrounded, Glide will never trim memory. Take, for instance, the most severe level, `TRIM_MEMORY_RUNNING_CRITICAL` which gets called when the app is foregrounded and the device has no remaining memory. This level's constant value is 15, which won't trigger a memory trim. And unless the app is backgrounded, none of the levels before this will trigger a trim either, because all of them are under 40.

I'd like to propose modifying this default `trimMemory()` behavior to take foreground events into account. If the community agrees this is a good idea, I'd also be happy to open a pull request myself when I have the time, but wanted to propose the idea first. Thanks for taking the time to read this.

**Glide Version**: 3.8.0 (but checked the source and this is still in issue on 4.5.0)

sjudd:
At the moment this is intended. If there's a good reason why we should expect those error messages in the foreground and why clearing the memory cache or bitmap pool will help I'd love to hear it.

[...]

PR #2889, opened 2018-02-10
Evict oldest half of cache on TRIM_MEMORY_RUNNING_CRITICAL

ygnessin:
## Description

Trim half of resource cache allocations when receiving [TRIM_MEMORY_RUNNING_CRITICAL](https://developer.android.com/reference/android/content/ComponentCallbacks2.html#TRIM_MEMORY_RUNNING_CRITICAL) in `trimMemory()`

See #2810 

Also, while in the code I updated some outdated comments, and replaced `maxSize` with `getMaxSize()` in `LruBitmapPool`

## Motivation and Context

Glide is excellent at trimming memory when backgrounded, but fails to do anything when `trimMemory()` is received while foregrounded. This fixes that, which may save an app from slowing down or encountering an OOM while foregrounded, especially if the foreground context isn't the one using the images.

@sjudd had some good points about this strategy in #2810, which is why I've elected to be pretty conservative by only evicting half the cache when TRIM_MEMORY_RUNNING_CRITICAL is encountered. I also think this makes sense because the next stage after CRITICAL is [TRIM_MEMORY_COMPLETE](https://developer.android.com/reference/android/content/ComponentCallbacks2.html#TRIM_MEMORY_COMPLETE), which Glide already handles by evicting the entire cache. Thus, this follows the same pattern as the backgrounded behavior: Half of the cache is evicted first, then one level up the whole cache is evicted.

If anybody thinks the strategy should be more or less aggressive I'd be happy to discuss, but hopefully that explains my rationale. I also think that if the community finds this effective, we could always make it more aggressive in a future update.

Thank you for considering my contribution!

### DIFF
--- a/library/src/main/java/com/bumptech/glide/load/engine/bitmap_recycle/LruArrayPool.java
+++ b/library/src/main/java/com/bumptech/glide/load/engine/bitmap_recycle/LruArrayPool.java
@@ -142,6 +142,8 @@ public final class LruArrayPool implements ArrayPool {
       clearMemory();
     } else if (level >= android.content.ComponentCallbacks2.TRIM_MEMORY_UI_HIDDEN) {
       evictToSize(maxSize / 2);
+    } else if (level == android.content.ComponentCallbacks2.TRIM_MEMORY_RUNNING_CRITICAL) {
+      evictToSize(maxSize / 2);
     }
   }
 
--- a/library/src/main/java/com/bumptech/glide/load/engine/bitmap_recycle/LruArrayPool.java
+++ b/library/src/main/java/com/bumptech/glide/load/engine/bitmap_recycle/LruArrayPool.java
@@ -140,9 +140,8 @@ public final class LruArrayPool implements ArrayPool {
   public synchronized void trimMemory(int level) {
     if (level >= android.content.ComponentCallbacks2.TRIM_MEMORY_BACKGROUND) {
       clearMemory();
-    } else if (level >= android.content.ComponentCallbacks2.TRIM_MEMORY_UI_HIDDEN) {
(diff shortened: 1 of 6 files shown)


library/src/main/java/com/bumptech/glide/load/engine/bitmap_recycle/LruBitmapPool.java
```diff
@@ -220,7 +220,9 @@ public void trimMemory(int level) {
     if (level >= android.content.ComponentCallbacks2.TRIM_MEMORY_BACKGROUND) {
       clearMemory();
     } else if (level >= android.content.ComponentCallbacks2.TRIM_MEMORY_UI_HIDDEN) {
```
sjudd: Use || instead of duplicating the size logic

library/src/main/java/com/bumptech/glide/load/engine/bitmap_recycle/LruArrayPool.java
```diff
@@ -142,6 +142,8 @@ public synchronized void trimMemory(int level) {
       clearMemory();
     } else if (level >= android.content.ComponentCallbacks2.TRIM_MEMORY_UI_HIDDEN) {
       evictToSize(maxSize / 2);
+    } else if (level == android.content.ComponentCallbacks2.TRIM_MEMORY_RUNNING_CRITICAL) {
```
sjudd: Use || instead of duplicating the size logic.

ygnessin:
@sjudd thanks for the feedback, I addressed your comments

sjudd:
Thanks!

Issue #2808, opened 2018-01-12
How to using "android-gif-drawable" library to decode gif instead of  "glide internal com.bumptech.glide.load.resource.gif.StreamGifDecoder"


Guang1234567:
@sjudd 

I want to using [android-gif-drawable](https://github.com/koral--/android-gif-drawable) to decode gif.

Should i do like this ?

Gradle (Android Studio)
Insert the following dependency to build.gradle file of your project.

```
dependencies {
    compile 'pl.droidsonroids.gif:android-gif-drawable:1.2.10'
}

```

```java

@GlideModule
public class MyGifModule extends AppGlideModule {
  @Override
  public void registerComponents(@NonNull Context context, @NonNull Glide glide,
      @NonNull Registry registry) {
        /* GIFs */
        registry.prepend(
                Registry.BUCKET_GIF,
                InputStream.class,
                pl.droidsonroids.gif.GifDrawable.class,  // <-------------- here
                new MyStreamGifDecoder(registry.getImageHeaderParsers(), glide.getArrayPool()));
  }

  // Disable manifest parsing to avoid adding similar modules twice.
  @Override
  public boolean isManifestParsingEnabled() {
    return false;
  }
}

```

```java

import android.support.annotation.NonNull;

import com.bumptech.glide.load.engine.Initializable;
import com.bumptech.glide.load.resource.drawable.DrawableResource;

import pl.droidsonroids.gif.GifDrawable; // <------------------- here

/**
 * @author Administrator
 * @date 2018/1/12 14:49
 */

public class MyGifDrawableResource extends DrawableResource<GifDrawable>
        implements Initializable {

    // Public API.
    @SuppressWarnings("WeakerAccess")
    public MyGifDrawableResource(GifDrawable drawable) {
        super(drawable);
    }

    @NonNull
    @Override
    public Class<GifDrawable> getResourceClass() {
        return GifDrawable.class;
    }

    @Override
    public int getSize() {
        return drawable.getFrameByteCount() * drawable.getNumberOfFrames();
    }

    @Override
    public void recycle() {
        drawable.stop();
        drawable.recycle();
    }

    @Override
    public void initialize() {
        drawable.seekToFrameAndGet(0).prepareToDraw();
    }
}

```

```java

import android.support.annotation.NonNull;

import com.bumptech.glide.load.ImageHeaderParser;
import com.bumptech.glide.load.ImageHeaderParserUtils;
import com.bumptech.glide.load.Options;
[...]

Issue #2606, opened 2017-11-16
Listview scrolling lags loading profile images from Firebase Storage


goodrickstar:
Before initially loading images from public Firebase url, scrolling the user list (ListView) is "laggy". The scrolling smoothness reacts as if Glide is operating on the UI thread. After images are loaded into memory, scrolling is smooth. Here's the java and xml. 

![image](https://user-images.githubusercontent.com/12790922/32915725-e5cbca60-cae7-11e7-9761-3400262aec9b.png)

>     compile 'com.github.bumptech.glide:glide:4.3.0'
>     compile 'com.github.bumptech.glide:okhttp3-integration:4.3.0'
>     compile 'com.squareup.okhttp3:okhttp:3.9.0'
>     compile 'com.squareup.okio:okio:1.13.0'

`static RequestOptions profilePictureOptions = new RequestOptions().circleCrop().error(R.drawable.error).diskCacheStrategy(DiskCacheStrategy.RESOURCE);`
```
@Override
        public View getView(final int i, View convertView, final ViewGroup parent) {
            final MyViewHolder holder;
            if (convertView == null) {
                LayoutInflater inflater = (LayoutInflater) context.getSystemService(Context.LAYOUT_INFLATER_SERVICE);
                convertView = inflater.inflate(R.layout.user_list_row, parent, false);
                holder = new MyViewHolder(convertView);
                convertView.setTag(holder);
            } else holder = (MyViewHolder) convertView.getTag();
            final UserListEntry entry = users.get(i);
            String title = entry.getTitle();
            if (Integer.parseInt(entry.getUser_id()) < OG) title = "(OG)";
            if (entry.getSilenced()) title = "(" + EmojiParser.parseToUnicode(":mute:") + ")";
            if (entry.getNewbie() == 1) title = "-" + title;
            if (entry.getSubscribed()) title = "+" + title;
            holder.title.setText(title);
            holder.handle.setText(entry.getRadio_hanlde());
            holder.location.setText(entry.getHometown());
            holder.carrier.setText(entry.getCarrier());
            Glide.with(context).load(entry.getProfileLink()).apply(RadioService.profilePictureOptions).into(holder.profile);
            Glide.with(context).load(getResources().getIdentifier("star" + entry.getRank(), "drawable", "com.cb3g.channel19")).into(holder.starsIm);
            holder.mail.setOnClickListener(new View.OnClickListener() {
                @Override
                public void onClick(View v) {
[...]

PR #2892, opened 2018-02-11
Don't throw exception about closed RecyclableBufferedInputStream when it is not closed

koral--:
Potential fix for #2808

## Motivation and Context
1. Exception message is invalid because stream is not closed.
2. https://github.com/koral--/android-gif-drawable/issues/503

### DIFF
--- a/library/src/main/java/com/bumptech/glide/load/resource/bitmap/RecyclableBufferedInputStream.java
+++ b/library/src/main/java/com/bumptech/glide/load/resource/bitmap/RecyclableBufferedInputStream.java
@@ -89,8 +89,11 @@ public class RecyclableBufferedInputStream extends FilterInputStream {
   public synchronized int available() throws IOException {
     // in could be invalidated by close().
     InputStream localIn = in;
-    if (buf == null || localIn == null) {
+    if (localIn == null) {
       throw streamClosed();
+    }
+    if (buf == null) {
+      return 0;
     }
     return count - pos + localIn.available();
   }
--- a/library/test/src/test/java/com/bumptech/glide/load/resource/bitmap/RecyclableBufferedInputStreamTest.java
+++ b/library/test/src/test/java/com/bumptech/glide/load/resource/bitmap/RecyclableBufferedInputStreamTest.java
@@ -252,6 +252,12 @@ public class RecyclableBufferedInputStreamTest {
     stream.close();
     stream.available();
   }
+    
+  @Test
+  public void testAvailableReturnsZeroIfStreamIsReleased() throws IOException {
+    stream.release();
+    assertEquals(0, stream.available());
+  }
 
   @Test(expected = IOException.class)
   public void testReadThrowsIfStreamIsClosed() throws IOException {


TWiStErRob:
Lgtm, except lack of tests. @sjudd build failure looks unrelated.

sjudd:
Why would the buffer be null if the stream is not closed? The buffer is set to a non-null value in the constructor and then set to null when the stream is closed. There are a few other methods that check if the buffer is null to see if the stream is closed.

For #2808 the two issues seem to be:
1. GIfDrawable requires a direct ByteBuffer
2. GifDrawable assumes that it can continue to read from the stream after the constructor completes

Glide definitely closes the stream after the decoder's decode method is finished, so 2 is expected.

Was there some other issue there that I missed?

koral--:
> Why would the buffer be null if the stream is not closed? The buffer is set to a non-null value in the constructor and then set to null when the stream is closed.

`null` is assigned here in `release()`:
https://github.com/bumptech/glide/blob/master/library/src/main/java/com/bumptech/glide/load/resource/bitmap/RecyclableBufferedInputStream.java#L118

> GIfDrawable requires a direct ByteBuffer

It is one of the options but not the only one.

> GifDrawable assumes that it can continue to read from the stream after the constructor completes

Indeed, input source is not buffered inside `GifDrawable` InputStream has to support marking.

sjudd:
Release and close are (or should be) called at more or less the same time, it's not going to be safe to continue to use the stream.

koral--:
It does not happen in case of #2606, released stream is not closed.

sjudd:
Yes, but the stream it wraps is. If it's not closed, then it would be a significant memory leak. The expected flow is more or less:

1. ModelLoader returns a DataFetcher
2. DataFetcher returns an InputStream
3. DataRewinder wraps the InputStream with a recyleable stream
4. All relevant decoders are tried in order, with the stream rewound by the DataRewinder between each attempt
5. The DataRewinder releases its recycleable stream
6. The DataFetcher closes its stream

Glide doesn't allow decoders to retain references to unclosed streams as a rule. It's probably possible to get it to mostly work by avoiding the call to close in a custom ModelLoader/DataFetcher, but there aren't any guarantees.

koral--:
OK. Thanks for the explanation.
Is it intended that `RecyclableBufferedInputStream#available()` throws exception when stream is released but not closed?

sjudd:
Basically yes, released and closed are more or less equivalent. The only distinction is that release does not close the underlying stream, but close does.

koral--:
So at least exception message is incorrect.

sjudd:
It's mostly correct? For all practical purposes the stream is unusable. Does it make it more confusing to distinguish between the two states given that the end result is the same?

sjudd:
I am still confused though, how does this solve #2808? What did you do to test it? The various read methods in the stream throw exactly the same exception for exactly the same reason. Why does changing the behavior of available() alone matter?

koral--:
I've just checked out this project: https://github.com/Guang1234567/glide/tree/issue_503 and started analyzing what caused that issue (artifacts on the left image). Then I discovered that the stream cannot be read due to exception thrown from `available()` (`read()` calls `available()`). I saw that stream is not closed despite exception message saying it is closed. So I've modified `available()` to return 0 when stream is released and it turned out that it fixes #2808.

sjudd:
Read also asserts that the buffer is non-null: https://github.com/bumptech/glide/blob/99229725401d5777e059da7b6331134bf73fbcdf/library/src/main/java/com/bumptech/glide/load/resource/bitmap/RecyclableBufferedInputStream.java#L277

I'm still curious about why just changing the behavior of available fixes anything.

That said, I don't think we're going to move forward on this PR, so closing for now.

koral--:
> I'm still curious about why just changing the behavior of available fixes anything.

Do you wan't to (or will you) analyze it further?
Here is the example: https://github.com/Guang1234567/glide/blob/issue_503/samples/svg/src/main/java/com/bumptech/glide/samples/svg/MyStreamGifDecoder.java#L62

koral--:
@sjudd sorry I missed that comment: https://github.com/bumptech/glide/pull/2892#issuecomment-364968436

> Does it make it more confusing to distinguish between the two states given that the end result is the same?

In case like this, when something does not work and there is an exception that stream is closed but it is not closed, my first thought was that it is a bug in glide. I would say that it is very confusing in this particular case.

sjudd:
Does GifDrawable read the stream it's given eagerly in a separate thread? If so it could just be a race condition I guess where available is called but the buffered stream already has the remainder of the image in memory so it never calls read again. Seems a little far fetched.

Fundamentally Glide doesn't provide support for progressive images, including GIFs. That feature could be added, but it's not currently supported. As long as GifDrawable is reading on a separate thread after the call to the constructor completes, the only safe way to implement this is to buffer the entire image in memory, which is best done via a ByteBuffer or `byte[]`.  From that perspective this is basically broken as expected. 

My curiosity is mostly about why it's broken in this particular way. I'd expect not all of the image data to be available so you'd at best see a partial GIF or a crash when you tried to read the remainder, at least once you change `available()`. It's not urgent enough for me to spend much time on since the high level issue (lack of support for progressive images) isn't supported.

koral--:
> Does GifDrawable read the stream it's given eagerly in a separate thread
Exactly.

Thanks for the detailed explanation.

PR #2913, opened 2018-02-22
Update README.md

knightcube:
GlideApp doesn't work anymore. We use Glide now. So that is what should be there in the documentation


knightcube:
i signed it

Tolriq:
You might want to read better the doc ;) We do use GlideApp as you shoud be using the annotation processor to generate it ;)

knightcube:
yes you are right. My bad :)

knightcube:
i will close the pull request

PR #2924, opened 2018-03-01
Improved @NonNull annotation check for Kotlin

mariotaku:
## Description
Improved @NonNull annotation check for Kotlin.

## Motivation and Context
When writing Glide extensions in Kotlin, I got complains from annotation processor:
````
ExtensionName#methodName(com.bumptech.glide.RequestBuilder<DrawableType>) is missing the android.support.annotation.NonNull annotation, please add it to ensure that your extension methods are always returning non-null values
````
Kotlin annotates NonNull/Nullable to every return type/parameter. After inspecting bytecode from its source, I found Kotlin uses `org.jetbrains.annotations.NotNull` for @NonNull annotation instead.

Adding support @NonNull annotation in Kotlin is unnecessary. This commit adds JetBrains NonNull annotation check.

### DIFF
--- a/annotation/compiler/src/main/java/com/bumptech/glide/annotation/compiler/GlideExtensionValidator.java
+++ b/annotation/compiler/src/main/java/com/bumptech/glide/annotation/compiler/GlideExtensionValidator.java
@@ -1,11 +1,13 @@
 package com.bumptech.glide.annotation.compiler;
 
 import static com.bumptech.glide.annotation.compiler.ProcessorUtil.nonNull;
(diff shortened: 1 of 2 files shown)


mariotaku:
@googlebot I signed it!

sjudd:
This seems reasonable to me, thanks for putting this up! I'll have the build breakage fixed shortly. If you're interested in making the change I suggested let me know.

sjudd:
Merged here: a1f5f1008d4e9588ad37acd731d73bd63903440c Thanks!

PR #2936, opened 2018-03-06
Updated documentation

PaulWoitaschek:
As into(int,int) is depreacted too, suggest to use submit (which the method itself does too)


sjudd:
Thanks!

PR #2970, opened 2018-03-21
Give generated code meaningful parameter names

SUPERCILEX:
The currently generated method parameters are meaningless `arg0`, `arg1` kind of stuff which makes for an unpleasant development experience. This PR makes the params as pretty as possible given the limitation that we don't know what the actual parameter names are.

Now I'm not gonna lie, the code to do this is kind of ugly, but I think the end result is well worth it.


SUPERCILEX:
Bump, waiting for review

sjudd:
Sorry for the delay! I agree that the solution isn't ideal, but it's more reasonable to implement than adding yet more annotations and I think the additional complication in the (already complicated) annotation processor is probably worth it given that a lot more people will use the output than read the code that produces it. 

Thanks!

SUPERCILEX:
No problem! I just wish we could get the actual parameter names instead of `arg0`, `arg1`, etc... 😭 Oh well! 😊

PR #2996, opened 2018-03-29
Fix don't start preload on open RecyclerView (without user intract)

Anton111111:
I see that ListPreloader don't start preload until the user don't start scroll. I think my commit fix it.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/ListPreloader.java
+++ b/library/src/main/java/com/bumptech/glide/ListPreloader.java
@@ -33,7 +33,7 @@ public class ListPreloader<T> implements AbsListView.OnScrollListener {
 
   private int lastEnd;
   private int lastStart;
-  private int lastFirstVisible;
+  private int lastFirstVisible = -1;
   private int totalItemCount;
 
   private boolean isIncreasing = true;


Anton111111:
I signed it!

sjudd:
Thanks for the pull request, but I don't believe preload is actually called until onScroll: https://github.com/Anton111111/glide/blob/c6b8c22fabd04c77c654c9dd6732ffb4f745e14d/library/src/main/java/com/bumptech/glide/ListPreloader.java#L142.

Anton111111:
@sjudd , may be i do something wrong. But i try to show you how fix work.
I add small changes to Gallery sample to demonstrate problem.
I've add TextView to show log of preloading. You can see it on branch "demo_preload_problem" and try it himself (simple start gallery sample from my branch)
https://github.com/Anton111111/glide/blob/demo_preload_problem/samples/gallery/src/main/java/com/bumptech/glide/samples/gallery/MainActivity.java

I've removed my fix from ListPreloader.java and add log to onScroll method:
https://github.com/Anton111111/glide/blob/demo_preload_problem/library/src/main/java/com/bumptech/glide/ListPreloader.java
On start preload i write to log "Do preload from...". 

I uploaded small video to youtube:
https://www.youtube.com/watch?v=rvkMHuh3-nE&feature=youtu.be

How you see "Do preload from" uppears in log only after user scroll. It happen because firstVisible and visibleCount variables equal 0 on first call onScroll (before user start scroll).

And i upload video with fix. You can see that preload start without user scroll. 
https://www.youtube.com/watch?v=0NKZe3w_1IQ
To check it himself simple init lastFirstVisible with -1 (how in my pull request)

sjudd:
Thanks for the clarification and detail! This works because onScrolled is called during layout when there's content.

PR #2999, opened 2018-03-30
Respect theme when loading Drawables while AppCompat is present

### LINKED ISSUE #2783: DrawableDecoderCompat#getDrawable(Context, int, Theme) does not respect theme when AppCompat v7 is being used
**Glide Version**: 4.4.0
**Integration libraries**: None
**Device/Android Version**: OnePlus 5 (cheeseburger), Android 8.0.0 (26)
**Issue details / Repro steps / Use case background**: I'm trying to use a night mode dependent drawable's resource id for Glide's error drawable option, hence it is necessary for the theme passed into RequestOptions to be used when the actual drawable gets loaded in case of the original load failing. However, the provided theme isn't being respected by Glide's error drawable load handling and Glide therefore always loads the notnight configuration's resource

**Glide load line**:
```kotlin
fun <T> requestArtwork(context: Context, resourceClass: Class<T>): RequestBuilder<T> {
    return Glide.with(context)                                                        
            .`as`(resourceClass)                                                      
            .load(getArtworkModel())                                                  
}                                                                                     
                                                                                      
@UiThread                                                                             
fun loadArtworkInto(target: ImageView) {                                              
    try {                                                                             
        requestArtwork(target.context, Bitmap::class.java)                            
                .apply(RequestOptions()                                               
                        .theme(target.context.theme)                                  
                        .error(defaultArtworkRes)                                     
                        .centerCrop())                                                
                .transition(BitmapTransitionOptions()                                 
                        .crossFade())                                                 
                .into(target)                                                         
    } catch (e: Exception) {                                                          
        e.printStackTrace()                                                           
    }                                                                                 
}                                                                                     
```

The critical aspect here is DrawableDecoderCompat's getDrawable method, which Glide uses internally to get the drawable. If the support v7 library AppCompat is present (which obviously is the case), it will just use AppCompatResources's getDrawable method, which however doesn't involve a theme parameter, thus the passed theme is dropped:
```java
/**                                                                                              
 * Loads a Drawable using {@link AppCompatResources} if available and {@link ResourcesCompat}    
 * otherwise, depending on whether or not the v7 support library is included in the application. 
 *                                                                                               
 * @param theme Used instead of the {@link Theme} returned from the given {@link Context} if     
 * non-null when loading the {@link Drawable}.                                                   
 */                                                                                              
public static Drawable getDrawable(Context context, @DrawableRes int id, @Nullable Theme theme) {
  try {                                                                                          
    // Race conditions may cause us to attempt to load using v7 more than once. That's ok since  
    // this check is a modest optimization and the output will be correct anyway.                
    if (shouldCallAppCompatResources) {                                                          
      return loadDrawableV7(context, id);                                                        
    }                                                                                            
  } catch (NoClassDefFoundError error) {                                                         
    shouldCallAppCompatResources = false;                                                        
  } catch (Resources.NotFoundException e) {                                                      
    // Ignored, this can be thrown when drawable compat attempts to decode a canary resource. If 
    // that decode attempt fails, we still want to try with the v4 ResourcesCompat below.        
  }                                                                                              
                                                                                                 
  return loadDrawableV4(context, id, theme != null ? theme : context.getTheme());                
}                                                                                                
                                                                                                 
private static Drawable loadDrawableV7(Context context, @DrawableRes int id) {                   
  return AppCompatResources.getDrawable(context, id);                                            
}                                                                                                
                                                                                                 
private static Drawable loadDrawableV4(                                                          
    Context context, @DrawableRes int id, @Nullable Theme theme) {                               
  Resources resources = context.getResources();                                                  
  return ResourcesCompat.getDrawable(resources, id, theme);                                      
}                                                                                                
```

julianostarek:
## Description
Makes DrawableDecoderCompat#loadDrawableV7 wrap a now passed Resources.Theme with the present Context, in order for the former to be respected as AppCompat tries to load the Drawable resource

## Motivation and Context
Resolves #2783

### DIFF
--- a/library/src/main/java/com/bumptech/glide/load/resource/drawable/DrawableDecoderCompat.java
+++ b/library/src/main/java/com/bumptech/glide/load/resource/drawable/DrawableDecoderCompat.java
@@ -8,6 +8,7 @@ import android.support.annotation.DrawableRes;
 import android.support.annotation.Nullable;
 import android.support.v4.content.res.ResourcesCompat;
 import android.support.v7.content.res.AppCompatResources;
+import android.support.v7.view.ContextThemeWrapper;
 
 /**
  * Handles decoding Drawables with the v7 support library if present and falling back to the v4
@@ -38,7 +39,7 @@ public final class DrawableDecoderCompat {
       // Race conditions may cause us to attempt to load using v7 more than once. That's ok since
       // this check is a modest optimization and the output will be correct anyway.
       if (shouldCallAppCompatResources) {
-        return loadDrawableV7(context, id);
+        return loadDrawableV7(context, id, theme);
       }
     } catch (NoClassDefFoundError error) {
       shouldCallAppCompatResources = false;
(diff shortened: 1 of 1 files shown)


julianostarek:
Signed the CLA

sjudd:
Thanks!

PR #3002, opened 2018-04-01
Fix swapped pixels in GifDecoder's output

mtopolnik:
`StandardGifDecoder` uses modified [public domain](http://show.docjava.com/book/cgij/exportToHTML/ip/gif/stills/GifDecoder.java.html) code to decode the GIF. Some changes were done to it that introduced a subtle bug: the initial two pixels are swapped, as well as the pixels after each `CLEAR` command in the LZW stream.

Comparing with the original code, the bug was introduced because the loop that pops the bytes from the stack into the output pixel array was moved. This code motion did not take into account the [`continue`](https://github.com/bumptech/glide/blob/master/third_party/gif_decoder/src/main/java/com/bumptech/glide/gifdecoder/StandardGifDecoder.java#L773) in the branch guarded by `oldCode == NULL_CODE`. The pixel that this branch pushes to the stack was supposed to be immediately transferred to the output, but since the stack-popping loop was moved, it happens after adding `first` to the output pixels, thereby swapping the two initial pixels.

The fix is simple: instead of pushing to the stack in the `oldCode == NULL_CODE` branch, send the pixel directly to the output pixel array.

This PR includes a fix, a small test GIF with 8 white-black-white-black pixels, and a JUnit test that fails without the fix.


mtopolnik:
I signed the CLA just now.

mtopolnik:
Thank you for your attention, `googlebot` and `stalebot`.

sjudd:
Thank you! Sorry for the delay. Nice find.

PR #3016, opened 2018-04-11
fix bug #2894

ooeoo:

### DIFF
--- a/library/src/main/java/com/bumptech/glide/load/model/StringLoader.java
+++ b/library/src/main/java/com/bumptech/glide/load/model/StringLoader.java
@@ -34,7 +34,7 @@ public class StringLoader<Data> implements ModelLoader<String, Data> {
 
   @Override
   public boolean handles(@NonNull String model) {
-    return true;
+    return uriLoader.handles(parseUri(model));
   }
 
   @Nullable


ooeoo:
I signed it!

ooeoo:
I signed it!

iBotasky:
@ooeoo Did this bug fix?
What reason course this happen?

iBotasky:
OK

sjudd:
I believe the underlying cause is actually that loader that we intend to load the url isn't attempted. This will just make Glide find one less loader, it won't make Glide find the right loader.

ooeoo:
The correct loader was executed, but an error occurred, causing the next loader to be executed, but it should not be executed.

sjudd:
The concerning cases in the referenced issue are the ones where the http loader was not executed. 

If you just want to reduce some of the guaranteed to fail load attempts you can, but it doesn't fix the referenced issue. 

If you do want the behavior you've added, then move the check for ``handles()`` into ``buildLoadData`` and return null if the wrapped loader returns `false`. That way we only  have to parse the Uri once.

ooeoo:
You are right, it just reduce some ModelLoader.

kozaxinan:
Do we know the core reason for this issue? Or does someone have a plan for fix?

PR #3191, opened 2018-07-09
Bump snapshot version to 4.8.0.

Muyangmin:
## Description

Since we have released 4.7.1 and 4.8.0-SNAPSHOT is actually available, I think this documentation has to be updated.
## Motivation and Context

Only changed version number.


sjudd:
Thanks!

PR #3308, opened 2018-09-10
Support loading resources from dynamic modules

SUPERCILEX:
### Background

Dynamic modules use split APKs to distribute resources with the module instead of the base APK. This means that, for example, the runtime package may be `com.supercilex.robotscouter` while the resource is located in `com.supercilex.robotscouter.teams`.

These differences cause problems when Glide tries to load the resource because Glide will try to create a context from the `com.supercilex.robotscouter.teams` package which doesn't really exist.

### Proposal

If resources are a subset of the context's package, use said context. Otherwise, try to create a new context.

@sjudd Can you think of a better solution? This is the best one I've come up with so far, but it seems hacky.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/load/resource/drawable/ResourceDrawableDecoder.java
+++ b/library/src/main/java/com/bumptech/glide/load/resource/drawable/ResourceDrawableDecoder.java
@@ -47,7 +47,7 @@ public class ResourceDrawableDecoder implements ResourceDecoder<Uri, Drawable> {
       @NonNull Options options) {
     @DrawableRes int resId = loadResourceIdFromUri(source);
     String packageName = source.getAuthority();
-    Context targetContext = packageName.equals(context.getPackageName())
+    Context targetContext = packageName.contains(context.getPackageName())
         ? context : getContextForPackage(source, packageName);
     // We can't get a theme from another application.
     Drawable drawable = DrawableDecoderCompat.getDrawable(context, targetContext, resId);
--- a/library/src/main/java/com/bumptech/glide/load/resource/drawable/ResourceDrawableDecoder.java
+++ b/library/src/main/java/com/bumptech/glide/load/resource/drawable/ResourceDrawableDecoder.java
@@ -47,14 +47,14 @@ public class ResourceDrawableDecoder implements ResourceDecoder<Uri, Drawable> {
       @NonNull Options options) {
     @DrawableRes int resId = loadResourceIdFromUri(source);
     String packageName = source.getAuthority();
-    Context targetContext = getContextForPackage(source, packageName);
+    Context targetContext = findContextForPackage(source, packageName);
     // We can't get a theme from another application.
     Drawable drawable = DrawableDecoderCompat.getDrawable(context, targetContext, resId);
     return NonOwnedDrawableResource.newInstance(drawable);
   }
 
   @NonNull
-  private Context getContextForPackage(Uri source, String packageName) {
+  private Context findContextForPackage(Uri source, String packageName) {
     // Fast path
     if (packageName.equals(context.getPackageName())) {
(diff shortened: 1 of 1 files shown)


SUPERCILEX:
Yeah, that's the issue 😞. Since it's just a package defined in the manifest, I'm pretty sure you can have it be whatever you want (though the official examples do [follow this pattern](https://developer.android.com/topic/google-play-instant/getting-started/create-base-feature-module#assoc-apk-feature)).

Sidenote: why are we throwing in the first place? Could we just return the original context if the new one isn't valid?

SUPERCILEX:
@sjudd should we just not throw? It'll fail later on when it tries to load the resource anyway.

sjudd:
What's the harm in throwing? Does it make a difference if the load's going to fail?

SUPERCILEX:
Well the problem is that we throw too early since we don't know what the correct package for the resource is. I'm saying that if creating a new context fails, we just return the original context instead of throwing because loading the resource from that context may or may not fail.

SUPERCILEX:
@sjudd Do you think this PR is good as is? Or should I not throw as discussed above?

SUPERCILEX:
@sjudd bump

library/src/main/java/com/bumptech/glide/load/resource/drawable/ResourceDrawableDecoder.java
```diff
@@ -47,18 +47,27 @@ public boolean handles(@NonNull Uri source, @NonNull Options options) {
...
-    Context targetContext = packageName.equals(context.getPackageName())
-        ? context : getContextForPackage(source, packageName);
+    Context targetContext = getContextForPackage(source, packageName);
     // We can't get a theme from another application.
     Drawable drawable = DrawableDecoderCompat.getDrawable(context, targetContext, resId);
     return NonOwnedDrawableResource.newInstance(drawable);
   }
 
   @NonNull
   private Context getContextForPackage(Uri source, String packageName) {
```
sjudd: Probably rename to findContext since you're no longer just returning the context for the given package name.

library/src/main/java/com/bumptech/glide/load/resource/drawable/ResourceDrawableDecoder.java
```diff
@@ -47,18 +47,27 @@ public boolean handles(@NonNull Uri source, @NonNull Options options) {
...
+    // Fast path
+    if (packageName.equals(context.getPackageName())) {
+      return context;
+    }
+
     try {
       return context.createPackageContext(packageName, /*flags=*/ 0);
     } catch (NameNotFoundException e) {
+      // The parent APK holds the correct context if the resource is located in a split
+      if (packageName.contains(context.getPackageName())) {
```
sjudd: I'm not sure I understand what this does, isn't this the same as the check above the try?

sjudd: Oh! Contains. Of course, disregard.

sjudd:
I'd rather keep the exception for now I think. It's probably better to clearly fail when the resource can't be found, even if we might have found it, than silently return the wrong image if the resource id happens to correspond with a different image in the containing package.

library/src/main/java/com/bumptech/glide/load/resource/drawable/ResourceDrawableDecoder.java
SUPERCILEX: Done

SUPERCILEX:
Ohhh, I didn't think of that latter case, so SGTM for still throwing. 👍 Is this PR ready for purple then?

sjudd:
LGTM, Thanks!

PR #3321, opened 2018-09-15
fix: replace "GlideApp" with "Glide"

hansemannn:
## Description

This tiny PR fixes the readme example of using Glide. It looks like previous versions used the "GlideApp" instance, whereby the current version uses "Glide" and works fine on my end.


Tolriq:
https://github.com/bumptech/glide/pull/2913 :)

hansemannn:
Ahh, thx!

PR #3326, opened 2018-09-22
Update README.md

5peak2me:
change clone error

## Description

if someone wants to clone repository and build jar for himself, it can't do it.just because of git submodules
## Motivation and Context

make README more clearly, do it follow the lead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ Build
 Building Glide with gradle is fairly straight forward:
 
 ```shell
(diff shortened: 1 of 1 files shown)


5peak2me:
@googlebot I signed it!

sjudd:
Submodules are initialized and updated by gradle build, is that not working? I'm not necessarily opposed to this, but wouldn't want to add it if it's not necessary.

Issue #2875, opened 2018-02-05
API suggestion - RequestOptions


MFlisar:
**Glide Version**: 4.5

**Integration libraries**: -

**Device/Android Version**: LL

**Issue details / Repro steps / Use case background**: 

Current `RequestOptions` differentiate between one and multiple transformations. Why? I see that internally the transformations are handled by a `MultiTransformation` if multiple transformations are used, but this could be totally hidden from the user I think. Additional, currently a unnecessary `MultiTransformation` object is created if only one transformation is passed to the `transforms` function

**Suggestion**

Why not change the API to look like following (combine `transform` and `transforms`):

	public RequestOptions transform(@NonNull Transformation<Bitmap>... transformations) {
		if (transformations.lengh > 1) {
			return transform(new MultiTransformation<>(transformations), /*isRequired=*/ true);
		} else if (transformations.lengh == 1) {
			return transform(transformation, /*isRequired=*/ true);
		} else {
			return this;
		}
	}

**Usage before**

This way as a user using an array (potentially empty) I could change following: 

    RequestOptions ro = new RequestOptions();
	if (transformations != null && transformations.size() > 0) {
		if (transformations .size() == 1) {
			ro = ro.transform(transformations .get(0));
		} else {
			ro = ro.transforms(transformations .toArray(new Transformation[transformations .size()]));
		}
	}
	
**Usage after**

    RequestOptions ro = new RequestOptions();
	if (transformations != null) {
		ro = ro.transform(transformations .toArray(new Transformation[transformations .size()]));
	}

sjudd:
Using varargs allocates an extra array, but you can solve that with an additional overload. 

If you want to send a pull request it should:

1. Deprecate ``transforms``
2. Add a varargs version of ``transform``

You won't need to do any bounds checking because the single argument method overload will be used if only one item is passed in anyway.

TacoTheDank:
I think this issue was meant to be closed with the merge...

PR #3365, opened 2018-10-20
Updated BaseRequestOptions.transform api (update for issue #2875)

aleien:
## Description

Added varargs overload method for BaseRequestOptions.transform.
Deprecated BaseRequestOptions.transforms

## Motivation and Context
This is fix for issue #2875, where user requested a little simplification to api.

While building locally, I kept failing build because of lint check in volley. It insists on having volley latest version (1.1.0, while there is 1.0.0 in project), so I am not sure if CI will be successfull too.


sjudd:
Thanks!

Issue #1161, opened 2016-04-27
java.lang.IllegalArgumentException: register too many Broadcast Receivers


francescocervone:
**Glide Version**: 3.6.1

**Device/Android Version**: Huawei RIO-L01 Android 5.1

**Issue details**: It seems to be caused by an issue in the Android framework that occurs only on Huawei RIO-L01. Anyway it could be solved catching the exception when `DefaultConnectivityMonitor` registers the receiver.

Actually I was considering to fix it myself providing a no-op `ConnectivityMonitor` only for the devices affected by this issue through `GlideBuilder#setConnectivityMonitorFactory`, but I have seen that this isn't possible in 3.x version.

**Stack trace / LogCat**:

``` ruby
Fatal Exception: java.lang.RuntimeException: Unable to start activity ComponentInfo{it.subito/it.subito.v2.detail.AdDetailActivity}: java.lang.IllegalArgumentException: regist too many Broadcast Receivers
       at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2450)
       at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2512)
       at android.app.ActivityThread.access$1200(ActivityThread.java:165)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1373)
       at android.os.Handler.dispatchMessage(Handler.java:102)
       at android.os.Looper.loop(Looper.java:135)
       at android.app.ActivityThread.main(ActivityThread.java:5593)
       at java.lang.reflect.Method.invoke(Method.java)
       at java.lang.reflect.Method.invoke(Method.java:372)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:960)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:755)
Caused by java.lang.IllegalArgumentException: regist too many Broadcast Receivers
       at android.app.ContextImpl.registerReceiverInternal(ContextImpl.java:1836)
       at android.app.ContextImpl.registerReceiver(ContextImpl.java:1794)
       at android.app.ContextImpl.registerReceiver(ContextImpl.java:1788)
       at android.content.ContextWrapper.registerReceiver(ContextWrapper.java:489)
       at com.bumptech.glide.manager.DefaultConnectivityMonitor.register(SourceFile:39)
       at com.bumptech.glide.manager.DefaultConnectivityMonitor.onStart(SourceFile:61)
[...]

PR #3368, opened 2018-10-23
Update library/src/main/java/com/bumptech/glide/manager/DefaultConnec…

zhaoxiangjun40:
…tivityMonitor.java

register too many Broadcast Receivers #1161
modify  receiver to static


zhaoxiangjun40:
I signed it!

sjudd:
This doesn't work. register() and unregister() are called in on onStart/onStop of every RequestManager, of which there's one per Fragment and Activity. With this change they'd all be sharing the same broadcast receiver and racing each other to register and unregister it.

Instead you'd need to consolidate the ConnectivityMonitor to a single instance and then do something like keep a count of all the starts and all the stops and only register when the count of started Activities/Fragments goes from 0->1 and only unregister when the count goes back to 0.

Issue #1872, opened 2017-04-20
Need Callbacks for when a GifDrawable with an animation finishes it's last loop


mattinger:
**Glide Version**: 3.8.0-SNAPSHOT

**Glide load line / `GlideModule` (if any) / list Adapter code (if any)**:
```java
glide.load(b).into(new GlideDrawableImageViewTarget(imageView, GlideDrawable.LOOP_INTRINSIC));
```

I have a series of 3 animated gifs that have to animate themselves in sequence.  There seems to be no current way with glide to know when a particular animation is done so that i can start the next one.

GifDrawable seems to be the place where this might have to be 

```java
    @TargetApi(Build.VERSION_CODES.HONEYCOMB)
    @Override
    public void onFrameReady(int frameIndex) {
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.HONEYCOMB && getCallback() == null) {
            stop();
            reset();
            return;
        }

        invalidateSelf();

        if (frameIndex == decoder.getFrameCount() - 1) {
            loopCount++;
        }

        if (maxLoopCount != LOOP_FOREVER && loopCount >= maxLoopCount) {
            stop();
        }
    }
```

i could override that method, however, i'd have no access to the decoder to see if there's any more frames available after the current one.

TWiStErRob:
How about overriding `stop()` and setting the `maxLoopCount` to `1`, so it runs once and then stops. In `stop` then you can fire the next one.

TWiStErRob:
Anyway, if you're up to it a PR would be welcome as well that adds this functionality.

mattinger:
I'm not sure stop() is necessarily the right way though.  It would trigger even if the animation was manually stopped.  I'm looking to trigger the next one only when it runs till completion.

mattinger:
So, i've managed to do this, however, what i've seen is that even with the .dontAnimate() flag set, i'm seeing the image being cleared out visually first, and then it animations.  This is an issue for me because i'm first loading the initial frame of my images:  

    Glide.with(imageView.getContext()).load(bytes).asBitmap().into(imageView);

Then i'm animating the first one, and when it stops, the next, and so forth:

    glide.load(b).dontAnimate().into(...)

I'm curious if there's a way for me to disable the clearing of the ImageView prior to starting the animation.

sjudd:
Each time you call into() into a View or Target, the previous load is cancelled and the View/Target is cleared. 

[...]

PR #3438, opened 2018-12-02
Add Gif loop end listener

AnwarShahriar:
## Description

This PR adds a listener to the `GifDrawable` to listen to gif loop end, when it's explicitly specified by the user.

Possibly solves #1872 

## Motivation and Context

I worked on a feature for showing and stopping a gif after certain loop count. The spec was,
* Show Gif in a RecyclerView
* Render it as a static picture (first frame of the Gif), but with an UI indicator that the picture is a gif
* When the user taps on the gif, hide the UI indicator and start playing it
* After 3 times of looping, stop the Gif and show the UI indicator again on top of the static first frame again.

To implement the above spec, (showing and hiding some UI stuff based on the Gif start/stop) I need to know when the Gif finishes looping 3 times. As glide doesn't expose decoder from GifDrawable and we cannot use anything other than GifDrawable in Request listener for gifs, I had to add a callback that can notify client when certain amount of looping finishes. This solution solves my issue and I wanted to share it with the community.

Please, let me know if it can be done in better way or any more work I need to do.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/load/resource/gif/GifDrawable.java
+++ b/library/src/main/java/com/bumptech/glide/load/resource/gif/GifDrawable.java
@@ -14,6 +14,7 @@ import android.graphics.drawable.Animatable;
 import android.graphics.drawable.Drawable;
 import android.support.annotation.NonNull;
 import android.support.annotation.VisibleForTesting;
+import android.support.graphics.drawable.Animatable2Compat;
 import android.view.Gravity;
 import com.bumptech.glide.Glide;
 import com.bumptech.glide.gifdecoder.GifDecoder;
@@ -21,12 +22,14 @@ import com.bumptech.glide.load.Transformation;
 import com.bumptech.glide.load.engine.bitmap_recycle.BitmapPool;
 import com.bumptech.glide.util.Preconditions;
 import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.List;
 
 /**
  * An animated {@link android.graphics.drawable.Drawable} that plays the frames of an animated GIF.
  */
 public class GifDrawable extends Drawable implements GifFrameLoader.FrameCallback,
-    Animatable {
+    Animatable, Animatable2Compat {
   /**
    * A constant indicating that an animated drawable should loop continuously.
    */
@@ -75,6 +78,11 @@ public class GifDrawable extends Drawable implements GifFrameLoader.FrameCallbac
   private boolean applyGravity;
   private Paint paint;
   private Rect destRect;
+
+  /**
+   * Callbacks to notify loop completion of a gif, where the loop count is explicitly specified.
+   */
+  private List<AnimationCallback> animationCallbacks;
 
   /**
    * Constructor for GifDrawable.
@@ -351,7 +359,16 @@ public class GifDrawable extends Drawable implements GifFrameLoader.FrameCallbac
     }
 
     if (maxLoopCount != LOOP_FOREVER && loopCount >= maxLoopCount) {
+      notifyAnimationEndToListeners();
       stop();
+    }
+  }
+
+  private void notifyAnimationEndToListeners() {
+    if (animationCallbacks != null) {
+      for (int i = 0, size = animationCallbacks.size(); i < size; i++) {
+        animationCallbacks.get(i).onAnimationEnd(this);
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


sjudd:
Good idea and thank you for putting this together. However it's probably best to stick to the Android interfaces. Can you change this to use Animatable2Compat? https://developer.android.com/reference/android/support/graphics/drawable/Animatable2Compat

AnwarShahriar:
@sjudd I'm a bit confused, and before putting more code I just want to make sure I'm understanding you correctly. Are you suggesting something like this?
```java
@Override
public boolean onResourceReady(GifDrawable gifDrawable, Object model, Target<GifDrawable> target, DataSource dataSource, boolean isFirstResource) {
  gifDrawable.setLoopCount(3);
  gifDrawable.setGifLoopEndedListener(new Animatable2Compat.AnimationCallback() {
    @Override
    public void onAnimationEnd(Drawable drawable) {
      // Animation is done. update the UI or whatever you want
    }
  });
  return false;
}
```
If this is our intention, shouldn't we also provide the `onAnimationStart` implementation? The unused interface may create confusion. What do you think?

sjudd:
@AnwarShahriar That's similar to what I'm suggesting. I'm suggesting making `GifDrawable` implement `Animatable2Compat`:

```java
class GifDrawable implements Animatable, Animatable2Compat {
...
```

Your listener would just be a [`Animatable2Compat.AnimationCallback`](https://developer.android.com/reference/android/support/graphics/drawable/Animatable2Compat.AnimationCallback.html). You'd add and remove with the register/unregister methods in `Animatable2Compat`:

```java
@Override
public boolean onResourceReady(GifDrawable gifDrawable, Object model, Target<GifDrawable> target, DataSource dataSource, boolean isFirstResource) {
  gifDrawable.setLoopCount(3);
  // Note that you'd need to unregister this in onLoadCleared to avoid a potential memory leak.
  gifDrawable.registerAnimationCallback(new Animatable2Compat.AnimationCallback() {
    @Override
    public void onAnimationEnd(Drawable drawable) {
      // Animation is done. update the UI or whatever you want
    }
  });
  return false;
}
```

Does that help clarify?

AnwarShahriar:
@sjudd updated the PR. Let me know if I need to modify anything.

library/src/main/java/com/bumptech/glide/load/resource/gif/GifDrawable.java
```diff
@@ -390,6 +407,50 @@ public void setLoopCount(int loopCount) {
...
+   * set by {@link GifDrawable#setLoopCount(int)}
+   *
+   * Note: Unregister this in onLoadCleared to avoid potential memory leak.
+   * @see GifDrawable#unregisterAnimationCallback(AnimationCallback).
+   *
+   * @param animationCallback Animation callback {@link Animatable2Compat.AnimationCallback}.
+   */
+  @Override
+  public void registerAnimationCallback(@NonNull AnimationCallback animationCallback) {
+    if (animationCallback == null) {
```
sjudd: I'd either skip this check or make it a Precondition. I think with the `@NonNull` annotation you're fine just skipping the check.

If you do want a check, I'd prefer a Precondition because it would immediately show the developer where they went wrong, where the existing check will just silently fail.

library/src/main/java/com/bumptech/glide/load/resource/gif/GifDrawable.java
```diff
@@ -390,6 +407,50 @@ public void setLoopCount(int loopCount) {
...
+
+  /**
+   * Unregister animation callback.
+   *
+   * @param animationCallback Animation callback {@link Animatable2Compat.AnimationCallback}.
+   * @return true when the animation callback is successfully removed, otherwise false.
+   */
+  @Override
+  public boolean unregisterAnimationCallback(@NonNull AnimationCallback animationCallback) {
+    if (animationCallbacks == null || animationCallback == null) {
```
sjudd: ditto here for animationCallback, I think you can skip the null check with the annotation (or make it a Precondition if you want one).

AnwarShahriar:
@sjudd fixed as you suggested, but I don't understand why few tests are failing, everything was passing before. Can you give me some hint?

AnwarShahriar:
@sjudd is there any update?

library/src/main/java/com/bumptech/glide/load/resource/gif/GifDrawable.java
```diff
@@ -390,6 +407,47 @@ public void setLoopCount(int loopCount) {
     }
   }
 
+  /**
+   * Register callback to listen to GifDrawable animation end event after specific loop count
+   * set by {@link GifDrawable#setLoopCount(int)}
```
sjudd: Worth noting that this will only be called if the Gif stops because it reaches the loop count. It doesn't look like you're notifying if someone calls stop() directly.

library/src/main/java/com/bumptech/glide/load/resource/gif/GifDrawable.java
```diff
@@ -390,6 +407,47 @@ public void setLoopCount(int loopCount) {
...
+  @Override
+  public void registerAnimationCallback(@NonNull AnimationCallback animationCallback) {
+    if (animationCallbacks == null) {
+      animationCallbacks = new ArrayList<>();
+    }
+    animationCallbacks.add(animationCallback);
+  }
+
+  /**
+   * Unregister animation callback.
```
sjudd: You can probably avoid javadoc for these methods that isn't specific to their behavior in this particular class. The javadoc for the interface will be shown by default.

sjudd:
Thanks!

pietroconta:
> ```java
>  gifDrawable.setGifLoopEndedListener(new Animatable2Compat.AnimationCallback() {
>     @Override
>     public void onAnimationEnd(Drawable drawable) {
>       // Animation is done. update the UI or whatever you want
>     }
>   });
> ```

i dont know why i have this error: Cannot resolve method 'setGifLoopEndedListener' in 'GifDrawable

PR #3444, opened 2018-12-05
Add BitmapTransformation to rotate the image

Gnzlt:
## Description
I've added a `BitmapTransformation` to rotate the image.

## Motivation and Context
I found myself creating the same `BitmapTransformation` class for rotation in multiple projects when this could be implemented in the library itself.
Glide already provides `TransformationUtils.rotateImage()` method so it seems logical to have its own `BitmapTransformation`.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/load/resource/bitmap/Rotate.java
+++ b/library/src/main/java/com/bumptech/glide/load/resource/bitmap/Rotate.java
@@ -1,0 +1,45 @@
+package com.bumptech.glide.load.resource.bitmap;
+
+import android.graphics.Bitmap;
+import android.support.annotation.NonNull;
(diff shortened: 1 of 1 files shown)


sjudd:
Out of curiosity what's the use case where you need to rotate the image? The user manually wants to change the orientation?

Gnzlt:
When using [Fotoapparat](https://github.com/RedApparat/Fotoapparat) lib, you need to rotate the bitmap based on the device sensor orientation.

sjudd:
Ok by me, this is mostly just wrapping existing code.

Ideally camera apps should be writing image formats with orientations specified correctly? I'm not sure if that particular library is special or not. Glide will already parse exif and rotate images according to the exif. 

Thanks!

PR #3446, opened 2018-12-07
Add nullability annotations to LazyHeaders.Builder functions

r-ralph:
## Description
Add `@NonNull` and `@Nullable` annotations to `LazyHeaders.Builder` and `StringHeaderFactory`. When there is a possibility of null, a warning is displayed on the IDE and bug entry can be prevented.

## Motivation and Context

`StringHeaderFactory` doesn't allow null `value` and `LazyHeaders.Builder.addHeader(String key, String value)` also doesn't allow null `value`. However, the user can not know this spec unless they actually read the code.


sjudd:
Thanks!

PR #3468, opened 2018-12-30
Improve load RequestManager clear performance

SUPERCILEX:
This PR is mostly just a hunch. 😕 Basically, I noticed that way too much time is spent removing items in my flame chart. My guess is that using a hashmap will yield better performance but I'm not sure how to benchmark this.

image

### DIFF
--- a/library/src/main/java/com/bumptech/glide/request/target/CustomViewTarget.java
+++ b/library/src/main/java/com/bumptech/glide/request/target/CustomViewTarget.java
@@ -21,7 +21,9 @@ import com.bumptech.glide.util.Preconditions;
 import com.bumptech.glide.util.Synthetic;
 import java.lang.ref.WeakReference;
 import java.util.ArrayList;
-import java.util.List;
+import java.util.Collections;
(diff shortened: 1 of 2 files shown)


library/src/main/java/com/bumptech/glide/request/target/CustomViewTarget.java
```diff
@@ -300,7 +302,8 @@ private void maybeRemoveAttachStateListener() {
     @Nullable
     static Integer maxDisplayLength;
     private final View view;
-    private final List<SizeReadyCallback> cbs = new ArrayList<>();
+    private final Set<SizeReadyCallback> cbs =
```
sjudd: There's typically only one callback (unless you're using lots of thumbnails) and at most there's one callback per request you pass in. It's unlikely that that Set will perform any better than list and it allocates at least an extra object per Entry that's avoided with a list.

Do you see significant gains with these changes?

SUPERCILEX:
Huh, I debugged it and the list is empty 😕. It's probably a profiler issue then so I'll close the PR.

Issue #3474, opened 2019-01-06
how to custom executor in v4


jokermonn:
how to custom executor in glide v4? It looks like we can only use [GlideExecutor](https://github.com/bumptech/glide/blob/master/library/src/main/java/com/bumptech/glide/load/engine/GlideException.java)?

PR #3480, opened 2019-01-09
replace 'GlideExecutor' to 'ExecutorService'

jokermonn:
## Description

#3474 

## Motivation and Context

In general, libraries should allow developers to customize `ExecutorService`, and [v3](https://github.com/bumptech/glide/blob/3.0/library/src/main/java/com/bumptech/glide/GlideBuilder.java#L29) does


jokermonn:
I signed it!

sjudd:
This doesn't build. You'd also be removing functionality in an API breaking way. We might consider doing this in a major version change, but probably not until then.

PR #3482, opened 2019-01-10
Add pipe in safe char

### LINKED ISSUE #3479: Unable to load Google Map images
Hi,

I recently switched to Glide but some images fail to load. The url is something like this https://maps.googleapis.com/maps/api/staticmap?markers=color:red|1.265152455,103.8226872

If you notice, there is a pipe in the URL. But when GlideUrl convert the URL, since `|` is not in the ALLOWED_URI_CHARS it will be concert to %7C and then the request will fail.

Is it possible to add the | in the ALLOWED_URI_CHARS ? Or is there any way I can work around this on my side ?

lucas34:
Resolve #3479

### DIFF
--- a/library/src/main/java/com/bumptech/glide/load/model/GlideUrl.java
+++ b/library/src/main/java/com/bumptech/glide/load/model/GlideUrl.java
@@ -27,7 +27,7 @@ import java.util.Map;
  * convenience. </p>
  */
 public class GlideUrl implements Key {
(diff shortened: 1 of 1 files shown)


lucas34:
I signed it!

sjudd:
It doesn't look like pipe is in the allowed character set for a url: https://tools.ietf.org/html/rfc3986#section-2

PR #3508, opened 2019-01-30
Add support for incremental annotation processing

### LINKED ISSUE #2983: Support incremental annotation processing
Gradle 4.7 provides support for incremental annotation processing.
https://docs.gradle.org/nightly/userguide/java_plugin.html#sec:incremental_annotation_processing
Gradle feature discussion: gradle/gradle#1320

It would be nice to see glide's annotation processor supporting it. 

I started to have a look at the code and I was surprised to find such a complex annotation processing. I don't think I can do the PR quickly by myself. But I would be happy to provide all the information I can to help a maintainer doing it.

technoir42:
I've had to change the retention of `GlideModule` and `GlideExtension` to `CLASS` because [aggregating](https://docs.gradle.org/current/userguide/java_plugin.html#compilejava) annotation processors support only `CLASS` and `RUNTIME` retention.

Resolves #2983 
cc @stephanenicolas

### DIFF
--- a/annotation/compiler/src/main/resources/META-INF/gradle/incremental.annotation.processors
+++ b/annotation/compiler/src/main/resources/META-INF/gradle/incremental.annotation.processors
@@ -1,0 +1,1 @@
+com.bumptech.glide.annotation.compiler.GlideAnnotationProcessor,aggregating
--- a/annotation/src/main/java/com/bumptech/glide/annotation/GlideExtension.java
+++ b/annotation/src/main/java/com/bumptech/glide/annotation/GlideExtension.java
@@ -11,5 +11,5 @@ import java.lang.annotation.Target;
  * @see GlideOption
  */
 @Target(ElementType.TYPE)
-@Retention(RetentionPolicy.SOURCE)
+@Retention(RetentionPolicy.CLASS)
 public @interface GlideExtension { }
--- a/annotation/src/main/java/com/bumptech/glide/annotation/GlideModule.java
+++ b/annotation/src/main/java/com/bumptech/glide/annotation/GlideModule.java
@@ -12,7 +12,7 @@ import java.lang.annotation.Target;
  * <p>Replaces <meta-data /> tags in AndroidManifest.xml.
  */
 @Target(ElementType.TYPE)
-@Retention(RetentionPolicy.SOURCE)
+@Retention(RetentionPolicy.CLASS)
 public @interface GlideModule {
   /**
    * Returns the name of the class that will be used as a replacement for


bingranl:
Thanks for your PR, Sergey! I agree with you for changing the retention policy and make it aggregating. But I am wondering why are you adding originating element for those generated classes while you already make glide an aggregating annotation processor. As far as I know, aggregating AP doesn't need that and we are not able to know all the originating elements which is why we make it aggregating.

sjudd:
Thanks! I'm working on the build failure, it seems to be an issue with travis setup.

I'm also curious about the originating elements. Isolating annotation processor documentation mentions originating elements: https://docs.gradle.org/current/userguide/java_plugin.html#example_an_isolated_annotation_processor. The aggregating documentation doesn't seem to. I've also found: https://github.com/gradle/gradle/issues/4130

sjudd:
Build error should be fixed if you rebase: b56dbabde3abeb5ed1c232a3f6b17a4a0af0b7e2

technoir42:
I was not sure whether originating elements were required for aggregating AP so I added them anyway. Now, looking at the https://github.com/gradle/gradle/blob/7e2af8a93d4e8f21ff044cc971492717cc14ed7f/subprojects/language-java/src/main/java/org/gradle/api/internal/tasks/compile/processing/AggregatingProcessingStrategy.java#L67 I see that Gradle simply ignores them in case of an aggregating AP.

I've reverted everything related to originating elements and tested again, it seems that it works exactly the same. We can add them later if Gradle ever starts requiring them.

technoir42:
Not sure what's the reason for this failure:

```
com.bumptech.glide.load.data.HttpUrlFetcherServerTest > testReturnsInputStreamOnStatusOk FAILED
    Wanted but not invoked:
    callback.onDataReady(<Capturing argument>);
    -> at com.bumptech.glide.load.data.HttpUrlFetcherServerTest.testReturnsInputStreamOnStatusOk(HttpUrlFetcherServerTest.java:73)
    However, there were other interactions with this mock:
    -> at com.bumptech.glide.load.data.HttpUrlFetcher.loadData(HttpUrlFetcher.java:65)
        at com.bumptech.glide.load.data.HttpUrlFetcherServerTest.testReturnsInputStreamOnStatusOk(HttpUrlFetcherServerTest.java:73)
```

I don't see it when running `./gradlew clean build --parallel` locally.

bingranl:
I can't reproduce the error locally either. Could that be a flaky test? How about we run it again?

sjudd:
I've restarted the job, I believe that is a flake.

Otherwise this looks good to me, I'll merge when I confirm the test passes. Thanks!

bingranl:
Great! Thanks!

Issue #3418, opened 2018-11-20
Proposal: Expirable LRU Disk Cache


dmytrodanylyk:
**Glide Version**: 4.8.0

With upcoming [GDPR](https://en.wikipedia.org/wiki/General_Data_Protection_Regulation) it would be nice to have some disk cache expiration policy. E.g. I want to cache image for 30 days.

This can be achieved by modifying `DiskLruCacheWrapper#get` method.

```java
@Override
public File get(Key key) {
    // ...
    if(expired)  {
        result = null;
        delete(key);
    }
    // ...
}
```

If you think this may be useful I can send PR.

dmytrodanylyk:
@sjudd @SUPERCILEXyz thoughts on this?

sjudd:
Are you ok with leaving "expired" files sitting on disk until the next time someone happens to request them? 

If not, you probably need a periodic clean up task and some accounting of when the file expires. You might be able to write that into the DiskLruCache implementation Glide uses, but you'd probably  need to modify Glide's DiskCache interface to make this coherent, which is tricky.

dmytrodanylyk:
> Are you ok with leaving "expired" files sitting on disk until the next time someone happens to request them?

@sjudd  I think that's ok for now. At least that's how I understand GDPR. Expired data must not be accessible.

TWiStErRob:
If a Glide user relocates the cache to `/sdcard` then the file is accessible, so it doesn't cover all cases.

dmytrodanylyk:
Yes sure, it doesn't cover all the edge cases. And I don't think it's possible to do that. At least with my proposal Glide will have the expirable caching mechanism out of the box.

PR #3514, opened 2019-02-11
Implemented Disk LRU Cache with entity expiration capabilities

dmytrodanylyk:
## Description

Implementation for [Proposal: Expirable LRU Disk Cache #3418](https://github.com/bumptech/glide/issues/3418).

Basically created `ExpirableDiskLruCacheWrapper` and some factories. `ExpirableDiskLruCacheWrapper` extends `DiskLruCacheWrapper` and overwrites `get` method which removes entity if it has expired.

Moved somme common code into `InternalCacheDirectoryGetter` and `ExternalCacheDirectoryGetter`.

Usage from client prospective:

```java
@GlideModule
public class YourAppGlideModule extends AppGlideModule {
  @Override
    long expirationMillis = TimeUnit.DAYS.toMillis(30);
    builder.setDiskCache(new ExternalExpirableCacheDiskCacheFactory(context, expiration));
  }
}
```

## Motivation and Context

With upcoming [GDPR](https://en.wikipedia.org/wiki/General_Data_Protection_Regulation) it would be nice to have some disk cache expiration policy. E.g. I want to cache image for 30 days.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/load/engine/cache/CacheDirectoryGetter.java
+++ b/library/src/main/java/com/bumptech/glide/load/engine/cache/CacheDirectoryGetter.java
@@ -1,0 +1,10 @@
+package com.bumptech.glide.load.engine.cache;
+
+import java.io.File;
+
+/**
+ * Interface called out of UI thread to get the cache folder.
+ */
+public interface CacheDirectoryGetter {
+  File getCacheDirectory();
+}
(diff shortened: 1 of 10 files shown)


dmytrodanylyk:
I am the only author, contributing on behalf of Atlassian. Any idea how to confirm that?

sjudd:
I think we can't accept this without making sure that expired files are deleted relatively promptly. I'm not sure we'd make this the default disk cache even if that feature were implemented. Have you considered hosting this separately from Glide? We could certainly point to it as an option in documentation for users who want expiration and are ok with the edge cases here.

dmytrodanylyk:
I am ok to host that separately but would prefer to have it in Glide, since we use similar code in our application and it would be nice to have it in Glide.

> deleted relatively promptly

Why do we need to do that? With the current implementation, if the expired file is accessed - it will be removed, if it's not accessed - it will be eventually removed by LRU mechanism or android system cleanup. 

The important piece of functionality is not removing expired files, but making expired files **not accessible**.

Issue #3184, opened 2018-07-05
Glide 4.8.0-SNAPSHOT doesn't work with AndroidX


Sander0542:
I have downloaded Glide 4.8.0-SNAPSHOT, but when I try to compile my app I get this error. This was after I upgraded to AndroidX

```
    [javac] /home/travis/build/Sander0542/Makeroid/appinventor/components/src/com/google/appinventor/components/runtime/util/MakeroidUtil.java:73: error: cannot access FragmentActivity
    [javac]             Glide.with(context).load(path).apply(new RequestOptions().override(Target.SIZE_ORIGINAL, Target.SIZE_ORIGINAL)).into(view);
    [javac]                  ^
    [javac]   class file for android.support.v4.app.FragmentActivity not found
```

gabrielfeo:
I'm having the same issue even when using 4.8.0-SNAPSHOT. I thought #3139 had fixed this already.

If I declare a `androidx.annotation:annotation:1.0.0-alpha3` dependency, or nothing at all, Glide gives errors because it can't find the annotation classes on the old namespace:

```
error: compiler message file broken: key=compiler.err.Processor: org.jetbrains.kotlin.kapt3.ProcessorWrapper@1f3f57ab arguments={0}, {1}, {2}, {3}, {4}, {5}, {6}, {7}
error: cannot access NonNull   class file for android.support.annotation.NonNull not found
error: compiler message file broken: key=compiler.err.Processor: org.jetbrains.kotlin.kapt3.ProcessorWrapper@1f3f57ab arguments={0}, {1}, {2}, {3}, {4}, {5}, {6}, {7}
error: cannot access Nullable   class file for android.support.annotation.Nullable not found
```


If I declare, however, a `com.android.support:support-annotations:28.0.0-alpha3` dependency, Glide gives errors for not finding the `FragmentActivity` class on the old namespace, as well as some `GlideOptions` issue:

```
error: compiler message file broken: key=compiler.err.Processor: org.jetbrains.kotlin.kapt3.ProcessorWrapper@4820a609 arguments={0}, {1}, {2}, {3}, {4}, {5}, {6}, {7}
error: cannot access FragmentActivity   class file for android.support.v4.app.FragmentActivity not found
[kapt] An exception occurred: java.lang.RuntimeException: javax.annotation.processing.FilerException: Attempt to recreate a file for type com.gabrielfeo.test.glide4.GlideOptions
```

dougnazar:
[...]

PR #3518, opened 2019-02-12
Use AndroidX instead of the support library

sjudd:


sjudd:
Fixes #3184.

However I can't submit this internally for the time being. I'm not going to block 4.9 on this.

TWiStErRob:
Nice one! Did a quick skim, does this mean the next 4.x release will break apps using non-androidx fragments, or a start of 5.x series?

CaiJingLong:
When will this PR merge into the main branch?
When will the AndroidX version be released?

DusilB:
I'm also interested in when this will reach the main branch, Will the AndroidX version of Glide be ready soon?

BladeF:
Since the stale bot closed this, when can we expect to see this PR merged?

TWiStErRob:
I don't think it was meant to be closed, but I don't know timeline.

mikehc:
Any updates on this issue? This is blocking my project from fully migrate to AndroidX.

sjudd:
Sorry for the delay. The internal migration at Google is progressing somewhat slower than hoped. My only other option is to use scripting to convert back and forth when code is imported/exported, which is non-trivial and I haven't yet had the chance to get to. 

I also need to change the tool I use to run migrations to copybara (released publicly). Fortunately it appears as though someone has already written the migrations to/from androidx internally, so I may be able to kill two birds with one stone. That's my current plan. 

I am curious as to how this is blocking anyone. I was under the impression that jetifier relieves most of these issues?

mikehc:
@sjudd in my case I'm using a custom glide module to register my own instance of OkHttpClient. The generated code used android support library imports. And I think in some classes they were using imports for both AndroidX and Support Library.

It's been a while since I tried making the migration so I don't have the code at hand.

wbervoets:
Also looking forward to having Glide depending on AndroidX. The Jetifier task does take build time, not ? Apart from play-services-location this is the only one still depending on the old support libraries in the app I'm working on.

Issue #3516, opened 2019-02-12
Unnecessary variable


javalong:
Version 4.8.0
com.bumptech.glide.GlideBuilder
has a  variable:
private GlideExecutor animationExecutor;

create a new GlideExecutor in build fuction:
Glide build(@NonNull Context context) {
...
 if (animationExecutor == null) {
      animationExecutor = GlideExecutor.newAnimationExecutor();
    }
...
}

but I can't find the usage.

I guess you want to use it in build fuction:
Glide build(@NonNull Context context) {
...

    if (engine == null) {
      engine =
          new Engine(
              memoryCache,
              diskCacheFactory,
              diskCacheExecutor,
              sourceExecutor,
              GlideExecutor.newUnlimitedSourceExecutor(),
              //  here is?
              GlideExecutor.newAnimationExecutor(),
              isActiveResourceRetentionAllowed);
    }
...
}

My English is poor ，sorry!!!

sjudd:
Feel free to send a pull request to remove it,thanks!

sjudd:
Fixed by 
#3525

PR #3525, opened 2019-02-18
fix the unused animationExecutor in GlideBuilder

phxnirvana:
## Description

#3516 

## Motivation and Context

fix the unused animationExecutor in GlideBuilder


sjudd:
Thanks!

PR #3537, opened 2019-02-21
* update okhttp library from 3.9.1 to 3.12.1

MichalStranik:
## Description
Crash fix for OkHttp issue: Unexpected TLS version: NONE

## Motivation and Context
This stack trace indicates a known issue with earlier versions of OkHttp, where failure to establish an SSL connection was not handled by the library, and instead bubbled up as an uncaught exception. Upgrading to an OkHttp version of 3.11 or greater should result in these failure modes to be correctly relayed to callbacks.


MichalStranik:
I signed it!

sjudd:
Thanks!

Issue #3582, opened 2019-03-25
CustomTarget constructor javadoc and implementation are inconsistent


JadenChoi:
**Glide Version**: 4.9.0

**Integration libraries**:

**Device/Android Version**: *

**Issue details / Repro steps / Use case background**: 
`CustomTarget` constructor javadoc says that width & height must be >= 0 or == Integer.MIN_VALUE.
However, the app crashes when width/height is 0 and throws exception with msg: `Width and height must both be > 0`. I looked at the implementation and we are enforcing that width & height to be > 0. Javadoc should also reflect it.

JadenChoi:
Created PR to update Javadoc https://github.com/bumptech/glide/issues/3582

eshaanb:
👍👍

PR #3583, opened 2019-03-25
CustomTarget javadoc is incorrect

JadenChoi:
`isValidDimensions` implementation enforces that width and height must be `> 0` or `== Target.SIZE_ORIGINAL`

Issue: https://github.com/bumptech/glide/issues/3582


sjudd:
Thanks!

PR #3783, opened 2019-07-26
Fix a wrong process

limuyang2:
It has been determined that `activity` belongs to `FragmentActivity` , so  should use the `get()` method of `FragmentActivity`


limuyang2:
I signed it!

PR #3821, opened 2019-08-08
Complete AndroidX support

arturogutierrez:
## Description
This PR removes all references to the old previous Android support libraries

## Motivation and Context
I was trying to remove Jetifier for my project (because I already migrated to AndroidX and also all third party libraries) but I couldn't because Glide was still having reference to `android.support.annotations`.
This PR removes all references to it so we don't need Jetifier to use Glide now.

I've already tested it publishing to MavenLocal and the new version is not having any dependency on Android Support.


arturogutierrez:
@googlebot I signed it!

sjudd:
I don't think any of these changes are actually necessary? ProcessorUtil references the old types by reflection, but it's only used if the dependency is already present. 

Does volley or dagger reference the support library?

arturogutierrez:
Yep, seems are not necessary indeed. Just cleaning some code then (Volley bump was required by a lint rule).

annotation/compiler/src/main/java/com/bumptech/glide/annotation/compiler/ProcessorUtil.java
```diff
@@ -48,7 +48,6 @@
 import javax.lang.model.type.TypeMirror;
 import javax.lang.model.type.TypeVariable;
 import javax.lang.model.util.ElementFilter;
-import javax.lang.model.util.Elements;
```
sjudd: The lines removed in this file are still needed for projects that use the support library, including internal projects.

sjudd:
I think we're not left with much other than the version bumps in the gradle files. If you'd like to submit those, I'm happy to accept.

arturogutierrez:
@sjudd You're right. This is unnecessary indeed so can be closed

PR #4015, opened 2019-12-18
support cronet net library whose version is 79.0.3945.79

bgylde:


sjudd:
Sorry for the delay here. We actually have an internal version of a cronet integration that I'd be happy to open source. My main issue is that I haven't had the chance to figure out how to build cronet externally.

Where did you get the cronet files? Did you build them yourself?

sjudd:
Ok to answer my own question, there are a few versions of cronet available on maven: https://mvnrepository.com/artifact/org.chromium.net. However the recommended way to do depend on cronet is via gms core: https://developer.android.com/guide/topics/connectivity/cronet/start

bgylde:
> Sorry for the delay here. We actually have an internal version of a cronet integration that I'd be happy to open source. My main issue is that I haven't had the chance to figure out how to build cronet externally.
> 
> Where did you get the cronet files? Did you build them yourself?

Yes, I compiled the cronet myself. Souce code was from chromium.

bgylde:
> Ok to answer my own question, there are a few versions of cronet available on maven: https://mvnrepository.com/artifact/org.chromium.net. However the recommended way to do depend on cronet is via gms core: https://developer.android.com/guide/topics/connectivity/cronet/start

I had ever tried the play-services-cronet, but it has to depend on the google service. Or I can delete the cronet jar and so, just use the compileOnly way.

sjudd:
Replaced by #4404

Issue #665, opened 2015-10-02
IGNORE exif image Rotation. 


xX7:
Hey. First of all, thank you for this great image loading library. 

Currently I'm working on an app where its not preferred, that glide automatically sets the correct rotation of the image when loading it as bitmap or in an image view. The app does the rotation handling itself from default and I want to keep the system. 

Is there a option to tell glide to ignore the image rotation and just load it as it is?

TWiStErRob:
Not really, it's hidden deep inside `Downsampler#decode`.

The only way I can imagine with the current version is to create your own `DataFetcher` that wraps the original one, returns a wrapped `InputStream` from `loadData`, then when that wrapping stream reads the underlying one you are also parsing the JPEG/EXIF structure and when you're in the right place just swap a `byte` before returning it. That way Glide won't know about the rotation, but I think this means that even the SOURCE cached image will have the hacked rotation saved; depending on how your app handles rotation by itself this may or may not jeopardize this workaround.

TWiStErRob:
Quick question: are you trying to use Glide as a networking library? (for example only using `.downloadOnly()` calls)

TWiStErRob:
Take a look at [`ExifOrientationStream`](https://github.com/bumptech/glide/blob/3.0/library/src/main/java/com/bumptech/glide/load/data/ExifOrientationStream.java), I think that does exactly this.

TheFinestArtist:
@xX7 I think you can do it just by commenting out this line `result = new ExifOrientationStream(result, orientation);` at [ThumbFetcher.java](https://github.com/bumptech/glide/blob/30c92551ee75c2109955ee653e8795c7c1d60bf8/library/src/main/java/com/bumptech/glide/load/data/mediastore/ThumbFetcher.java). I think you might have to import library in an old way though...

TWiStErRob:
That's for media store thumbnails, loading Files and URLs go on a different path. If you really want to modify the source code go for Downsampler, every load goes through that code.

TWiStErRob:
Let us know if we can help further.

thinkluffy:
Thanks for the great library again!
Is there an easy way to ignore the EXIF in v4?

sjudd:
[...]

PR #4051, opened 2020-01-16
Option to ignore exif orientation

emarc-m:
## Description

This PR adds an option to request for ignoring the `EXIF` orientation parsing in `Downsampler`.  

## Motivation and Context

Addresses: https://github.com/bumptech/glide/issues/665. For some applications (like ours), the application keeps track of the orientation not stored as the EXIF data.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/load/resource/bitmap/Downsampler.java
+++ b/library/src/main/java/com/bumptech/glide/load/resource/bitmap/Downsampler.java
@@ -112,8 +112,8 @@ public final class Downsampler {
           "com.bumptech.glide.load.resource.bitmap.Downsampler.AllowHardwareDecode", false);
 
   /**
-   * Indicates if the {@link androidx.exifinterface.media.ExifInterface.TAG_ORIENTATION} should be
-   * ignored and load the bitmap without rotating the decoded image
+   * Indicates if {@link androidx.exifinterface.media.ExifInterface.TAG_ORIENTATION} should be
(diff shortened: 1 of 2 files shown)


emarc-m:
@googlebot I signed it!

sjudd:
I don't believe it's possible to respect this flag with the new ImageDecoder library? If it is, can you update https://github.com/bumptech/glide/blob/3d54ef50ea4fa41694695971b7d858435c784057/library/src/main/java/com/bumptech/glide/load/resource/ImageDecoderResourceDecoder.java#L39.

emarc-m:
> I don't believe it's possible to respect this flag with the new ImageDecoder library? If it is, can you update
> 
> https://github.com/bumptech/glide/blob/3d54ef50ea4fa41694695971b7d858435c784057/library/src/main/java/com/bumptech/glide/load/resource/ImageDecoderResourceDecoder.java#L39
> 
> .

Thanks for the comment @sjudd. You are correct that this could not be enforced with `ImageDecoder`.  Do you have a suggestion on how to proceed with this?

sjudd:
I don't have any good ideas. If it can't be done with ImageDecoder, it's probably best not to expose the option for BitmapFactory. We're (slowly) moving towards ImageDecoder in Glide and it wouldn't surprise me if the framework continues to do the same.

emarc-m:
I understand. I'll close the PR. Thanks for your help taking a look at this.

PR #4095, opened 2020-02-17
Change Log Level

TerryJung:
## Description
Change log level for exception

## Motivation and Context
To be easy debugging,
if has some problem in image loading, it occured exception stacktrace log but it is info level
Many cases can not check to info level
so i think it need to be change Log level


TerryJung:
> @googlebot I signed it!

sjudd:
See http://bumptech.github.io/glide/doc/configuration.html#log-level

PR #4129, opened 2020-03-05
Improve the annotation of Encoder interface

Danny7226:
**Description**: 
Improve the annotation of Encoder interface

**Motivation and Context**:
Issue 1: “@​param​ file The F​​ile to write the data to.”

- F​ is unnecessarily capitalized

- Suggest editing annotation to: “@​param​ file The ​f​ile to write the data to.”

Issue 2: “@​param​ options The ​put​ ​of options to apply when encoding.”

- Options.java is described as a set of Option objects placed within an ArrayMap.

- Suggest that “​put​”​ ​could be changed to “​set​” or “​map​” to be more clear as to what the annotation “options” represents.


Danny7226:
@googlebot I fixed it.

Issue #4127, opened 2020-03-05
Glide RequestOptions cannot be backward compatible


penfeizhou:
In Class `com.bumptech.glide.RequestBuilder`,there's an important  called `apply`.
In Glide version `4.8.0`,The api is 
```java
public RequestBuilder<TranscodeType> apply(@NonNull RequestOptions requestOptions) {
...
}
```
But when on version `4.10.0`,The api changes to 
```java
 public RequestBuilder<TranscodeType> apply(@NonNull BaseRequestOptions<?> requestOptions) {
    Preconditions.checkNotNull(requestOptions);
    return super.apply(requestOptions);
  }

```
Despite that `RequestOptions` inherit `BaseRequestOptions`,The method  signature is not the same.
So this causes a problem: WhenI use Glide 4.10.0 in my APP,but one of my dependent library use Glide 4.8.0 and use `apply` method in it's code which is already compiled to class file.It cannot find the right method.
In another library:
```java
RequestBuilder<Drawable> requestBuilder = Glide.with(getContext()).load(prop.asString().value());
Drawable placeHolderDrawable = getPlaceHolderDrawable();
if (placeHolderDrawable != null) {
   requestBuilder = requestBuilder.apply(new RequestOptions().placeholder(placeHolderDrawable));
}
```
It causes folloiw problem:

**Stack trace / LogCat**:
```ruby
E/AndroidRuntime: FATAL EXCEPTION: main
    Process: xxxxx, PID: 25279
    java.lang.NoSuchMethodError: No virtual method placeholder(Landroid/graphics/drawable/Drawable;)Lcom/bumptech/glide/request/RequestOptions; in class Lcom/bumptech/glide/request/RequestOptions; or its super classes (declaration of 'com.bumptech.glide.request.RequestOptions' appears in /data/app/xxxxxx-lCBjFUgNVZ5bnGfcjjv3Ow==/base.apk)
```

PR #4142, opened 2020-03-17
Fix backward compatiable issue

Danny7226:
## Description
* Fix [issue.](https://github.com/bumptech/glide/issues/4127)

## Motivation and Context
* When the #placeholder() method is called with 4.8.0 dependency, the previously compiled .class file will search for the method in RequestionOptions, where it doesn’t exist. To resolve the issue, we overwrote the #placeholder() method in the RequestOptions class, which will call its super class accordingly.


sjudd:
So this only fixes the issue for a couple of methods. I think you'd need to go through and override all of the inherited methods to be sure? Would you be up for doing that?

PR #4167, opened 2020-04-01
Mark RequestListener.onLoadFailed() model parameter as Nullable

opayen:
## Description
Add a `@Nullable` annotation to the `model` parameter of the `RequestListener.onLoadFailed()` method.

## Motivation and Context
We had a crash happening when using Kotlin that could have been avoided if the `model` parameter was annotated as nullable. 
The crash happened with the following code:
```
override fun onLoadFailed(
    e: GlideException?,
    model: Any,
    target: Target<Drawable>,
    isFirstResource: Boolean
): Boolean {
    adView.setGone()
    return false
}
```
Then if the `model` turns out to be null, a `java.lang.IllegalArgumentExceptionParameter` happens.

Adding this annotation will prevent the issue. It also acts as documentations, indicating that the parameter can be null.


opayen:
@googlebot I signed it!

PR #4193, opened 2020-04-22
Avoid IllegalArgumentException in GifFrameLoader.

dalinaum:
## Description

Avoid IllegalArgumentException in GifFrameLoader.

## Motivation and Context

I encountered IllegalArgumentException, which is  thrown when the target is invisible and startFromFirstFrame is called  and the target subsequently becomes visible.

repo steps:
1. target.setVisible(false, ...)
2. drawable.startFromFirstFrame()
3. (after a while) target.setVisible(true, ...)

### DIFF
--- a/library/src/main/java/com/bumptech/glide/load/resource/gif/GifFrameLoader.java
+++ b/library/src/main/java/com/bumptech/glide/load/resource/gif/GifFrameLoader.java
@@ -268,7 +268,11 @@ class GifFrameLoader {
     // already incremented the frame pointer and can't decode the same frame again. Instead we'll
     // just hang on to this next frame until start() or clear() are called.
     if (!isRunning) {
-      pendingTarget = delayTarget;
+      if (startFromFirstFrame) {
+        handler.obtainMessage(FrameLoaderCallback.MSG_CLEAR, delayTarget).sendToTarget();
+      } else {
+        pendingTarget = delayTarget;
+      }
       return;
     }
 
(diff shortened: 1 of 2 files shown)


dalinaum:
@sjudd Is there something to improve this PR?

dalinaum:
Leave a comment to prevent it from closing automatically.

sjudd:
Sorry for the delay!

sjudd:
So my understanding is that the exception occurs when:

1. The gif starts playing
2. A frame is posted with a target delay
3. The gif is stopped, clearing the target but not removing the delay
4. Someone calls startFromFirstFrame
5. The delay finishes calling onFrameReady which sets the pendingTarget to be non-null
6. Eventually the gif is restarted and crashes because pendingTarget is non-null and startFromFirstFrame is true.

Does that  match your understanding? 

If so, it seems like your fix will work, but only by effectively racily ignoring calls to startFromFirstFrame.

I might suggest an alternative fix, which is to avoid setting the pendingTarget if it completes while isRunning is false and startFromFirstFrame is true. It also seems to suggest there's a larger problem with the delay mechanism here where the delayed targets are not being cancelled. If we fixed that, you'd never get the call to onFrameReady and wouldn't have to make any changes.

dalinaum:
@sjudd I updated the PR.

dalinaum:
I applied MSG_CLEAR instead of invoking requestManager.clear directly.

dalinaum:
@sjudd  Google internal checks FAILED but I cannot visit http://cl/311808317 . What should I fix for it?

PR #4214, opened 2020-05-11
Add support for width / height to the SvgDecoder

chggr:
## Description
Modify the sample SvgDecoder to use parameters width and height that are
passed into the decode() method.

## Motivation and Context
The sample SvgDecoder does not currently use parameters with and height
that are passed into the decode() method. This can be misleading for
Glide users and causes basic functionality such as image scaling to not
work properly. This change adds code to change the SVG document width
and height according to the provided parameters.


ashishpatel16:
@chggr I've been facing problems scaling my SVGs to ImageView since couple of hours now. I was fetching SVG from url but they would not scale properly (gets cropped). I actually fixed this by removing the code committed in this PR. I hope this helps people facing same issues.

chggr:
@ashishpatel16 thanks for raising this. I am not familiar with your setup, but the change has been working without issues for me since it was committed here.

PR #4222, opened 2020-05-15
use new API to learn about connectivity changes

jiyewushen:
use the ConnectivityManager.NetworkCallback API replace BroadcastReceiver when api >= 23 ,  see [ConnectivityManager.CONNECTIVITY_ACTION and NetworkInfo deprecated](https://stackoverflow.com/questions/36421930/connectivitymanager-connectivity-action-deprecated)


library/src/main/java/com/bumptech/glide/manager/DefaultConnectivityMonitor.java
```diff
@@ -27,43 +31,97 @@
...
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+      connectivityReceiver = null;
+      callback =
+          new ConnectivityManager.NetworkCallback() {
+            @Override
+            public void onAvailable(@NonNull Network network) {}
+
+            @Override
+            public void onLost(@NonNull Network network) {
+              onConnectivityChanged(false);
```
sjudd, line 49: Why do you need to set connectivity false here? is onCapabilitiesChanged not always called when a network is lost or becomes available?

library/src/main/java/com/bumptech/glide/manager/DefaultConnectivityMonitor.java
```diff
@@ -27,43 +31,97 @@
...
+            }
+
+            @Override
+            public void onCapabilitiesChanged(
+                @NonNull Network network, @NonNull NetworkCapabilities networkCapabilities) {
+              boolean connected =
+                  networkCapabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET)
+                      && networkCapabilities.hasCapability(
+                          NetworkCapabilities.NET_CAPABILITY_VALIDATED);
+              onConnectivityChanged(connected);
```
sjudd, line 59: Can this just call isConnected? It seems like you've updated that method anyway?

library/src/main/java/com/bumptech/glide/manager/DefaultConnectivityMonitor.java
```diff
@@ -27,43 +31,97 @@
...
-            }
-
-            listener.onConnectivityChanged(isConnected);
-          }
-        }
-      };
+  private final BroadcastReceiver connectivityReceiver;
+  private final ConnectivityManager.NetworkCallback callback;
 
   DefaultConnectivityMonitor(@NonNull Context context, @NonNull ConnectivityListener listener) {
```
sjudd, line 37: It's probably cleaner to split this into two classes, a pre M class and a post M class. Then the factory can just pick the appropriate class once, rather than having all of the if/else statements through the single class.

library/src/main/java/com/bumptech/glide/manager/DefaultConnectivityMonitor.java
jiyewushen: ConnectivityManager.getNetworkCapabilities is synchronous method as this is prone to race conditions

library/src/main/java/com/bumptech/glide/manager/DefaultConnectivityMonitor.java
jiyewushen: I'll check and reformat it

library/src/main/java/com/bumptech/glide/manager/ConnectivityStrategy.java
```diff
@@ -0,0 +1,10 @@
+package com.bumptech.glide.manager;
+
+/** A strategy that to perceive connectivity changes. */
+public interface ConnectivityStrategy {
```
sjudd, line 4: Can you use the existing ConnectivityMonitor interface instead of creating a new one?

library/src/main/java/com/bumptech/glide/manager/PostMConnectivityStrategyImpl.java
```diff
@@ -0,0 +1,150 @@
...
+import com.bumptech.glide.manager.ConnectivityMonitor.ConnectivityListener;
+import com.bumptech.glide.util.Preconditions;
+import com.bumptech.glide.util.Synthetic;
+
+/**
+ * A strategy that use {@link ConnectivityManager.NetworkCallback} and {@link
+ * ConnectivityManager#getNetworkCapabilities} to perceive connectivity changes.
+ */
+@RequiresApi(api = VERSION_CODES.M)
+public class PostMConnectivityStrategyImpl implements ConnectivityStrategy {
```
sjudd, line 24: final

library/src/main/java/com/bumptech/glide/manager/ConnectivityStrategy.java
jiyewushen: why do that? I'm a little confused.  in my eyes, strategy is mean the  way of check network ,and ConnectivityMoniter only care when check.

library/test/src/test/java/com/bumptech/glide/manager/PostMConnectivityMonitorTest.java
```diff
@@ -0,0 +1,100 @@
...
+  public void testRegistersNetworkCallbackOnRegister() {
+    monitor.register();
+    assertThat(harness.getNetworkCall()).contains(monitor.callback);
+  }
+
+  @Test
+  public void testDoesNotRegisterTwiceOnRegister() {
+    monitor.register();
+    monitor.register();
+    assertThat(harness.getNetworkCall()).contains(monitor.callback);
```
kanelbulle, line 49: This does not actually verify it's only registered once - I suggest you use `.containsExactly(monitor.callback)`

library/src/main/java/com/bumptech/glide/manager/PostMConnectivityMonitor.java
```diff
@@ -0,0 +1,161 @@
...
+  @SuppressWarnings("WeakerAccess")
+  @Synthetic
+  // Permissions are checked in the factory instead.
+  @SuppressLint("MissingPermission")
+  boolean isConnected(@NonNull ConnectivityManager connectivityManager) {
+    NetworkCapabilities networkCapabilities;
+    try {
+      networkCapabilities =
+          connectivityManager.getNetworkCapabilities(connectivityManager.getActiveNetwork());
+    } catch (RuntimeException e) {
```
kanelbulle, line 133: Why is it necessary to make such a broad catch of RuntimeException here? At a minimum, could you add a comment on why here? Even better would be if we could slim this down to the specific exception we're concerned about (and probably document that as well).

library/test/src/test/java/com/bumptech/glide/manager/PostMConnectivityMonitorTest.java
jiyewushen: ok

library/src/main/java/com/bumptech/glide/manager/PostMConnectivityMonitor.java
jiyewushen: getNetworkCapabilities may throw RemoteException that was wrapped by RuntimeException

Tolriq:
Beware of issues on Android 7 with callbacks being called without actual status yet being OK.

See for example https://github.com/google/ExoPlayer/issues/7453 and I've faced that in my apps too.

I would recommend to only use the new approach on Android O(8.0)+

Issue #4177, opened 2020-04-13
VerifyError on KitKat after R8 optimization


hqzxzwb:
**Glide Version**: 4.11.0

**Integration libraries**: okhttp3

**Device/Android Version**: Redmi 3 4.4. But probably shows on all dalvik devices.

**Issue details / Repro steps / Use case background**: 
Use R8 with proguard-android-optimization.txt rules. The app will crash with a VerifyError.

I am skipping usage details here because it's probably not relevant.

**Stack trace / LogCat**:
```ruby
E/AndroidRuntime(  567): FATAL EXCEPTION: main
E/AndroidRuntime(  567): Process: ***, PID: 567
E/AndroidRuntime(  567): java.lang.VerifyError: l/f/a/m/i/m
E/AndroidRuntime(  567): 	at com.bumptech.glide.Glide.<init>(SourceFile:43)
E/AndroidRuntime(  567): 	at com.bumptech.glide.Glide.a(SourceFile:102)
E/AndroidRuntime(  567): 	at com.bumptech.glide.Glide.a(SourceFile:14)
E/AndroidRuntime(  567): 	at com.bumptech.glide.Glide.b(SourceFile:2)
E/AndroidRuntime(  567): 	at com.bumptech.glide.Glide.a(SourceFile:119)
E/AndroidRuntime(  567): 	at e.a.c.k.c$a.a(SourceFile:122)
E/AndroidRuntime(  567): 	at ***.databinding.BottomButtonBindingImpl.executeBindings(SourceFile:26)
E/AndroidRuntime(  567): 	at androidx.databinding.ViewDataBinding.executeBindingsInternal(SourceFile:11)
E/AndroidRuntime(  567): 	at androidx.databinding.ViewDataBinding.executeBindingsOn(SourceFile:1)
E/AndroidRuntime(  567): 	at ***.databinding.RootViewBindingImpl.executeBindings(SourceFile:17)
E/AndroidRuntime(  567): 	at androidx.databinding.ViewDataBinding.executeBindingsInternal(SourceFile:11)
E/AndroidRuntime(  567): 	at androidx.databinding.ViewDataBinding.executePendingBindings(SourceFile:2)
E/AndroidRuntime(  567): 	at e.a.a.a.a.c.<init>(SourceFile:13)
E/AndroidRuntime(  567): 	at ***.fragment.root.RootViewFragment.onCreateView(SourceFile:1)
[...]

PR #4225, opened 2020-05-19
Update README.md with updated proguard rules

carlonzo:
## Description
Updates proguard rules following the rules in `glide/library/proguard-rules.txt`

## Motivation and Context
The rules suggested in the README are not in sync with the ones in the library.
This will just copy-paste the rules used there.
Also those rules are necessary to fix https://github.com/bumptech/glide/issues/4177


carlonzo:
(up)

sjudd:
Thanks!

PR #4241, opened 2020-06-02
Add test for Glide.with before Activity's super.onCreate call

sjudd:
This makes sure we're not actually adding a duplicate Fragment. No logic changes.


Juansbulbisful:
Fixed

Juansbulbisful:
Lfix

PR #4243, opened 2020-06-04
Update README.md

dalinaum:
## Description

Use jcenter instead of mavenCentral. It is specified by new Android projects.

## Motivation and Context

Since jcenter is specified by Android Studio automatically, I think using jcenter is more easy way to import Glide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,8 +26,8 @@ Or use Gradle:
 
 ```gradle
 repositories {
-  mavenCentral()
   google()
+  jcenter()
 }
 
 dependencies {


dalinaum:
(wait)

dalinaum:
(up)

jameswald:
jcenter is shutting down on May 1 so `google()` or `mavenCentral()` should be considered.

https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

dalinaum:
Many libraries of Google are still belong to jCenter. So Android application developers cannot use mavenCentral yet. It won't be too late to consider such a concern after Google migrates its libraries to mavenCentral or their repositories.

PR #4316, opened 2020-08-19
Deduplicate HTTP Uri ModelLoaders

### LINKED ISSUE #2943: Question: HttpUriLoader and UrlUriLoader both used when image has 404: Expected?
**Glide Version**: 4.6.1
**Integration libraries**: None
**Device/Android Version**: Robolectric environment
**Issue summary**: When attempting to load an https image which returns a 404 error, the request is performed twice.
I have a test case which requests an image url, and the server returns a 404 not found. I use Wiremock for the server.

The glide code in the test:
```java
        String url = "https://127.0.0.1:31626/static/country/FR/image.jpg";
        GlideApp.with(RuntimeEnvironment.application)
                .asBitmap()
                .load(url)
                .into(new SimpleTarget<Bitmap>(100, 100) {
                    @Override
                    public void onLoadFailed(@Nullable Drawable errorDrawable) {
                        super.onLoadFailed(errorDrawable);
                        System.out.println("error");
                    }

                    @Override
                    public void onResourceReady(@NonNull Bitmap resource, @Nullable Transition<? super Bitmap> transition) {
                        System.out.println("ready");
                    }
                });
```
I noticed from Wiremock logs that the request for the image is actually being requested twice.

When debugging, I noticed:

A `MultiModelLoader` is being used. Its `modelLoaders` field has six loaders, two of which are a `HttpUriLoader` and a `UrlUriLoader`.  In `buildLoadData()`, both of these match the `handles(model)` test, so their corresponding fetchers are added to the `fetchers` field. In this case both fetchers are of type `HttpUrlFetcher`.

I didn't follow the debugger further than this to understand more, but it looks like the first fetcher fails to load the image (the 404), and so the second one must be attempted (which will also fail with a 404). I'm not sure this is efficient. Is this expected?

dlew:
## Description

This change deduplicates the identical efforts made by `UrlUriLoader` and `HttpUriLoader` by deprecating `HttpUriLoader` and removing it from the registry.

Fixes #2943

## Motivation and Context

Glide had two essentially identical ModelLoaders: UrlUriLoader and HttpUriLoader. They both were being used to defer HTTP/HTTPS Uris to another ModelLoader.

Best-case scenario, this was a harmless distraction; but in some bad cases, it was possible for both ModelLoaders to fire, duplicating work.

With this change, only one of the two HTTP Uri ModelLoaders are entered into the registry. The one no longer being used has been deprecated as well.

I chose to deprecate HttpUriLoader because it can be implemented as a UrlUriLoader (but not vice versa, due to generics). To demonstrate that it still behaves the same, the test (which now targets UrlUriLoader) is unchanged but still passes.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/Glide.java
+++ b/library/src/main/java/com/bumptech/glide/Glide.java
@@ -50,7 +50,6 @@ import com.bumptech.glide.load.model.UnitModelLoader;
 import com.bumptech.glide.load.model.UriLoader;
 import com.bumptech.glide.load.model.UrlUriLoader;
 import com.bumptech.glide.load.model.stream.HttpGlideUrlLoader;
-import com.bumptech.glide.load.model.stream.HttpUriLoader;
 import com.bumptech.glide.load.model.stream.MediaStoreImageThumbLoader;
 import com.bumptech.glide.load.model.stream.MediaStoreVideoThumbLoader;
 import com.bumptech.glide.load.model.stream.QMediaStoreUriLoader;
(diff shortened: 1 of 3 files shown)


dlew:
Is there anything I could do to help get this PR approved and merged?

dlew:
Should I reopen this pull request, or start a new one with changes? It seems like this never got any maintainer attention, but the change is still valid IMO.

PR #4326, opened 2020-08-31
I think it can be optimized here.

ThirdGoddess:
## Description

Optimize the loop structure.

## Motivation and Context

Avoid reuse `preloadTargetQueue.queue.size()`. So I think it can be modified in this way.


ThirdGoddess:
@googlebot I signed it!

PR #2771, opened 2017-12-31
PMD 6.0.0

TWiStErRob:
I suggest a Squash Merge again. I suggest you take a look at the two revert commits and then the extra commits I added later. Please make sure you review thoroughly, hopefully rebase didn't mess anything up. Feel free to commit any amendments to the branch (you have write rights to all PR branches).

## Description
Upgrade PMD from 5.8.1 to 6.0.0 based on https://github.com/sjudd/glide/tree/pmd_6_0_0 branch rebased onto #2762. Mostly PMD triggered fixes and refactors.

## Motivation and Context
Better checks for finding hidden issues in the code.

PR #4354, opened 2020-09-23
Make consistent with #2771

T45K:
## Description

In #2771, `HttpUrlFetcher#loadData` was refactored ([here](https://github.com/bumptech/glide/pull/2771/files#diff-6e349798f5099eef01d1d3bce3b36058L55-L70)), but its similar method, `AssetPathFetcher#loadData` was not modified.
I thought the modification is needed for consistency.

## Motivation and Context

This PR makes this project stay consistent.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/load/data/AssetPathFetcher.java
+++ b/library/src/main/java/com/bumptech/glide/load/data/AssetPathFetcher.java
@@ -30,14 +30,13 @@ public abstract class AssetPathFetcher<T> implements DataFetcher<T> {
   public void loadData(@NonNull Priority priority, @NonNull DataCallback<? super T> callback) {
     try {
       data = loadResource(assetManager, assetPath);
+      callback.onDataReady(data);
     } catch (IOException e) {
       if (Log.isLoggable(TAG, Log.DEBUG)) {
         Log.d(TAG, "Failed to load data from asset manager", e);
       }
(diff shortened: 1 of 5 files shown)


sjudd:
Thanks. I don't have super strong opinions one way or another, but there are several other places where we follow the pattern you're refactoring away from: https://docs.google.com/document/d/1tTnd0SFd_e7UExQZClqDekMWxL7_zgS_0qlnkZRXJuw/edit?disco=AAAAG3g04TY&usp_dm=true&ts=5f6a8ec0.

Do you want to make this change across all of these similar implementations?

T45K:
@sjudd Thank you for your comment.

I have now requested access to the document. 
As soon as the access rights are granted, I'll start the modification.

sjudd:
Ugh sorry, wrong link, here's the right one: https://github.com/bumptech/glide/blob/50ebdbff0bebee4e31075f26d4263312ab5ef04a/library/src/main/java/com/bumptech/glide/load/data/LocalUriFetcher.java#L52

T45K:
@

T45K:
@sjudd 
Done. Would you check it again?

T45K:
@sjudd 

I've modified all `loadData` methods, which have `try-catch` statement
Would you check again?

sjudd:
Thanks!

PR #4365, opened 2020-10-08
Add Gif bounce effect

ku8ar:
## Description
Added GifOptions.BOUNCE option, which, when activated, changes the GIF playback support.

GifOptions.BOUNCE = false (default, as before):
1st frame, 2nd frame, 3d frame, 1st frame, 2nd frame, 3d frame, etc.

GifOptions.BOUNCE = true
1st frame, 2nd frame, 3d frame, 2nd frame, 1st frame, etc.

## Motivation and Context
We are making an application where GIFs have to be played this way.
We currently have it done using transcoding GIFs (adding extra frames), unfortunately the solution is bad, because it uses 2x more bandwidth transfer

At first I just wanted to fork this library but it couldn't be possible because Glide is not compatible with JitPack...

@sjudd Please reply. Such a feature can be included in this library? Because I don't know if I should fix tests etc.
I know that such a feature is not very universal, but even TikTok does something similar.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/load/resource/gif/ByteBufferGifDecoder.java
+++ b/library/src/main/java/com/bumptech/glide/load/resource/gif/ByteBufferGifDecoder.java
@@ -163,11 +163,6 @@ public class ByteBufferGifDecoder implements ResourceDecoder<ByteBuffer, GifDraw
   @VisibleForTesting
   static class GifDecoderFactory {
     GifDecoder build(
-        GifDecoder.BitmapProvider provider, GifHeader header, ByteBuffer data, int sampleSize) {
-      return new StandardGifDecoder(provider, header, data, sampleSize, false);
-    }
-
-    GifDecoder build(
         GifDecoder.BitmapProvider provider,
         GifHeader header,
         ByteBuffer data,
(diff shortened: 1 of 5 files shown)


ku8ar:
@googlebot I signed it!

ku8ar:
@googlebot I fixed it.

sjudd:
Thanks for sending this, sorry for the delay.

I think this needs more work though. GIFs generally decode frame N+1 starting from the contents of frame N. There are a few different types of GIFs that use different disposal options, but certainly this is one of the more common types.

As a result, it's not safe to just decode the frames in reverse order. The starting points for the frames won't be correct.

I think the only way to implement this feature correctly is to cache all of the generated frames and just iterate backwards and forwards between them. I suppose you could also try to decode all frames up to N+1, then all frames up to N etc, but that would be quite slow. Of course simply caching the frames is also risky because a sufficiently large GIF probably can't be held entirely in memory. We could try to make sure the frames are cached on disk, but that also gets tricky because the decoder needs to decode frames in order, it doesn't support random access.

All of this to say, we'd need a significantly more complicated implementation.

I wonder if there's a gif library for java/android that supports this already? If so you might be able to write an integration library to use that one instead of Glide's default. See http://bumptech.github.io/glide/doc/configuration.html#registering-components and/or the examples in Glide's integration package: https://github.com/bumptech/glide/tree/master/integration

ku8ar:
@sjudd Thank you for your answer. As I understand it, it's not so trivial...

But suppose this algorithm meets my goals, because all the GIFs I use are saved in cache. So, in this case, am I able to write an extension or override StandardGifDecoder to somehow replace the logic of the advance() method?

Issue #3421, opened 2018-11-20
load fails: wrong rounding and division by zero with fitCenter()


mice777:
**Glide Version**: 4.8.0

**Issue details / Repro steps / Use case background**: 
Very similar to issue #3420 
Now `Downsampler.calculateScaling` computes target dimensions, and can get to width or height of zero, and immediately divides by this, resulting in load error.
I suggest that the computations always produce Math.min(1, ...) so that rounding to zero never happens:
```
Downsampler.java:
private static int round(double value) {
    return Math.max(1, (int) (value + 0.5d));
  }
}
```

**Glide load line / `GlideModule` (if any) / list Adapter code (if any)**:
```java
GlideApp.with(act).load(file).fitCenter().into(v)
```

**Stack trace / LogCat**:
```ruby
    java.lang.ArithmeticException: divide by zero
        at com.bumptech.glide.load.resource.bitmap.Downsampler.calculateScaling(Downsampler.java:372)
        at com.bumptech.glide.load.resource.bitmap.Downsampler.decodeFromWrappedStreams(Downsampler.java:245)
        at com.bumptech.glide.load.resource.bitmap.Downsampler.decode(Downsampler.java:206)
        at com.bumptech.glide.load.resource.bitmap.StreamBitmapDecoder.decode(StreamBitmapDecoder.java:62)
        at com.bumptech.glide.load.resource.bitmap.StreamBitmapDecoder.decode(StreamBitmapDecoder.java:18)
        at com.bumptech.glide.load.resource.bitmap.BitmapDrawableDecoder.decode(BitmapDrawableDecoder.java:58)
        at com.bumptech.glide.load.engine.DecodePath.decodeResourceWithList(DecodePath.java:72)
        at com.bumptech.glide.load.engine.DecodePath.decodeResource(DecodePath.java:55)
        at com.bumptech.glide.load.engine.DecodePath.decode(DecodePath.java:45)
        at com.bumptech.glide.load.engine.LoadPath.loadWithExceptionList(LoadPath.java:58)
        at com.bumptech.glide.load.engine.LoadPath.load(LoadPath.java:43)
        at com.bumptech.glide.load.engine.DecodeJob.runLoadPath(DecodeJob.java:507)
        at com.bumptech.glide.load.engine.DecodeJob.decodeFromFetcher(DecodeJob.java:472)
        at com.bumptech.glide.load.engine.DecodeJob.decodeFromData(DecodeJob.java:458)
[...]

Issue #3420, opened 2018-11-20
load fails due to fitCenter computing zero-dimension bitmap size


mice777:
**Glide Version**: 4.8.0

**Issue details / Repro steps / Use case background**: 
I load image 5x900 pixels into view with size 238x178.
`TransformationUtils.fitCenter` got to situation where `minPercentage` was 0.19
then this code computed zero-width image size:
```
int targetWidth = Math.round(minPercentage * inBitmap.getWidth());
int targetHeight = Math.round(minPercentage * inBitmap.getHeight());
```
and this leads to load error due to failure in log as seen below

I suggest that above computations always produce Math.min(1, ...) so that rounding to zero never happens.

**Glide load line / `GlideModule` (if any) / list Adapter code (if any)**:
```kotlin
GlideApp.with(act).asDrawable()
                        .optionalFitCenter()
                        .load(this).into(v)
```

**Stack trace / LogCat**:
```ruby
    java.lang.IllegalArgumentException: width and height must be > 0
        at android.graphics.Bitmap.createBitmap(Bitmap.java:1061)
        at android.graphics.Bitmap.createBitmap(Bitmap.java:1040)
        at android.graphics.Bitmap.createBitmap(Bitmap.java:1007)
        at com.bumptech.glide.load.engine.bitmap_recycle.LruBitmapPool.createBitmap(LruBitmapPool.java:149)
        at com.bumptech.glide.load.engine.bitmap_recycle.LruBitmapPool.get(LruBitmapPool.java:131)
        at com.bumptech.glide.load.resource.bitmap.TransformationUtils.fitCenter(TransformationUtils.java:191)
        at com.bumptech.glide.load.resource.bitmap.FitCenter.transform(FitCenter.java:20)
        at com.bumptech.glide.load.resource.bitmap.BitmapTransformation.transform(BitmapTransformation.java:81)
        at com.bumptech.glide.load.engine.DecodeJob.onResourceDecoded(DecodeJob.java:540)
        at com.bumptech.glide.load.engine.DecodeJob$DecodeCallback.onResourceDecoded(DecodeJob.java:604)
        at com.bumptech.glide.load.engine.DecodePath.decode(DecodePath.java:46)
        at com.bumptech.glide.load.engine.LoadPath.loadWithExceptionList(LoadPath.java:58)
        at com.bumptech.glide.load.engine.LoadPath.load(LoadPath.java:43)
[...]

PR #4370, opened 2020-10-15
[#3420, #3421] Fix load fails due to fitCenter computing zero-dimension bitmap size 

PaulinaSadowska:
## Description
Limit bitmap scaling in fitCenter to not produce zero-dimension bitmap and avoid error during image load. 
FIX: 
[- #3420 load fails due to fitCenter computing zero-dimension bitmap size ](https://github.com/bumptech/glide/issues/3420)
[- #3421 load fails: wrong rounding and division by zero with fitCenter() ](https://github.com/bumptech/glide/issues/3421)

### DIFF
--- a/instrumentation/src/androidTest/java/com/bumptech/glide/load/resource/bitmap/DownsamplerEmulatorTest.java
+++ b/instrumentation/src/androidTest/java/com/bumptech/glide/load/resource/bitmap/DownsamplerEmulatorTest.java
@@ -295,6 +295,21 @@ public class DownsamplerEmulatorTest {
         .run();
   }
 
+  // fixes issues #3420 #3421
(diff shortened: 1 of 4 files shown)


kanelbulle (changes requested):
The FIT_CENTER downsampler is documented as:
"Scales, maintaining the original aspect ratio, so that one of the image's dimensions is exactly equal to the requested size and the other dimension is less than or equal to the requested size."
(at the top of Downsampler.java)

Could you update this documentation to include the special case for 0-sized output?

sjudd:
Thanks for sending this! I'm closing it since we haven't heard a response in a while. Please feel free to re-open when you get the chance to address the comments.

PR #4403, opened 2020-11-12
Readme: Added info about kapt dependency on kotlin

arberg:


sjudd:
This is mentioned in the more detailed documentation here: http://bumptech.github.io/glide/doc/download-setup.html#kotlin. The page is linked from the readme.

PR #4448, opened 2020-12-17
ExternalCacheDiskCacheFactory is Deprecated

Ezaldeen99:
## Motivation and Context

According to the [documentation](https://bumptech.github.io/glide/javadocs/490/com/bumptech/glide/load/engine/cache/ExternalCacheDiskCacheFactory.html),  `ExternalCacheDiskCacheFactory` is Deprecated and we should use `ExternalPreferredCacheDiskCacheFactory` instead.


Ezaldeen99:
@googlebot I signed it!

PR #4478, opened 2021-01-21
implemented GlideProvider.java for retriever Context

ImanX:
declared with() without context method
declared get() without context method


ImanX:
@googlebot I signed it

PR #4546, opened 2021-04-12
TranscoderRegistry.getTranscodeClasses method should return registered transcode classes

fightyz:
## Description
The original `TranscoderRegistry.getTranscodeClasses` method always add the argument `transcode` to the result list `transcodeClasses`. For the purpose of this method, it should return the registered transcode classes, as `ResourceDecoderRegistry.getResourceClasses` do.

## Motivation and Context
This bug has not caused an error because element in the result of `getTranscodeClasses` has not been used yet.


PtPrashantTripathi (commented):
Approved

PR #4591, opened 2021-07-03
Add FitXY BitmapTransformation

wp529:
When ImageView with FitXY scaleType, RoundCorners for it maybe not work well, Because FitXY takes effect after set RoundCorners, FitXY use with RoundCorners can solve it, I hope it could be help.


wp529:
> @googlebot I signed it!

Issue #4631, opened 2021-09-10
The very first example from the documentation doesn't work


nilTheDev:
**Glide Version**: 4.12.0

**Integration libraries**: N/A

**Device/Android Version**:  Pixel_3a_API_28_x86 (Emulator)

**Issue details / Repro steps / Use case background**: 

It's my first time with Glide. Haven't tried any other image loading library ever. 

I have tried the first example the `README` provides. But it didn't work.

```kotlin
override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)
        setContentView(R.layout.activity_main)
        val image = findViewById<ImageView>(R.id.image)
        
        val url = "http://goo.gl/gEgYUd"
        
        Glide.with(this).load(url).into(image)
    }
```
It shows nothing on the screen. And the logs aren't of much help to me. 

``` ruby
W/Glide: Load failed for http://goo.gl/gEgYUd with size [2088x2088]
    class com.bumptech.glide.load.engine.GlideException: Failed to load resource
    There was 1 root cause:
    com.bumptech.glide.load.HttpException(Failed to connect or obtain data, status code: -1)
     call GlideException#logRootCauses(String) for more detail
      Cause (1 of 1): class com.bumptech.glide.load.engine.GlideException: Fetching data failed, class java.io.InputStream, REMOTE
    There was 1 root cause:
    com.bumptech.glide.load.HttpException(Failed to connect or obtain data, status code: -1)
     call GlideException#logRootCauses(String) for more detail
        Cause (1 of 1): class com.bumptech.glide.load.engine.GlideException: Fetch failed
    There was 1 root cause:
    com.bumptech.glide.load.HttpException(Failed to connect or obtain data, status code: -1)
     call GlideException#logRootCauses(String) for more detail
          Cause (1 of 1): class com.bumptech.glide.load.HttpException: Failed to connect or obtain data, status code: -1
I/Glide: Root cause (1 of 1)
    com.bumptech.glide.load.HttpException: Failed to connect or obtain data, status code: -1
        at com.bumptech.glide.load.data.HttpUrlFetcher.loadDataWithRedirects(HttpUrlFetcher.java:98)
        at com.bumptech.glide.load.data.HttpUrlFetcher.loadData(HttpUrlFetcher.java:58)
        at com.bumptech.glide.load.model.MultiModelLoader$MultiFetcher.loadData(MultiModelLoader.java:100)
[...]

PR #4650, opened 2021-09-22
Update example URL README.md from HTTP to HTTPS

jwinwood:
## Description
Updated the example URL in the README.md from HTTP to HTTPS. Android by default does not allow HTTP connections anymore unless `android:usesCleartextTraffic="true"` is set in the manifest.

Resolves issue #4631 

## Motivation and Context
Updating the README.md to avoid confusion for users of Glide for a non-Glide related issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ Simple use cases will look something like this:
   ...
   ImageView imageView = (ImageView) findViewById(R.id.my_image_view);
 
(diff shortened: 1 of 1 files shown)


jwinwood:
@googlebot I signed it!

jwinwood:
Needs the label 'non-library' but can't seem to add that

sir-njuguna (approved):

PR #4813, opened 2022-06-05
Clean up and rearrange gradle.properties

TacoTheDank:
## Description
I attempted to make it much easier to read the ``gradle.properties`` file. I also cleaned up some inconsistencies.

## Motivation and Context
Some different AndroidX libraries were using the same version property (e.g. ANDROID_X_VERSION, ANDROID_X_TEST_VERSION), so I fixed that.

I did not update any of the dependencies.

I got rid of a few unused properties (e.g. ANDROID_SUPPORT_VERSION, FINDBUGS_VERSION).

I got rid of a final jcenter usage hidden away in the disklrucache submodule.

I got rid of an unused dependency (``androidx.legacy:legacy-support-v4``).


sjudd:
Thanks for the cleanup!

PR #4828, opened 2022-07-05
Update gradle wrapper, fix some deprecations

TacoTheDank:
### 1st commit
- Updated gradle wrapper to 7.3.3 using Git Bash.

The command I ran was `./gradlew wrapper --gradle-version 7.3.3 --distribution-type bin` in Git Bash. You can independently verify the integrity of these files by running this command on your own machine and then using a file comparison tool (like WinMerge for example) to compare the results between this PR and your machine.

The updated JAR's integrity can be verified through this as well: https://gradle.org/release-checksums/

### 2nd commit
- The second commit fixes some gradle script deprecations and access violations.

### 3rd commit
- The third commit fixes some Android Gradle Plugin deprecations.

## Motivation and Context
Some things are deprecated, so I'm fixing them :)

### DIFF
--- a/annotation/compiler/build.gradle
+++ b/annotation/compiler/build.gradle
@@ -32,8 +32,8 @@ def repackagedJar = file("${packagingFolder}/repackaged.jar")
 def proguardedJar = file("${packagingFolder}/proguarded.jar")
 
 task compiledJar(type: Jar, dependsOn: classes) {
-    destinationDir = packagingFolder
-    archiveName = 'compiled.jar'
+    destinationDirectory.set(packagingFolder)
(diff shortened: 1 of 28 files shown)


sjudd:
Are the gradlew and gradlew.bat changes required? I hope not? Right now we can't accept those because we use a single shared version of Gradle internally which is, of course, on an older version.

We can update the properties file independently, which I think is sufficient to run on 7.3.3? At least it seems to be when I build?

TacoTheDank:
Luckily, it isn't required, I just figured it would be good to do it.

In this case, I'll revert 👍

Issue #3778, opened 2019-07-24
Glide does not find images in drawable-night


frankyxcs:
Hi 

soon with Android Q the night mode will be very important in my opinion and it would be nice if you can fix this little problem. Glide does not find pictures in the following directories:

drawable-night
drawable-night-hdpi    (also mdpi, xhdpi, xxhdpi and xxxhdpi)

Greets

Frank

michaelbukachi:
@frankyxcs Try using:
```
Glide.with(this)
            .load(ContextCompat.getDrawable(context, R.drawable.some_drawble))
```
instead of 
```
Glide.with(this)
            .load(R.drawable.some_drawable)
```

hborders:
I'm having the same problem. If I use [michaelbukachi's workaround](https://github.com/bumptech/glide/issues/3778#issuecomment-517909978) it works fine for me.

Also, if I explicitly setting a different key for my `Drawable` doesn't work.
```
boolean isNightModeEnabled = ...
String key = (isNightModeEnabled ? "night" +  : "notnight") + drawableResourceId;
Glide.with(context).load(drawableResourceId).signature(new ObjectKey(key)).into(imageView);
```

Neither did explicitly setting the theme:
```
Glide.with(context).load(drawableResourceId).theme(context.getTheme()).into(imageView);
```

neworld:
The main reason why are using Glide is that Glide handles OOM and also reduces fragmentation. Current workaround `.load(ContextCompat.getDrawable(context, R.drawable.some_drawble))` negates Glide benefits. We found an increase of OOM after this workaround. 

The problem why Glide is not working with the dark mode is because Glide loads images from the `application` context. Always.  Maybe we could work around on `Application`, but some activities have different day/night mode settings. 

Sorry for the friendly pinging you @sjudd, but I am really worried, what is the long strategy of this problem is? Is the problem too complex to be fixed, or just no one has a time for it (maybe we could help you).

sjudd:
Duplicate of #3751

Issue #3751, opened 2019-07-11
Placeholder cache with values and values-night resources


Skylive:
**Glide Version**: 4.8.0
**Integration libraries**: OkHttp -> 3.9.1
**Device/Android Version**: Galaxy S8 | Android 9.0 Pie

**Issue details / Repro steps / Use case background**: 

I'm working on a Day/Night application and I have two different placeholders for each mode.

(To understand how values and values-night works, see this : https://medium.com/androiddevelopers/appcompat-v23-2-daynight-d10f90c83e94)

To simplify, I'm using the same resource with two background color :

**placeholder.xml**
 `<?xml version="1.0" encoding="utf-8"?>
<shape xmlns:android="http://schemas.android.com/apk/res/android"
    android:shape="rectangle">
            <solid android:color="@color/placeholder_background" />
</shape>`

**values/colors.xml**
`<color name="placeholder_background">@color/white</color>`

**values-night/colors.xml**
`<color name="placeholder_background">@color/black</color>`

**Glide request**
`.....
new GlideOptions()
                        .placeholder(R.drawable.placeholder)
                        .error(R.drawable.placeholder)
                        .transform(transformation)
......`

My issue here is when the user is in the day mode, a white placeholder will be loaded and displayed, but if the user change to night mode, the placeholder will stay in cache and won't change to the night version of the placeholder.

I know there is a glide cache for images but I don't know how to clear the placeholder cache.

If you have any idea to clear this resource and display the right one.

sjudd:
Do you have a sample app that shows how this works? 

Glide doesn't actually cache placeholders, we just call Resources#getDrawable. It is possible though that the request itself is being cached: https://github.com/bumptech/glide/blob/71359c7e385dc7d287f0da5d9e6d533ca41c76e4/library/src/main/java/com/bumptech/glide/RequestBuilder.java#L624

michaelbukachi:
@Skylive Checkout my answer in issue #3778

sjudd:
Internal bug is b/139004348

sjudd:
Oh I take it back, there is no internal bug for this. The issue here is probably that we use the application context to load the placeholders, rather than the Activity context. Still need to investigate that.

sjudd:
There's a workaround for this here: https://github.com/bumptech/glide/issues/3778#issuecomment-517909978. It won't work when actually loading a resource Drawable with Glide, only for placeholder/fallback/error.

PR #4842, opened 2022-07-20
fixed placeholders load drawable resources with wrong context

shijiayao-main:
…t mode

## Description

issues #3778 #3751

fix unable to find night resources when using night mode

## Motivation and Context

when using 'AppCompatDelegate.seDefaultNightMode' change theme mode, glide will not use the right resources.

in ContextThemeWrapper if mOverrideConfiguration is null it will get resource by 'mBase.getResources()', in 'DrawableDecoderCompat' mBase is applicationContext, should use 'applyOverrideConfiguration' to get resource from current theme.

ResourcesCompat.getDrawable() resource should use 'theme.getResources' rather than 'context.getResources' because context is  applicationContext


Haoxiqiang:
@sjudd this commit is necessary for us. would you release a version?

kanelbulle:
Thank you for the fix! Would you be willing to add a test case for this?

osfunapps:
What about merging this branch with the main? The whole library is unusable for us using the dark mode

PR #4851, opened 2022-07-28
Add pre-release compose support to the Gallery sample app.

sjudd:
This isn't in any state to be released, but it might be the minimal
amount of code to get feedback on the implementation. There's lots of
work left to do before this could be merged, mostly specificed in TODOs
in the code.


samples/gallery/src/main/java/com/bumptech/glide/samples/gallery/HorizontalGalleryFragment.kt
```diff
@@ -1,50 +1,559 @@
...
+   * images exceeds the memory cache, preloading is effectively useless. However if you preload too
+   * few things, the buffer may be small enough that images are not available when they could be, so
+   * it's always a balancing act. The smaller the preloaded image, the more you can preload.
+   * @param fixedVisibleItemCount The number of visible items. In some cases this can vary widely
+   * in which case you can leave this value `null`. If the number of visible items is always one
+   * or two, it might make sense to just set this to the larger of the two to reduce churn in the
+   * preloader.
+   * @param requestBuilderModifier See [ListPreloader.PreloadModelProvider.getPreloadRequestBuilder]
+   */
+  // TODO(judds): Consider wrapping a LazyRow / LazyColumn and providing state instead of a separate
```
arriolac, line 151: ++ I think this would be a better API as well. Though I think it's good to also have an API that is not tied to LazyRow & LazyColumn for usages outside of those contexts.

samples/gallery/src/main/java/com/bumptech/glide/samples/gallery/HorizontalGalleryFragment.kt
```diff
@@ -1,50 +1,559 @@
...
+  // alternative way to pass through the Model so that it doesn't have to be set on the request
+  // builder and passed through.
+  @Composable
+  fun GlideImage(
+    requestManager: RequestManager,
+    model: Any?,
+    // TODO(judds): RequestBuilder doesn't extend BaseRequestOptions equals/hashcode method, so
+    // we may not recompose when we should. This should be easy to fix.
+    requestBuilder: RequestBuilder<Drawable>,
+    imageParameters: ImageParameters,
```
arriolac, line 250: Most composables in the toolkit accept a `Modifier` parameter. This should accept one as well so that it can be modified (for example, `Modifier.clip(CircleShape)` to clip the image). Related: https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-api-guidelines.md#elements-accept-and-respect-a-modifier-parameter This page also contains other Compose API guidelines that you can refer to.

samples/gallery/build.gradle
```diff
@@ -17,6 +17,9 @@ dependencies {
     implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:$JETBRAINS_KOTLINX_COROUTINES_VERSION"
     implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:$JETBRAINS_KOTLINX_COROUTINES_VERSION"
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$JETBRAINS_KOTLIN_VERSION"
+    implementation "androidx.compose.foundation:foundation:1.2.0-rc03"
+    implementation 'androidx.compose.ui:ui:1.2.0-rc03'
```
arriolac, line 21: This can use the stable (1.2.0) version of the libraries. Generally, we want libraries to use the latest stable version of Compose.

samples/gallery/src/main/java/com/bumptech/glide/samples/gallery/HorizontalGalleryFragment.kt
```diff
@@ -1,50 +1,559 @@
...
+  @Composable
+  private fun <DataTypeT: Any> rememberGlidePreloader(
+    data: List<DataTypeT>,
+    size: Size,
+    numberOfItemsToPreload: Int,
+    requestBuilderModifier:
+      (data: DataTypeT, requestBuilder: RequestBuilder<*>) -> RequestBuilder<*>,
+  ): ListPreloader<DataTypeT> {
+    Log.d("TEST", "rememberGlidePreloader")
+    val currentData = remember(data) { data }
```
zach-klippenstein, line 203: This is redundant, just read `data` directly. This actually introduces overhead, since `remember` has to read from and write to the slot table, which isn't free.

samples/gallery/src/main/java/com/bumptech/glide/samples/gallery/HorizontalGalleryFragment.kt
```diff
@@ -1,50 +1,559 @@
...
+    val requestManager = Glide.with(LocalContext.current)
```
zach-klippenstein, line 204: If `Glide.with` does significant work, you might want to `remember` this. If it's cheap, there's no need. If you do, it would look something like this:
```kotlin
val context = LocalContext.current
val requestManager = remember(context) { Glide.with(context) }
```

samples/gallery/src/main/java/com/bumptech/glide/samples/gallery/HorizontalGalleryFragment.kt
```diff
@@ -1,50 +1,559 @@
...
+      override fun getPreloadItems(position: Int): MutableList<DataTypeT> {
+        return mutableListOf(this.currentData[position])
+      }
+
+      override fun getPreloadRequestBuilder(item: DataTypeT): RequestBuilder<*> {
+        return requestBuilderModifier(item, requestManager.asDrawable())
+      }
+    }
+
+    val preloadModelProvider = remember { PreloadModelProvider() }
```
zach-klippenstein, line 216: Because this class captures parameters from the composable function, move this to an anonymous class (`object :`) inside the `remember` where you instantiate it to make it a bit more clear what the dependencies are that need to be keys for the remember. Or maybe even better, pull it out to a separate class that takes those dependencies as explicit constructor parameters. Right now the remember is missing some keys, and because the class captures the values outside the remember, I'm not even sure if providing the keys will be enough to make this work.

```suggestion
    val preloadModelProvider = remember(requestManager, requestBuilderModifier) {
      object : ListPreloader.PreloadModelProvider<DataTypeT> {
        var currentData: List<DataTypeT> = emptyList()
        override fun getPreloadItems(position: Int): MutableList<DataTypeT> {
          return mutableListOf(this.currentData[position])
        }

        override fun getPreloadRequestBuilder(item: DataTypeT): RequestBuilder<*> {
          return requestBuilderModifier(item, requestManager.asDrawable())
        }
      }
    }
```

samples/gallery/src/main/java/com/bumptech/glide/samples/gallery/HorizontalGalleryFragment.kt
```diff
@@ -1,50 +1,559 @@
...
+    preloadModelProvider.currentData = currentData
+    return remember {
```
zach-klippenstein, line 218: `remember`s need to be keyed with their dependencies:
```suggestion
    return remember(requestManager, preloadModelProvider, numberOfItemsToPreload) {
```
Note that because these are two remembers that feed into each other, they could be collapsed into a single remember.

samples/gallery/src/main/java/com/bumptech/glide/samples/gallery/HorizontalGalleryFragment.kt
```diff
@@ -1,50 +1,559 @@
...
+      ListPreloader(
+        requestManager,
+        preloadModelProvider,
+        { _, _, _ -> intArrayOf(size.width.toInt(), size.height.toInt()) },
```
zach-klippenstein, line 222: This captures the `size` value from the initial composition of this composable function, but will not get any updates if it changes. You probably want something like this:
```kotlin
// Above, directly in the composable function:
val updatedSize = rememberUpdatedState(size)

…
// On this line:
{ _, _, _ -> intArrayOf(updatedSize.width.toInt(), updatedSize.height.toInt()) },
```

samples/gallery/src/main/java/com/bumptech/glide/samples/gallery/HorizontalGalleryFragment.kt
```diff
@@ -1,50 +1,559 @@
...
+  private fun rememberGlidePainter(
+    requestManager: RequestManager,
+    model: Any?,
+    requestBuilder: RequestBuilder<Drawable>,
+    size: ResolvableGlideSize,
+  ): GlidePainter =
+    // TODO(judds): Calling onRemembered here manually might make a minor improvement in how quickly
+    //  the image load is started, but it also triggers a recomposition. I can't figure out why it
+    //  triggers a recomposition
+    remember { GlidePainter(requestManager, model, requestBuilder, size) }
```
zach-klippenstein, line 429: Probably needs some keys, otherwise the `GlidePainter` will be using stale captures if any of the parameters change.
```suggestion
    remember(requestManager, model, requestBuilder, size) {
      GlidePainter(requestManager, model, requestBuilder, size)
    }
```

samples/gallery/src/main/java/com/bumptech/glide/samples/gallery/HorizontalGalleryFragment.kt
```diff
@@ -1,50 +1,559 @@
...
+    override fun onForgotten() {
+      clear()
+      (delegate as? RememberObserver)?.onForgotten()
+    }
+
+    override fun onRemembered() {
+      if (rememberScope != null) {
+        return
+      }
+      val scope = CoroutineScope(SupervisorJob() + Dispatchers.Main.immediate)
```
zach-klippenstein, line 467: This class should probably accept the coroutine scope as a constructor parameter, and then the scope should be obtained from the composable using `rememberCoroutineScope`. Then the scope will take advantage of structured concurrency and you don't have to manage it yourself. E.g.

```kotlin
// Above, in rememberGlidePainter.
val scope = rememberCoroutineScope()
remember(…) { GlidePainter(…, scope) }

private inner class GlidePainter(
  …,
  scope: CoroutineScope
) {
  private val scope = scope + SupervisorJob(parent = scope.coroutineContext.job) + Dispatchers.Main.immediate
}
```

zach-klippenstein (commented):
I left some comments about general Compose correctness issues, but I'm not familiar with Glide so didn't look at the overall design of this integration.

samples/gallery/build.gradle
sjudd: Done.

samples/gallery/src/main/java/com/bumptech/glide/samples/gallery/HorizontalGalleryFragment.kt
```diff
@@ -1,50 +1,559 @@
...
+      )
+    doPreload(preloader = preloader, state = state, fixedVisibleItemCount = fixedVisibleItemCount)
+  }
+
+  @Composable
+  private fun <DataTypeT: Any> doPreload(
+    preloader: ListPreloader<DataTypeT>, state: LazyListState, fixedVisibleItemCount: Int?
+  ) {
+    Log.d("TEST", "doPreload")
+    val lazyListVisibleInfo = derivedStateOf { state.lazyListVisibleInfo(fixedVisibleItemCount) }
```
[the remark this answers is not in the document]
sjudd: Done, thanks this discussion is helping me understand this all a lot better!

samples/gallery/src/main/java/com/bumptech/glide/samples/gallery/HorizontalGalleryFragment.kt
```diff
@@ -1,50 +1,559 @@
...
+    // TODO(judds): Without this remember call, we're never recomposed. It's unclear what the right
+    // way to do this is?
+    remember(lazyListVisibleInfo.value) {}
+    SideEffect {
+      preloader.onScroll(
+        null,
+        lazyListVisibleInfo.value.firstVisibleItemIndex,
+        lazyListVisibleInfo.value.visibleItemCount,
+        lazyListVisibleInfo.value.totalItemCount)
+    }
```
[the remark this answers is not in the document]
sjudd: Done

samples/gallery/src/main/java/com/bumptech/glide/samples/gallery/HorizontalGalleryFragment.kt
sjudd: Ahh got it, ok done, thanks!

samples/gallery/src/main/java/com/bumptech/glide/samples/gallery/HorizontalGalleryFragment.kt
sjudd: Done

samples/gallery/src/main/java/com/bumptech/glide/samples/gallery/HorizontalGalleryFragment.kt
sjudd: Done

samples/gallery/src/main/java/com/bumptech/glide/samples/gallery/HorizontalGalleryFragment.kt
sjudd: Done.

I think the same is true here for data? If the data backing the list changes, we don't need a new preloader, but we do want to use the updated list.

samples/gallery/src/main/java/com/bumptech/glide/samples/gallery/HorizontalGalleryFragment.kt
sjudd: Done.

I'm a little concerned that the way I'm handling the size is incorrect here. The AsyncSize is a data class that wraps a function. The output of the function is dependent on the Modifiers. So in that case we probably should try to find a way to recompose if the input modifier changes?

samples/gallery/src/main/java/com/bumptech/glide/samples/gallery/HorizontalGalleryFragment.kt
sjudd: Done.

Just like with the TODO around calling `.onRemembered, in the new version of the code I see two calls to SizedGlideImage if I launch the image load in `init {}`, but one if I launch it in `onRemembered`:

```
2022-08-12 16:31:51.986 24711-24711/com.bumptech.glide.samples.gallery D/TEST: GlidePreloader
2022-08-12 16:31:51.986 24711-24711/com.bumptech.glide.samples.gallery D/TEST: rememberGlidePreloader
2022-08-12 16:31:52.141 24711-24711/com.bumptech.glide.samples.gallery D/TEST: GlidePreloader
2022-08-12 16:31:52.141 24711-24711/com.bumptech.glide.samples.gallery D/TEST: rememberGlidePreloader
2022-08-12 16:31:52.160 24711-24711/com.bumptech.glide.samples.gallery D/TEST: GlideImage: content://media/external/file/759
2022-08-12 16:31:52.161 24711-24711/com.bumptech.glide.samples.gallery D/TEST: SizedGlideImage: content://media/external/file/759, 2050449207
2022-08-12 16:31:52.165 24711-24711/com.bumptech.glide.samples.gallery D/TEST: Start load for content://media/external/file/759 into com.bumptech.glide.samples.gallery.HorizontalGalleryFragment$glideRequestFlow$1$target$1@e7d60c2
2022-08-12 16:31:52.176 24711-24711/com.bumptech.glide.samples.gallery D/TEST: collect
2022-08-12 16:31:52.184 24711-24711/com.bumptech.glide.samples.gallery D/TEST: layout!
2022-08-12 16:31:52.275 24711-24711/com.bumptech.glide.samples.gallery D/TEST: collect
2022-08-12 16:31:52.285 24711-24711/com.bumptech.glide.samples.gallery D/TEST: SizedGlideImage: content://media/external/file/759, 2050449207
2022-08-12 16:31:52.288 24711-24711/com.bumptech.glide.samples.gallery D/TEST: layout!
2022-08-12 16:31:52.314 24711-24711/com.bumptech.glide.samples.gallery D/TEST: collect
2022-08-12 16:31:52.318 24711-24711/com.bumptech.glide.samples.gallery D/TEST: layout!



2022-08-12 16:32:44.052 24816-24816/com.bumptech.glide.samples.gallery W/samples.galler: Accessing hidden method Landroid/content/pm/ApplicationInfo;->isSystemApp()Z (blacklist,test-api, linking, denied)
2022-08-12 16:32:44.881 24816-24816/com.bumptech.glide.samples.gallery D/TEST: GlidePreloader
2022-08-12 16:32:44.881 24816-24816/com.bumptech.glide.samples.gallery D/TEST: rememberGlidePreloader
2022-08-12 16:32:45.035 24816-24816/com.bumptech.glide.samples.gallery D/TEST: GlidePreloader
2022-08-12 16:32:45.035 24816-24816/com.bumptech.glide.samples.gallery D/TEST: rememberGlidePreloader
2022-08-12 16:32:45.051 24816-24816/com.bumptech.glide.samples.gallery D/TEST: GlideImage: content://media/external/file/759
2022-08-12 16:32:45.052 24816-24816/com.bumptech.glide.samples.gallery D/TEST: SizedGlideImage: content://media/external/file/759, 2050449207
2022-08-12 16:32:45.062 24816-24816/com.bumptech.glide.samples.gallery D/TEST: Start load for content://media/external/file/759 into com.bumptech.glide.samples.gallery.HorizontalGalleryFragment$glideRequestFlow$1$target$1@7720909
2022-08-12 16:32:45.072 24816-24816/com.bumptech.glide.samples.gallery D/TEST: collect
2022-08-12 16:32:45.073 24816-24816/com.bumptech.glide.samples.gallery D/TEST: layout!
2022-08-12 16:32:45.165 24816-24816/com.bumptech.glide.samples.gallery D/TEST: collect
2022-08-12 16:32:45.176 24816-24816/com.bumptech.glide.samples.gallery D/TEST: layout!
2022-08-12 16:32:45.203 24816-24816/com.bumptech.glide.samples.gallery D/TEST: collect
2022-08-12 16:32:45.209 24816-24816/com.bumptech.glide.samples.gallery D/TEST: layout!
```

Any thoughts on why starting the image load immediately triggers a recomposition, but waiting for `onRemembered` does not? The Flow that's called here should immediately produce a null Drawable, but if that's the cause, I'd expect recomposition regardless of which place I launch it.

samples/gallery/src/main/java/com/bumptech/glide/samples/gallery/HorizontalGalleryFragment.kt
sjudd: Done

Issue #4819, opened 2022-06-20
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.String.hashCode()' on a null object reference


michael-winkler:
**Description:**

It looks like there is a uncatched null pointer exception in the hash code function.    
`com.bumptech.glide.load.engine.EngineKey.hashCode(EngineKey.java:63)`

**Environment**

`implementation 'com.onesignal:OneSignal:4.7.3'`

**Steps to Reproduce Issue:**

I can't recreate the bug since it was posted by Firebase Crashlytics.

**Anything else:**

```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.String.hashCode()' on a null object reference
       at java.io.UnixFileSystem.hashCode(UnixFileSystem.java:427)
       at java.io.File.hashCode(File.java:2115)
       at com.bumptech.glide.load.engine.EngineKey.hashCode(EngineKey.java:63)
       at java.util.HashMap.hash(HashMap.java:338)
       at java.util.HashMap.get(HashMap.java:556)
       at com.bumptech.glide.load.engine.ActiveResources.get(ActiveResources.java:94)
       at com.bumptech.glide.load.engine.Engine.loadFromActiveResources(Engine.java:327)
       at com.bumptech.glide.load.engine.Engine.loadFromMemory(Engine.java:302)
       at com.bumptech.glide.load.engine.Engine.load(Engine.java:190)
       at com.bumptech.glide.request.SingleRequest.onSizeReady(SingleRequest.java:468)
       at com.bumptech.glide.request.SingleRequest.begin(SingleRequest.java:256)
       at com.bumptech.glide.manager.RequestTracker.runRequest(RequestTracker.java:41)
       at com.bumptech.glide.RequestManager.track(RequestManager.java:671)
       at com.bumptech.glide.RequestBuilder.into(RequestBuilder.java:813)
       at com.bumptech.glide.RequestBuilder.into(RequestBuilder.java:780)
       at com.bumptech.glide.RequestBuilder.into(RequestBuilder.java:771)
       at com.mugler.sitemanTicketSystem.adapters.FotosAdapter.onBindViewHolder(FotosAdapter.kt:89)
       at com.mugler.sitemanTicketSystem.adapters.FotosAdapter.onBindViewHolder(FotosAdapter.kt:42)
[...]

PR #4856, opened 2022-08-08
Fix NullPointerException 'String.hashCode()' on a null object reference

Mika-89:
## Description

This PR fixes the Issue https://github.com/bumptech/glide/issues/4819 NullPointerException on String.hashCode() call

## Motivation and Context

Required because there are a lot of crashs.


sjudd:
So this is actually an object re-use issue (https://bumptech.github.io/glide/doc/resourcereuse.html). Even if we fix the NPE, it doesn't lead to coherent behavior. Ideally we're just going to remove the object pooling. My guess is that on Art devices it's not necessary to avoid the object creation.

Mika-89:
Okay. So what's to do now? Or is there any working fix to get out of these crashs as described inside the issue?

PR #4876, opened 2022-08-19
fix: DecodeForamt.RGB_565 not working

yacyet:
Glide.with(this)
        .load(url)
        .format(DecodeFormat.PREFER_RGB_565);
When set 565 format, actually  this does not work. This problem has discussed in #4461. 
https://github.com/bumptech/glide/issues/4461

### DIFF
--- a/library/src/main/java/com/bumptech/glide/load/resource/bitmap/Downsampler.java
+++ b/library/src/main/java/com/bumptech/glide/load/resource/bitmap/Downsampler.java
@@ -908,17 +908,23 @@ public final class Downsampler {
       }
       // On API 26 outConfig may be null for some images even if the image is valid, can be decoded
       // and outWidth/outHeight/outColorSpace are populated (see b/71513049).
-      expectedConfig = options.outConfig;
+      // expectedConfig = options.outConfig;
+    }
+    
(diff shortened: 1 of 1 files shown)


sjudd:
I'd rather not comment out code as a general practice.

That said, I think the issue here is fundamentally that the framework is claiming it's going to use ARGB_8888 when we ask it, even if we say we want RGB_565. We might choose to ignore outConfig entirely, but I'm not sure that's the best idea. After all there may be cases where the framework can actually only produce ARGB_8888.

I think there's two things to do:

1. File a framework bug if we use inPreferredConfig = RGB_565,  inJustDecodeBounds=true, outConfig gives us ARGB_8888, but if we provide an input Bitmap with RGB_565 the decode just works fine. Preferably we'd have a test case for this
2. Up until whenever the framework bug in #1 is fixed, in Glide, try to decode with RGB_565 even if outConfig gives us ARGB_8888, but being sure to retry with ARGB_8888 if our RGB_565 load fails.

The best place to start is probably tests, rather than code. There's a bunch here, maybe we could add a new one or two? https://github.com/bumptech/glide/tree/master/instrumentation

yacyet:
Got it, thanks!

PR #4880, opened 2022-08-22
Avoid launching Glide's flow's clear

sjudd:
Launching the clear means a long delay when a lot of requests build up,
which in turns adds latency when requests are churning (ie scrolling
quickly through a list).

It was added to avoid clearing a Target in response to a Target
callback, but that only happens in certain types of forced main thread
requests, which won't happen in prod. In fact I've already removed the
behavior from the tests, so it's entirely bad.

### DIFF
--- a/integration/ktx/src/test/java/com/bumptech/glide/integration/ktx/FlowsTest.kt
+++ b/integration/ktx/src/test/java/com/bumptech/glide/integration/ktx/FlowsTest.kt
@@ -13,7 +13,9 @@ import com.bumptech.glide.Glide
 import com.bumptech.glide.GlideBuilder
 import com.bumptech.glide.RequestManager
 import com.bumptech.glide.load.DataSource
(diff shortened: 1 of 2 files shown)


integration/ktx/src/test/java/com/bumptech/glide/integration/ktx/FlowsTest.kt
```diff
@@ -336,6 +342,311 @@ class FlowsTest {
         )
     }
 
+<<<<<<< Updated upstream
```
kanelbulle, line 345: leftover merge markers..

kanelbulle (changes requested):

integration/ktx/src/test/java/com/bumptech/glide/integration/ktx/FlowsTest.kt
sjudd: Done!

PR #4934, opened 2022-10-25
Add `loading` and `failure` parameters to GlideImage

### LINKED ISSUE #4922: [Compose] Provide a mechanism way to provide a composable for placeholder and error states
Provide a way to provide a composable for loading and error states. For example:

```
GlideImage(
  model = imageUrl,
  contentDescription = null,
  placeHolder = {
    Image(...)
  },
  error = {
    Image(...)
   }
)
```

sjudd:
These parameters will override equivalent values set via requestBuilderTransform if the parameters are non-null. If the parameters are null, then we'll apply whatever is specified by requestBuilderTransform. I plan to add similar explicit parameters for other values settable via requestBuilderTransform over time with similar override behavior unless someone objects.

The `fallback` on `RequestBuilder` is not replicated. Callers can use `failure` or if/else to provide custom behavior for null models.

Similarly `loading` is only shown while the request is loading. Unlike `RequestBuilder.placeholder`, `loading` will never be shown after the load fail. `failure` should instead be used to explicitly duplicate the `loading` behavior if that functionality is desired.

Fixes #4922

### DIFF
--- a/instrumentation/src/androidTest/java/com/bumptech/glide/CachingTest.java
+++ b/instrumentation/src/androidTest/java/com/bumptech/glide/CachingTest.java
@@ -31,8 +31,8 @@ import com.bumptech.glide.request.target.Target;
 import com.bumptech.glide.test.GlideApp;
 import com.bumptech.glide.test.ResourceIds;
 import com.bumptech.glide.test.ResourceIds.raw;
-import com.bumptech.glide.test.WaitModelLoader;
-import com.bumptech.glide.test.WaitModelLoader.WaitModel;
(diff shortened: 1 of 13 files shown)


dimitriz09:
This a very nice feature i'd like to have ! Any alpha release including this planned soon ??
Regards,
Dimitri.

Abishek1997:
@sjudd I see a PR linked to this issue, but looks like its not part of the GlideImage yet? Any ideas when we can expect this?

danielgomesoliveira:
+1 here, this feature would be very useful

PR #4985, opened 2022-12-28
GlideUrl.equals with respect to headers elements

asimaruk:
GlideUrl equality by headers elements, not headers objects themself.

Why it is needed.
As a kotlin user you can easily create GlideUrl like this
```
GlideUrl(someUrl) {
    someMapOfHeaders
}
```
It will be an anonymous implementation of Headers interface, and now two GlideUrl with same url and headers won't be equal. Which is wrong in my opinion. It causes image flickering when data updates and GlideUrls constructed like this.


sjudd:
It's used because sometimes creating the headers is super expensive. You can implement equals/hashcode on the Headers implementation without generating the headers to fix the same issue.

PR #5022, opened 2023-02-14
Update 2015-05-17-getting-started.md

KirillMakarov:
Add missed bracket

## Description
add missed bracket in doc


sjudd:
Thanks!

Issue #942, opened 2016-01-29
Clearing resources onStop


TepesLucian:
This issue is based on #526 .
In case of an application that has a deep stack of activities (and some other use cases like device having a low heap memory) displaying images using Glide, I think it would be a great if the app could indicate Glide to clear/recycle resources on `onStop()` and reload the requests on `onStart()`. Since the activity/fragment is no longer visible to the user the memory can be reused by future activities.

I've managed to implement by manually loading the images `onStart` and clearing them `onStop` using `Glide.clear(view)` to avoid OOM in the case user opened a deep stack of activities. I think is fairly acceptable if the user opens multiple screens and navigating back to see at some point placeholders instead previous loaded images.

TWiStErRob:
@sjudd do you think this is feasible, especially the reloading on return part?

Some thoughts:
 * How would this affect rotation? Is the RMFragment stopped on rotation?
 * What kind of API could convey this best? I guess `Glide.with(...).autoManage()` or something.
 * Probably need a way to reset normal behavior
 * GlideModule also need to be able to set default to autoManage in case the whole app would use it.

sjudd:
I like this idea, though you'd want to be a bit careful with your cache size and Activity flow. I've worked on an app where they always cleared resources in onStop and re-loaded them in onStart, and we regularly had issues where a user would open an Activity then immediately hit the back button and see a bunch of grey squares where they had just seen images. 

* Rotation is ok, Glide will clear/restart when the Activity is destroyed and re-created anyway.
* API would be a method on the RequestManager, I wouldn't be opposed to adding something in the GlideModule as well to set a default.

It's less neat than autoManage, but a bit more descriptive, maybe: ``Glide.with(...).setClearOnStop(boolean)``

TepesLucian:
@sjudd I never seen this problem so far on my testing (maybe has to do with using a Nexus 5 or emulators). I'm curious what version you were using or doesn't have anything to do with it.

Wanted to point out that I recently refactored to a fragment approach because the client wanted stack like navigation with common ui so the app replaces fragments in container. Even though fragments aren't always a great thing, in this case the memory consumption should be lower due to fragment views being destroyed when replacing them (`onDestroyView()` being called). In this particular case, the resources loaded by the fragment must be cleared `onDestroyView()` or else at some point the memory will be filled if user opens an enormous stack (I know Glide clears resources `onDestroy()`).

sjudd:
[...]

PR #5145, opened 2023-05-18
Allow host app to provide a way to clear all resources onStop()

osamaaftab:
This pull request contain changes which allows host app to provide a way to clear all resources `onStop()` using api `GlideApp.with(this).clearOnStop()`. This is related to issue https://github.com/bumptech/glide/issues/942

### DIFF
--- a/library/src/main/java/com/bumptech/glide/RequestManager.java
+++ b/library/src/main/java/com/bumptech/glide/RequestManager.java
@@ -95,6 +95,8 @@ public class RequestManager
 
   private boolean pauseAllRequestsOnTrimMemoryModerate;
 
+  private boolean clearOnStop;
+
   public RequestManager(
       @NonNull Glide glide,
       @NonNull Lifecycle lifecycle,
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/bumptech/glide/RequestManager.java
```diff
@@ -354,12 +367,19 @@ public synchronized void onStart() {
...
    * Lifecycle callback that unregisters for connectivity events (if the
-   * android.permission.ACCESS_NETWORK_STATE permission is present) and pauses in progress loads.
+   * android.permission.ACCESS_NETWORK_STATE permission is present) and pauses in progress loads
+   * and clears all resources if {@link #clearOnStop()} is called.
    */
   @Override
   public synchronized void onStop() {
     pauseRequests();
     targetTracker.onStop();
+    if (clearOnStop) {
```
sjudd, line 377: Could you extract a clearRequests() method and re-use it here and in onDestroy?

sjudd:
Thanks for implementing this, sorry for the delay! Just one comment.

library/src/main/java/com/bumptech/glide/RequestManager.java
sjudd: Oh and I think you could then do:
```java
if (clearOnStop) {
  clearRequests();
} else { 
  pauseRequests();
}
```

osamaaftab: @sjudd thanks for the suggestion mate.
Please review, I have implemented.

sjudd:
Great thanks!

PR #5146, opened 2023-05-21
Updated docs

pratikbutani:
## Description

Changed `implementation` instead `compile`

## Motivation and Context

It should be updated as `compile` is deprecated


sjudd:
Thanks!

PR #5161, opened 2023-06-05
Add factory method to indicate when executors are not owned by Glide in tests.

paulsowden:
Previously Glide would shutdown all executors when `Glide.teardown` was called, which is also implicitly called when `Glide.init` is called and a Glide has either been previously explicitly or implicitly initialized. Some test environments wrap their own managed executors with `GlideExecutor` and these executors are not always owned by Glide.

### DIFF
--- a/library/test/src/test/java/com/bumptech/glide/load/engine/executor/GlideExecutorTest.java
+++ b/library/test/src/test/java/com/bumptech/glide/load/engine/executor/GlideExecutorTest.java
@@ -3,10 +3,15 @@ package com.bumptech.glide.load.engine.executor;
 import static com.google.common.truth.Truth.assertThat;
 
 import androidx.annotation.NonNull;
+import androidx.test.core.app.ApplicationProvider;
+import com.bumptech.glide.Glide;
+import com.bumptech.glide.GlideBuilder;
+import com.bumptech.glide.util.Executors;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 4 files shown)


sjudd:
```

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':library:test:violations'.
> javax.script.ScriptException: Too many violations found, max is 0 but found 1
  /home/runner/work/glide/glide/library/test/src/test/java/com/bumptech/glide/load/engine/executor/GlideExecutorTest.java
  ╔════════════╤═══════════════════════════════════════════════════════════════════╤══════════╤══════╤══════════════════════════════════════════════════════════╗
  ║ Reporter   │ Rule                                                              │ Severity │ Line │ Message                                                  ║
  ╠════════════╪═══════════════════════════════════════════════════════════════════╪══════════╪══════╪══════════════════════════════════════════════════════════╣
  ║ Checkstyle │ com.puppycrawl.tools.checkstyle.checks.imports.UnusedImportsCheck │ ERROR    │ 15   │ Unused import - java.util.concurrent.ThreadPoolExecutor. ║
  ╚════════════╧═══════════════════════════════════════════════════════════════════╧══════════╧══════╧══════════════════════════════════════════════════════════╝
  ```

File: mocks/src/main/java/com/bumptech/glide/load/engine/executor/MockGlideExecutor.java [hunk 1]
```diff
@@ -17,12 +17,32 @@ private MockGlideExecutor() {
...
+  /**
+   * Wraps an external executor in a {@link GlideExecutor}.
+   *
+   * <p>This should only be used when the executor used in a test is externally managed and will be
+   * shutdown by the calling code or application, Glide will not shutdown the executor when {@link
+   * com.bumptech.glide.Glide#tearDown()} is called. Most users should prefer {@link
+   * #newTestExecutor(ExecutorService)}.
+   */
+  @SuppressWarnings("WeakerAccess")
+  public static GlideExecutor wrapExecutor(ExecutorService executorService) {
```
sjudd, line 42: You could do this via your own ExecutorService wrapper that just ignores shutdown? If it's not a large burden, I'd probably prefer that to expanding the wrapper and shutdown APIs

library/test/src/test/java/com/bumptech/glide/load/engine/executor/GlideExecutorTest.java
```diff
@@ -49,6 +55,46 @@ public void onRun(int priority) {
...
+          new GlideBuilder()
+              .setAnimationExecutor(glideExecutor)
+              .setDiskCacheExecutor(glideExecutor)
+              .setSourceExecutor(glideExecutor));
+
+      Glide.tearDown();
+
+      assertThat(executor.isShutdown()).isFalse();
+    } finally {
+      executor.shutdown();
```
sjudd, line 94: Slightly better to use https://github.com/bumptech/glide/blob/26ecf64550a07d1b8f958a8e3ef192028b4892ca/library/src/main/java/com/bumptech/glide/util/Executors.java#L41 because it will wait if any tasks are scheduled.

Really there shouldn't be anything on the executor, but conceivably there could be some initialization task at some point that leaks across tests via the singleton.

mocks/src/main/java/com/bumptech/glide/load/engine/executor/MockGlideExecutor.java
paulsowden: After trying out your suggestion I learned that this change is actually fundamentally unsafe, if we allow runnables that are submitted by Glide to run after it has awaited termination of its executors then they can run and continue interacting with Glide after it subsequently clears the caches which causes exceptions interacting with locks that have been flushed. So then this pull request shouldn't be accepted, Glide needs to be able to terminate its executors before it continues with teardown.

paulsowden:
See comment https://github.com/bumptech/glide/pull/5161#discussion_r1254988465, should not be merged.

PR #5166, opened 2023-06-07
Remove unnecessary thread checks

MeandNi:
## Description

It seems unnecessary to do thread checking here.

## Motivation and Context

There may be errors when rendering asynchronously.


sjudd:
I don't think it's a good idea to be interacting with views on other threads?

MeandNi:
> I don't think it's a good idea to be interacting with views on other threads?↳

I agree with you, but other similar `into()` methods do not do validation, we can put view-related operations into mainThreadExecutor.

MeandNi:
I think this question can be closed, it's really not a good proposal. @sjudd

PR #5203, opened 2023-07-04
Fix appcompat version to < 1.4

sjudd:
Higher versions require java8

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -16,6 +16,10 @@
     {
       "matchPackageNames": ["androidx.annotation:annotation"],
       "allowedVersions": "<1.6.0"
+    },
+    {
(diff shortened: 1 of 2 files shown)


evelynmarie:
I don't agree with making glide stick to Java 7.

Most Android versions are EOL, and you have to target Android API level 32 (soon 33) to publish in the Play Store anyways, which supports Android 11.

I see no point to keeping support for a significantly old and end of life version of Java.

sjudd:
We will actually get around to updating to java 8. It may be sufficient to upgrade the sample apps to work around this set of errors. If not it'll mean a major version increment for the library

For now though it would be nice to unblock the automation and a lot of new work is happening in compose where were already at higher API levels.

PR #5204, opened 2023-07-04
Add an experimental rememberGlidePainter API

sjudd:
A separate painter can be used outside of GlideImage with other composables. It can also be used for crossfades or custom animations. 

This is very similar in concept to Coil's implementation but the details are a bit different.

### DIFF
--- a/integration/compose/src/main/java/com/bumptech/glide/integration/compose/GlidePainter.kt
+++ b/integration/compose/src/main/java/com/bumptech/glide/integration/compose/GlidePainter.kt
@@ -3,12 +3,16 @@ package com.bumptech.glide.integration.compose
 import android.graphics.drawable.BitmapDrawable
 import android.graphics.drawable.ColorDrawable
 import android.graphics.drawable.Drawable
+import androidx.compose.runtime.Composable
 import androidx.compose.runtime.MutableState
 import androidx.compose.runtime.RememberObserver
 import androidx.compose.runtime.Stable
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
 import androidx.compose.ui.geometry.Size
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.ColorFilter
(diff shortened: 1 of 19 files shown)


integration/compose/src/main/java/com/bumptech/glide/integration/compose/GlidePainter.kt
```diff
@@ -36,54 +45,345 @@ import kotlinx.coroutines.SupervisorJob
...
+ * you want to display the image in or at least the screen size rather than using
+ * [com.bumptech.glide.request.target.Target.SIZE_ORIGINAL].
+ */
+@ExperimentalGlideComposeApi
+@OptIn(InternalGlideApi::class)
+@Composable
+public fun rememberGlidePainter(
+  model: Any?,
+  requestBuilderTransform: RequestBuilderTransform<Drawable> = { it },
+): Pair<Painter, GlidePainterState> {
```
arriolac, line 114: This API is a bit unusual for Compose code. Typically a `remember*` method will return the direct mutable state object, rather than something wrapper in another type such as a `Pair`. For example, I'd expect that this would return a `GlidePainter` object instead. Might be good to create a class like that?

integration/compose/src/main/java/com/bumptech/glide/integration/compose/GlidePainter.kt
```diff
@@ -36,54 +45,345 @@ import kotlinx.coroutines.SupervisorJob
...
+ *
+ * @throws IllegalArgumentException if [requestBuilderTransform] sets an [RequestBuilder.override]
+ * size.
+ */
+@ExperimentalGlideComposeApi
+@OptIn(InternalGlideApi::class)
+@Composable
+public fun rememberGlidePainter(
+  model: Any?,
+  modifier: Modifier,
```
arriolac, line 193: Typically, modifiers are used for UI elements and not in `remember` functions. I don't have a good suggestion here though on how to accomplish what you are trying to do however as I understand that the modifier is the mechanism for determining the image size.

sjudd: One other option is extracting a method that deals with modifiers and then having people pass in ResolvableGlideSize to rememberGlidePainter instead:

```kotlin
  val (size, updatedModifier) = waitForSizeFromModifier(modifier);
  val painter = rememberGlidePainter(model, size);
  Image(painter, contentDescription, updatedModifier)
```

Maybe that's a bit nicer?

integration/compose/src/main/java/com/bumptech/glide/integration/compose/GlidePainter.kt
sjudd: I think the disadvantage is that you expose a broader API surface, you get all of GlidePainter when all you need is a state.

I could extract an interface to avoid that:
[
```kotlin 
GlidePainter extends Painter(public val state: GlidePainterState) { }

val painter = rememberGlidePainter(model)
match (painter.state) {
...
}
```

It feels slightly less nice to return a Painter interface that's just Painter + State? That is the way that Coil does this though.

integration/compose/src/main/java/com/bumptech/glide/integration/compose/GlidePainter.kt
sjudd: Done

integration/compose/src/main/java/com/bumptech/glide/integration/compose/GlidePainter.kt
sjudd: Done

sjudd:
Replaced with https://github.com/bumptech/glide/pull/5230

PR #5207, opened 2023-07-06
Introduce a constant for Robolectric tests using sdk = 18.

brettchabot:
This commit replaces hardcoded instances of 'Config(sdk = 18') with a constant. The intent is to make it easier to update the value in the future when Robolectric stops supporting sdk 18.

Tested via gradle test testDebugUnitTest.

### DIFF
--- a/library/test/src/test/java/com/bumptech/glide/GlideTest.java
+++ b/library/test/src/test/java/com/bumptech/glide/GlideTest.java
@@ -1,5 +1,6 @@
 package com.bumptech.glide;
 
+import static com.bumptech.glide.RobolectricConstants.ROBOLECTRIC_SDK;
 import static com.bumptech.glide.request.RequestOptions.decodeTypeOf;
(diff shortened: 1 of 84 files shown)


sjudd:
Thanks!

library/test/src/test/java/com/bumptech/glide/GlideTest.java
```diff
@@ -94,7 +95,7 @@
 @LooperMode(LEGACY)
 @RunWith(RobolectricTestRunner.class)
 @Config(
-    sdk = 18,
+    sdk = ROBOLECTRIC_SDK,
```
utzcoz, line 98: @brettchabot Hi Brett, why not using property file for it?

brettchabot: I didn't want to introduce a behavior change for tests that currently don't specify a @config(sdk)

However, it's a good suggestion that @sjudd might want to consider as a future change.

PR #5208, opened 2023-07-07
Increment ROBOLECTRIC_SDK to 19 from 18.

brettchabot:
A future version of Robolectric will drop support for SDK 18.

Tested via gradle test testDebugUnitTest.


sjudd:
Thanks again!

PR #5220, opened 2023-07-26
Add @RequiresPermission to NotificationTarget

TWiStErRob:
## Description
Add `RequiresPermission` to help users and tools recognize that this class needs permissions. Ideally it would be just one `@RequiresPermission(Manifest.permission.POST_NOTIFICATIONS)` on the class, but `androidx.annotation.RequiresPermission` is not applicable there (even though lint checks for it 😕). I checked androidx.annotation 1.6.0.

`InlinedApi` suppression is necessary because minSdk=14, compileSdk=33. The constant is new. Could inline it instead, but I personally don't like hard-coded copy-pasted constants when there's a named one.

## Motivation and Context
Motivation is this lint error:
image

It alerts when Glide is on the classpath, even if this class is not used. Noticed on AGP 8.

I read the code of `NotificationPermissionDetector` and noticed that it wouldn't alert if the class asked for the permission, that's why I decided it might be a good solution to add it, and fix everyone's lint at once :)

### DIFF
--- a/library/src/main/java/com/bumptech/glide/request/target/NotificationTarget.java
+++ b/library/src/main/java/com/bumptech/glide/request/target/NotificationTarget.java
@@ -1,5 +1,7 @@
 package com.bumptech.glide.request.target;
 
+import android.Manifest;
+import android.annotation.SuppressLint;
 import android.app.Notification;
 import android.app.NotificationManager;
 import android.content.Context;
@@ -8,6 +10,7 @@ import android.graphics.drawable.Drawable;
 import android.widget.RemoteViews;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.annotation.RequiresPermission;
 import com.bumptech.glide.request.transition.Transition;
 import com.bumptech.glide.util.Preconditions;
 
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/bumptech/glide/request/target/NotificationTarget.java
```diff
@@ -61,6 +67,9 @@ public NotificationTarget(
    * @param notificationTag The notificationTag of the Notification that we want to load the Bitmap.
    *     May be {@code null}.
    */
+  @SuppressLint("InlinedApi")
```
sjudd, line 70: Why do you need to suppress inlined API?

sjudd:
Just one question, otherwise looks good. Thanks!

library/src/main/java/com/bumptech/glide/request/target/NotificationTarget.java
TWiStErRob: See PR OP.

sjudd:
Thanks!

TWiStErRob:
If anyone finds this from Google's Places SDK (4.2.0), here's the issue to track a "fix":
https://issuetracker.google.com/issues/412011932

---

@sjudd I thought the Google monorepo uses the latest of Glide, or not all of the projects?

PR #5232, opened 2023-08-07
Add a isEquivalentTo method to correctly check equality

### LINKED ISSUE #4958: ImageView briefly losing the image only on 4.14.2 (produces a flicker)
**Glide Version**: 4.14.2

**Integration libraries**: OkHttp 4.9.0

**Device/Android Version**: Reproducible in any device

Since we've upgraded to 4.14.2, we've been having an issue where, briefly, the `ImageView` loses the image. This will make its size be 0x0 (we're using `wrap_content` for this specific view) and it'll produce a flicker.

I recorded a video highlighting the problem:
https://user-images.githubusercontent.com/67684386/202150949-fd0cc824-b35f-4ae9-9a91-ac359e171958.mp4

The video shows that we had no image (haven't been loaded yet), then it loads and show, then it disappears, then it shows again.

We're loading the image in a `RecyclerView` (no compose at all here) and there's multiple rebinds happening at that time, so it could be related.

From debugging we could track this down into loading the image with properties on the `RequestBuilder` that are actually a new instance. Such as:

`.transition(DrawableTransitionOptions.withCrossFade(crossFadeFactory))`
or
`.listener(listener)`

Not calling those methods on `RequestBuilder` fixes the problem.

A very important detail is that this does not happen on 4.14.1, which makes me very suspicious of this commit https://github.com/bumptech/glide/commit/f3d6ff7603e4759a89ba7bd9a8b1a77281e3a4b0. In the meantime we had to rollback to this version.

```kotlin
requestBuilder
    .listener(listener)
    .transition(DrawableTransitionOptions.withCrossFade(crossFadeFactory))
    .skipMemoryCache(config.skipMemoryCache)
    .diskCacheStrategy(config.cacheStrategy)
    .into(this)
```

This code will be called multiple times in the rebind of a `ViewHolder`.

**Layout XML**:
```xml
<ImageView
        android:id="@+id/pdp_title_logo_img"
        android:layout_width="wrap_content"
        android:layout_height="wrap_content"
        android:layout_marginTop="@dimen/pdp_title_margin_vertical"
        android:adjustViewBounds="true"
        android:maxWidth="@dimen/pdp_title_frame_width"
        android:maxHeight="@dimen/pdp_title_frame_height"
        android:scaleType="fitStart"
        android:visibility="gone"
        app:layout_constraintDimensionRatio="1"
        app:layout_constraintEnd_toEndOf="parent"
        app:layout_constraintHorizontal_bias="0"
        app:layout_constraintStart_toStartOf="@id/guide_start"
        app:layout_constraintTop_toBottomOf="@id/pdp_channel_logo_img"
        tools:srcCompat="@drawable/ic_app_logo_small"
        tools:visibility="visible" />
```

mori-atsushi:
## Description

There was a change in equals method on Glide 4.14.2 (f3d6ff7).
After that, the cache doesn't work in several cases.
This PR fixes the bug by adding a `isEquivalentTo` method.

close #4958

## Motivation and Context

See [my comment](https://github.com/bumptech/glide/issues/4958#issuecomment-1630408035) in the Issue.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/request/BaseRequestOptions.java
+++ b/library/src/main/java/com/bumptech/glide/request/BaseRequestOptions.java
@@ -1209,31 +1209,42 @@ public abstract class BaseRequestOptions<T extends BaseRequestOptions<T>> implem
     return selfOrThrowIfLocked();
   }
 
+  /**
+   * Returns {@code true} if this {@link BaseRequestOptions} is equivalent to the given
+   * {@link BaseRequestOptions} (has all of the same options and sizes).
+   *
+   * <p>This method is identical to {@link #equals(Object)}, but this can not be overridden. We need
+   * to use this method instead of {@link #equals(Object)}, because child classes may have additional
+   * fields, such as listeners and models, that should not be considered when checking for equality.
+   */
+  public final boolean isEquivalentTo(BaseRequestOptions<?> other) {
+    return Float.compare(other.sizeMultiplier, sizeMultiplier) == 0
+        && errorId == other.errorId
+        && Util.bothNullOrEqual(errorPlaceholder, other.errorPlaceholder)
+        && placeholderId == other.placeholderId
+        && Util.bothNullOrEqual(placeholderDrawable, other.placeholderDrawable)
+        && fallbackId == other.fallbackId
+        && Util.bothNullOrEqual(fallbackDrawable, other.fallbackDrawable)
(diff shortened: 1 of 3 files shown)


sjudd:
Thanks, I really appreciate you digging in to the issue and sending the PR!

mori-atsushi:
@sjudd Thank you! When will this be released?

PR #5235, opened 2023-08-14
Add a Transition API and a CrossFade Transition for Compose

sjudd:
Similar to Glide's existing Transitions, but intended for use with Animatable instead of views. I've started with a simple CrossFade. We probably should add at least a FadeOnTopOf (or something similar) to mimic the existing Crossfade support.

### DIFF
--- a/integration/compose/api/compose.api
+++ b/integration/compose/api/compose.api
@@ -1,8 +1,21 @@
+public final class com/bumptech/glide/integration/compose/CrossFade : com/bumptech/glide/integration/compose/Transition$Factory {
+	public static final field $stable I
+	public static final field Companion Lcom/bumptech/glide/integration/compose/CrossFade$Companion;
+	public fun <init> (Landroidx/compose/animation/core/AnimationSpec;)V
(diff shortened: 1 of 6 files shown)


sjudd:
@lelandrichardson if you have any thoughts here they'd be welcome.

sjudd:
Assuming this pattern is ok: https://github.com/bumptech/glide/pull/5237, I might be able to extract the animation portion into another modifier node.

sjudd:
> Assuming this pattern is ok: #5237, I might be able to extract the animation portion into another modifier node.

Ok so I decided not to do this. Transitions are pretty concerned not just with the state of the request, but the presence / behavior of the placeholder, so I think they're a bit of a special case. We could still use this listener to compose a modifier that animates the layout, or even the combined placeholder + thumbnail(s) + main request during draw. But it's not very practical to use to animate between a placeholder and the finished request. 

I suppose another option would be to extract the request, drawing the placeholder, and drawing the full image to separate nodes, but that's getting a bit complicated.

PR #5257, opened 2023-08-23
Fix Maven Central badge

Goooler:
[Maven Central](https://search.maven.org/artifact/com.github.bumptech.glide/glide)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 Glide
 =====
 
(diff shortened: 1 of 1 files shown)


sjudd:
What's wrong with the current badge?

Goooler:
I see there is an "unknown" shown on its version. Alright.

sjudd:
Probably just a bit of delay? Thanks for putting this up, if you think it's better to use this version, let me know. Otherwise I'm inclined to keep the status quo. All of these badge sites seem mildly sketchy...

PR #5279, opened 2023-09-14
Finish animation ignoring cancellation.

### LINKED ISSUE #5278: Transition.kt invokes cancellable suspend function in finally clause
https://github.com/bumptech/glide/blob/cbdc1e78b4ee0f335054bb2781f46cf8db82f914/integration/compose/src/main/java/com/bumptech/glide/integration/compose/Transition.kt#L144

`animatable.snapTo(OPAQUE_ALPHA)` is suspending, and should not be invoked in the finally clause, since if the coroutine has been cancelled, any call to a suspending function will immediately throw CancellationException, thus not performing the final animation move cleanup.

This can be fixed by invoking the suspending function non-cancellable:
```
  withContext(NonCancellable) {
    animatable.snapTo(OPAQUE_ALPHA)
  }
```

kanelbulle:
Fixes #5278 

Invokes the animation finish non-cancellable in case the coroutine has been cancelled. Without it I think there's a risk we might leave something half-animated. Even if that can't practically happen, I believe this is best practice.


sjudd:
Thanks!

PR #5296, opened 2023-09-26
Add a Painter variant of the placeholder APIs

### LINKED ISSUE #5224: Compose GlideImage set loading placeholder with @Composable Painter  bugs
**Glide Version**: 4.15.1

**Integration libraries**: 
com.github.bumptech.glide:compose:1.0.0-alpha.3

When setting a placeholder of type Painter, LazyColumn scrolling through the list triggers image redraws, causing visual flickering of the image displayed on the screen.
If a ColorDrawable is set as the placeholder, the display appears normal without any flickering issues.

```
@OptIn(ExperimentalGlideComposeApi::class)
@Composable
actual fun PlatformImage(
    model: Any?,
    contentDescription: String?,
    modifier: Modifier,
    alignment: Alignment,
    contentScale: ContentScale,
    alpha: Float,
    colorFilter: ColorFilter?,
    filterQuality: FilterQuality,
    placeholderPainter: Painter?
) {
    GlideImage(
        model = model,
        contentDescription = null,
        modifier = modifier,
        alignment = alignment,
        alpha = alpha,
        colorFilter = colorFilter,
        contentScale = contentScale,
//        loading = placeholder(ColorDrawable(0xff000000.toInt()))
    loading = placeholder { placeholderPainter }
    )
}
```

sjudd:
Fixes #5224

We could go further and remove the drawable / resource id variants entirely, but it won't lead to a huge simplification internally because we still have to deal with those types from RequestBuilder.

### DIFF
--- a/integration/compose/src/main/java/com/bumptech/glide/integration/compose/GlideModifier.kt
+++ b/integration/compose/src/main/java/com/bumptech/glide/integration/compose/GlideModifier.kt
@@ -65,7 +65,7 @@ import kotlin.math.roundToInt
 
 @ExperimentalGlideComposeApi
 internal interface RequestListener {
-  fun onStateChanged(model: Any?, drawable: Drawable?, requestState: RequestState)
+  fun onStateChanged(model: Any?, painter: Painter?, requestState: RequestState)
 }
 
 @ExperimentalGlideComposeApi
@@ -79,6 +79,8 @@ internal fun Modifier.glideNode(
   transitionFactory: Transition.Factory? = null,
   requestListener: RequestListener? = null,
   draw: Boolean? = null,
+  loadingPlaceholder: Painter? = null,
+  errorPlaceholder: Painter? = null,
 ): Modifier {
   return this then GlideNodeElement(
     requestBuilder,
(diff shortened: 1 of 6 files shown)


integration/compose/src/main/java/com/bumptech/glide/integration/compose/GlideModifier.kt
```diff
@@ -65,7 +65,7 @@ import kotlin.math.roundToInt
 
 @ExperimentalGlideComposeApi
 internal interface RequestListener {
-  fun onStateChanged(model: Any?, drawable: Drawable?, requestState: RequestState)
+  fun onStateChanged(model: Any?, painter: Painter?, requestState: RequestState)
```
Anuwatrodcharoen, line 68: fun onStateChanged(model: Any?, painter: Painter?, requestState: RequestState)

Anuwatrodcharoen:
> Fixes #5224
> 
> We could go further and remove the drawable / resource id variants entirely, but it won't lead to a huge simplification internally because we still have to deal with those types from RequestBuilder.

PR #5306, opened 2023-10-02
Make Placeholder a type that can be convered to Painter

sjudd:
This removes the deprecated Composable variant of Placeholder. Users who require that functionality should instead use GlideSubcomposition. It's useful to remove this now because it allows for a much simpler Placholder interface and it unblocks us from moving most of GlideImage into GlideModifier. In turn that will let us expose GlideModifier as a real composable API component.

### DIFF
--- a/integration/compose/src/androidTest/java/com/bumptech/glide/integration/compose/GlideImageCustomDrawableTransformationTest.kt
+++ b/integration/compose/src/androidTest/java/com/bumptech/glide/integration/compose/GlideImageCustomDrawableTransformationTest.kt
@@ -34,7 +34,8 @@ class GlideImageCustomDrawableTransformationTest(
   // though it's only used by Parameters to create the test name.
   @Suppress("unused") private val name: String,
 ) {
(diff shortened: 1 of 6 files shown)


sjudd:
@kanelbulle can you let me know how painful removing the subcomposition variants of placeholder would be? I could keep them around for a while longer as special cases, but it would be nice to remove them.

sjudd:
@kanelbulle let me know what you think

kanelbulle:
Thanks for checking, I think we can remove it, but it would be nice to have it @Deprecated for a month or two first if it's not too much trouble. We have less than 10 usages in a couple of different projects so it's not completely trouble-free.

sjudd:
> Thanks for checking, I think we can remove it, but it would be nice to have it @deprecated for a month or two first if it's not too much trouble. We have less than 10 usages in a couple of different projects so it's not completely trouble-free.

Ok it turns out it's kind of annoying to get it to work again with the refactor... so I've left the API in but it doesn't work. I'm a real Android engineer now right? 

At least it'll compile when you import it

Issue #5211, opened 2023-07-13
Glide v4.15.1  java.lang.IllegalStateException: Cannot recycle a resource while it is still acquired


hpdx:
com.bumptech.glide.load.engine.EngineResource.recycle

Fatal Exception: java.lang.IllegalStateException: Cannot recycle a resource while it is still acquired
       at com.bumptech.glide.load.engine.EngineResource.recycle(EngineResource.java:68)
       at com.bumptech.glide.load.engine.ResourceRecycler$ResourceRecyclerCallback.handleMessage(ResourceRecycler.java:37)
       at android.os.Handler.dispatchMessage(Handler.java:102)
       at android.os.Looper.loopOnce(Looper.java:223)
       at android.os.Looper.loop(Looper.java:324)
       at android.app.ActivityThread.main(ActivityThread.java:8464)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:582)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1059)

lyjSmile:
I hava same probleam, did you solve it?

java.lang.IllegalStateException: Cannot recycle a resource while it is still acquired
       at com.bumptech.glide.load.engine.EngineResource.recycle(EngineResource.java:68)
       at com.bumptech.glide.load.engine.ResourceRecycler$ResourceRecyclerCallback.handleMessage(ResourceRecycler.java:37)
       at android.os.Handler.dispatchMessage(Handler.java:102)
       at android.os.Looper.loop(Looper.java:250)
       at android.app.ActivityThread.main(ActivityThread.java:7851)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:592)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:958)

TWiStErRob:
Duplicate of https://github.com/bumptech/glide/issues/4507, please "Subscribe" to that issue.

Issue #4507, opened 2021-02-23
Fatal Exception: java.lang.IllegalStateException


ayyaztalat:
I am currently using Glide version : 4.11.0
implementation 'com.github.bumptech.glide:glide:4.11.0'
annotationProcessor 'com.github.bumptech.glide:compiler:4.11.0'

Cannot recycle a resource while it is still acquired
com.bumptech.glide.load.engine.EngineResource.recycle (EngineResource.java:68)
com.bumptech.glide.load.engine.ResourceRecycler$ResourceRecyclerCallback.handleMessage (ResourceRecycler.java:37)
android.os.Handler.dispatchMessage (Handler.java:103)
android.os.Looper.loop (Looper.java:227)
android.app.ActivityThread.main (ActivityThread.java:7822)
java.lang.reflect.Method.invoke (Method.java)
com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:492)
com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1026)

mkocus:
We've got exactly the same issue.
It seems that the glide crashes randomly on production - this issue cannot be replicated locally.

Full stack log:
```
Fatal Exception: java.lang.IllegalStateException: Cannot recycle a resource while it is still acquired
       at com.bumptech.glide.load.engine.EngineResource.recycle(EngineResource.java:68)
       at com.bumptech.glide.load.engine.ResourceRecycler$ResourceRecyclerCallback.handleMessage(ResourceRecycler.java:37)
       at android.os.Handler.dispatchMessage(Handler.java:102)
       at android.os.Looper.loop(Looper.java:193)
       at android.app.ActivityThread.main(ActivityThread.java:6680)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
```

Environment:
- Android 9
- Glide 4.12.0

Do you have any ideas how to debug this issue?

ayyaztalat:
For my case i have invalidate my current bitmap clear glide cache and then
reassigned bitmap thus it solved the problem but the main problem it
arrose. It slowed down the loading of receyclerview images

mkocus:
Well, we are using glide a lot and cannot locate where precisely this crash happen :/
[...]

PR #5310, opened 2023-10-12
Proof of concept for investigation of "Cannot recycle a resource while it is still acquired"

TWiStErRob:
## Description
Trace asynchronous calls across Handler messages. It's probably not acceptable in its current form due to many allocations. I wonder how this could make it to production. Is there a way to "debug" Glide? Feature flag?

## Motivation and Context
https://github.com/bumptech/glide/issues/4507 and https://github.com/bumptech/glide/issues/5211 have no idea why it's happening, this would highlight which part of the app is causing the issue.


sjudd:
It could be done with a default off log tag that's checked once statically? You could avoid capturing the stack trace unless it's enabled.

TWiStErRob:
I think a Log tag is not enough, they need to deploy it to production so that they can get crash traces from real devices. And they [can't execute `adb shell setprop`](https://stackoverflow.com/a/11623309/253468) there It's not reproducible in dev envs as I understood. Glide users need to control this explicitly to opt in.

TWiStErRob:
Although a simple static boolean variable would work, I guess.

sjudd:
Hmm for production use cases we could use the flags system, might be annoying to pipe it through, but once done you can read from some set of flag values during initialization to control this: https://github.com/bumptech/glide/blob/75599b7af05eeca1393c8f4267adb78e88e7821e/library/src/main/java/com/bumptech/glide/GlideBuilder.java#L452

Issue #5308, opened 2023-10-09
Compose crash


Caij:
* version beta01

crash on HorizontalPager  change page

```
fun PlayPageContain(
    rememberHorizontalPagerState: PagerState
) {
    HorizontalPager(state = rememberHorizontalPagerState) { page: Int ->
        if (page == 1) {
            PlayerPage(rememberHorizontalPagerState)
        } else if (page == 0) {
            SimplePlayingQueuePage(rememberHorizontalPagerState)
        } else {
            SimpleCurrentLyricsDetailPage(rememberHorizontalPagerState)
        }
    }
}

@Composable
fun PlayerPage(
    rememberPagerState: PagerState
) {
    val playerComponent by remember { mutableStateOf(PlayerComponent()) }
    val song by playerComponent.model.subscribeAsState()
    Column(modifier = Modifier.fillMaxSize()) {
        val painter = painterResource(MR.images.default_audio_art)
        Box {
            PlatformSongAsyncImage(song, modifier = Modifier.padding(top = getSafeTop()).padding(16.dp)
                .fillMaxWidth().aspectRatio(1f).clip(Colors.songRoundedCornerShape),
                contentDescription = null, placeholderPaint = painter)

            val isBuffing by playerComponent.isBuffing.subscribeAsState()
            if (isBuffing) {
                CircularProgressIndicator(modifier = Modifier.width(36.dp).height(36.dp).align(
                    Alignment.Center
                ))
            }
        }

        PlayerControlWidget(playerComponent, song, Modifier.weight(1f), rememberPagerState)

        BottomMenu(playerComponent, rememberPagerState)
    }

    DisposableEffect(playerComponent) {
        playerComponent.uiRun()
        onDispose {
            playerComponent.uiDispose()
        }
    }
}
```

![image](https://github.com/bumptech/glide/assets/8597839/a2c9ec95-4a94-4cee-9458-8eea0471162c)

java.lang.IllegalArgumentException: 
                                                                                                    	at com.bumptech.glide.util.Preconditions.checkArgument(Preconditions.java:21)
                                                                                                    	at com.bumptech.glide.util.Preconditions.checkArgument(Preconditions.java:16)
                                                                                                    	at com.bumptech.glide.integration.compose.GlideNode$launchRequest$1.invoke(GlideModifier.kt:404)
                                                                                                    	at com.bumptech.glide.integration.compose.GlideNode$launchRequest$1.invoke(GlideModifier.kt:400)
[...]

PR #5313, opened 2023-10-17
Ignore rather than crash on duplicate requests in GlideModifier

sjudd:
Speculative fix for #5308


cameronhenige:
Hey @sjudd It has been a while since this has been merged in. Are there any plans to do an official release? It looks like the compose library hasn't been updated since Sept 26, 2023.

PR #5341, opened 2023-12-04
Support tint list on GifDrawable.

sumita12:
## Description
This PR implements `GifDrawable#setTintList()` and `GifDrawable#setTintMode()` to support tint list (e.g. android:tint).

## Motivation and Context
Our Android app uses a tint list to change the drawable colors, but GifDrawable doesn't support tint thus it doesn't work well.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/load/resource/gif/GifDrawable.java
+++ b/library/src/main/java/com/bumptech/glide/load/resource/gif/GifDrawable.java
@@ -3,12 +3,16 @@ package com.bumptech.glide.load.resource.gif;
 import static com.bumptech.glide.gifdecoder.GifDecoder.TOTAL_ITERATION_COUNT_FOREVER;
 
 import android.content.Context;
+import android.content.res.ColorStateList;
 import android.content.res.Resources;
 import android.graphics.Bitmap;
 import android.graphics.Canvas;
+import android.graphics.Color;
 import android.graphics.ColorFilter;
 import android.graphics.Paint;
 import android.graphics.PixelFormat;
+import android.graphics.PorterDuff;
+import android.graphics.PorterDuffColorFilter;
 import android.graphics.Rect;
 import android.graphics.drawable.Animatable;
 import android.graphics.drawable.Drawable;
@@ -66,6 +70,9 @@ public class GifDrawable extends Drawable
 
   private boolean applyGravity;
   private Paint paint;
+  private ColorStateList tint;
+  private PorterDuff.Mode tintMode;
+  private ColorFilter tintFilter;
   private Rect destRect;
 
   /** Callbacks to notify loop completion of a gif, where the loop count is explicitly specified. */
@@ -288,7 +295,17 @@ public class GifDrawable extends Drawable
     }
 
     Bitmap currentFrame = state.frameLoader.getCurrentFrame();
-    canvas.drawBitmap(currentFrame, null, getDestRect(), getPaint());
+    Paint paint = getPaint();
+    ColorFilter colorFilter = paint.getColorFilter();
+    if (colorFilter != null || tintFilter == null) {
+      // ColorFilter disables tint list. See Drawable#setColorFilter().
+      canvas.drawBitmap(currentFrame, null, getDestRect(), paint);
+    } else {
+      // Temporary set a tint filter then restore.
+      paint.setColorFilter(tintFilter);
+      canvas.drawBitmap(currentFrame, null, getDestRect(), paint);
+      paint.setColorFilter(colorFilter);
+    }
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


library/src/main/java/com/bumptech/glide/load/resource/gif/GifDrawable.java
```diff
@@ -298,7 +315,47 @@ public void setAlpha(int i) {
 
   @Override
   public void setColorFilter(ColorFilter colorFilter) {
-    getPaint().setColorFilter(colorFilter);
+    if (getColorFilter() != colorFilter) {
```
paulsowden, line 318: why use reference equality instead of Objects.equals?

and why does this specific method check for equality but the other methods not? I guess this is trying to handle the case where colorFilter is null to prevent it from clearing the tint? If that's the case please can you make the code explicit (preferred) or add a comment to say it is, because otherwise it's not at all obvious from the code that this is the intent.

library/src/main/java/com/bumptech/glide/load/resource/gif/GifDrawable.java
```diff
@@ -66,6 +70,9 @@ public class GifDrawable extends Drawable
 
   private boolean applyGravity;
   private Paint paint;
+  private ColorStateList tint;
+  private PorterDuff.Mode tintMode;
```
paulsowden, line 74: Default tintMode is documented as SRC_IN

https://developer.android.com/reference/android/graphics/drawable/Drawable#setTintMode(android.graphics.PorterDuff.Mode)

The Android Drawable implementations also imply that this should not be settable to null but null should set back to the default (this behavior is not documented and is not marked as nullable, but the base Drawable compat implementation treats null as default values)

library/test/src/test/java/com/bumptech/glide/load/resource/gif/GifDrawableTest.java
```diff
@@ -568,12 +569,53 @@ public void testSetAlphaSetsAlphaOnPaint() {
...
+            new int[][] {new int[] {android.R.attr.state_pressed}, new int[0]},
+            new int[] {Color.RED, Color.GREEN});
+    drawable.setTintList(tint);
+    drawable.setTintMode(Mode.ADD);
+    when(paint.getColorFilter()).thenReturn(null);
+    drawable.draw(new Canvas());
+
+    // draw() temporary sets tint filter then restore.
+    verify(paint).setColorFilter(eq(new PorterDuffColorFilter(Color.GREEN, Mode.ADD)));
+    verify(paint).setColorFilter(null);
```
paulsowden, line 589: this test is not testing drawable behavior but internal implementation details of the draw method, what we care about is that the bitmap was drawn with a paint configured with the tint filter, not that paint was interacted with.

library/test/src/test/java/com/bumptech/glide/load/resource/gif/GifDrawableTest.java
```diff
@@ -568,12 +569,53 @@ public void testSetAlphaSetsAlphaOnPaint() {
...
+
+    assertThat(drawable.setState(new int[] {android.R.attr.state_pressed})).isTrue();
+    drawable.draw(new Canvas());
```
paulsowden, line 592: performing multiple acts in a test implies that there should be two tests, please split this into a separate test that verifies the state is reflected in the filter.

paulsowden (changes requested):

library/src/main/java/com/bumptech/glide/load/resource/gif/GifDrawable.java
sumita12: I added the check because GradientDrawable does the same thing.
However, I think it is not directly related to this PR, so please let me revert this change to simplify this PR.

library/src/main/java/com/bumptech/glide/load/resource/gif/GifDrawable.java
sumita12: > null should set back to the default

You are right. I updated this PR to use the default mode if null.

> The Android Drawable implementations also imply that this should not be settable to null

FYI, Drawable#setTintMode() accept null.
https://cs.android.com/android/platform/superproject/+/master:frameworks/base/graphics/java/android/graphics/drawable/Drawable.java;l=692?q=drawable.java

library/test/src/test/java/com/bumptech/glide/load/resource/gif/GifDrawableTest.java
sumita12: Thank  you for pointing it out!
Previously I followed the test code of setColorFilter().

Now, I updated the test code of both setColorFilter() and setTintList() to check the Paint object which is used by Canvas#drawBitmap().
PTAL?

(Ideally it's better to check the drawn bitmap colors, but we cannot do it on Robolectric tests.)

library/test/src/test/java/com/bumptech/glide/load/resource/gif/GifDrawableTest.java
sumita12: Done.

library/test/src/test/java/com/bumptech/glide/load/resource/gif/GifDrawableTest.java
sumita12: Note:
I updated the behavior of GifDrawable#draw() to make it test-friendly.

We had the following code inside GifDrawable#draw().

>  // Temporary set a tint filter then restore.
> paint.setColorFilter(tintFilter);
> canvas.drawBitmap(currentFrame, null, getDestRect(), paint);
> paint.setColorFilter(colorFilter);

With this code, we cannot verify whether `canvas#drawBitmap()` uses the `tintFilter` or not with ArgumentCaptor, because `paint.setColorFilter(colorFilter)` is called before `ArgumentCaptor#capture()`. The new code doesn't call `paint.setColorFilter(colorFilter)` to avoid the issue.

Issue #5328, opened 2023-11-08
@GlideModule annotated classes must implement AppGlideModule


manishindiainc:
I have started migration of my existing Android app from `kapt` to `ksp` and getting an error

`[ksp] @GlideModule annotated classes must implement AppGlideModule or LibraryGlideModule: ` 

My Glide Module structure looks like this
```kotlin
// code in my app
@GlideModule
class GlideAppModule : MyGlideModule() {
    @Inject lateinit var modelLoaderFactory: GlideUrlLoaderFactory

    override fun registerComponents(context: Context, glide: Glide, registry: Registry) {
        super.registerComponents(context, glide, registry)
        // some specific registration
    }
}

// class from platform library, difficult to change
abstract class MyGlideModule : AppGlideModule() {

    override fun registerComponents(context: Context, glide: Glide,
                                    registry: Registry) {
	 // registration common to platform
     }

    override fun applyOptions(context: Context, builder: GlideBuilder) {
        super.applyOptions(context, builder)
       // set disk cache
    }
}
```
The `MyGlideModule` class is from a platform library and is difficult to change. The `GlideAppModule` is defined in my Android library module `app/impl`

Now, this inheritance was working perfectly fine with `kapt` but with `ksp`, started getting this error

```
com.github.bumptech.glide:ksp:4.16.0 // also tried with 4.14.2 and 4.15.1, same error

kotlin version: 1.9.0
```
Any thought would be great help.

manishindiainc:
Team, could you please help with this

harshitAtlassian:
Team, I'm facing this similar issue. Please provide any solution.

manishindiainc:
Team, blocked on this, please provide a solution.

nirazv:
i am also facing the same issue.

trietbui85:
Any update?

PR #5344, opened 2023-12-28
Support multiple level inheritance for lib and app modules in case of KSP usage

0xnm:
## Description

In Datadog we [have instrumentation](https://github.com/DataDog/dd-sdk-android/blob/ab6f38f5b8ddf8d9f500d4043ef8ac77b2cf8ca1/integrations/dd-sdk-android-glide/src/main/kotlin/com/datadog/android/glide/DatadogGlideModule.kt) for Glide by the means of creating `DatadogGlideModule` which inherits from `AppGlideModule`. Customers can extend `DatadogGlideModule` instead of `AppGlideModule` and get their Glide calls instrumented with Datadog.

This worked fine for KAPT, but doesn't work for KSP, because in case of KSP usage compilation with some class which inherits `DatadogGlideModule` instead of `AppGlideModule` directly gives the same error as in https://github.com/bumptech/glide/issues/5328.

This happens because `supertypes` call [here](https://github.com/bumptech/glide/blob/e77562a432352da19741d93b8c1f6599b558676a/annotation/ksp/src/main/kotlin/com/bumptech/glide/annotation/ksp/ModuleParser.kt#L26) checks only types at declaration site and doesn't go through the whole inheritance chain up.

This PR fixes that issue by traversing the whole inheritance chain, so that functionality is aligned with KAPT compiler (which permits such inheritance chains).

**NB**: Since we have to traverse the whole inheritance chain and `AppGlideModule` extends actually `LibraryGlideModule`, it may be tricky to not register app module as library module as well, but since this PR is using `first` item in the supertype match, app module will be still registered as app module and not as both.

### DIFF
--- a/annotation/ksp/integrationtest/src/test/java/com/bumptech/glide/annotation/ksp/integrationtest/IntegrationLibraryGlideModuleTests.kt
+++ b/annotation/ksp/integrationtest/src/test/java/com/bumptech/glide/annotation/ksp/integrationtest/IntegrationLibraryGlideModuleTests.kt
@@ -57,6 +57,53 @@ class IntegrationLibraryGlideModuleTests(override val sourceType: SourceType) : 
   }
 
   @Test
+  fun compile_withOnlyAppGlideModuleThroughBaseClass_generatesGeneratedAppGlideModule_thatCallsDependencyLibraryGlideModules() {
+    val kotlinAppModule =
+      KotlinSourceFile(
+        "AppModule.kt",
+        """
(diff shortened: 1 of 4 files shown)


sjudd:
Thanks for the fix, I really appreciate the thorough tests!od

0xnm:
Hi @sjudd! I had to sync the branch with main for the auto-merge, but it seems workflow needs your approval.

manishindiainc:
This is awesome. Do we have any idea when the next version containing this fix will be released?

PR #5357, opened 2024-01-24
Add workaround for AOSP bug with decoding single channel hardware gai…

falhassen:
…nmaps using BitmapFactory.

Android U devices that use skiagl for hwui renderering (currently the vast majority) have a bug where uplading single channel bitmaps to hardware bitmaps fails. This bug essentially breaks single channel gainmap rendering using hardware bitmaps, which is a common use case on devices that have Ultra HDR as the default capture format.

The workaround is as follows:
1. Detect the presence of the bug by attempting to copy a 1x1 single channel bitmap to hardware.
2. Memoize the result
3. For BitmapFactory.decode* calls, if the input bitmap has a gainmap, copy it to a three-channel gainmap. This step seems wasteful, but in practice, the Android OS already does some copying when creating hardware bitmaps. Also, more importantly, there is an AOSP fix for this bug, and as devices are updated with this fix, this flow will no longer be exercised.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/GlideBuilder.java
+++ b/library/src/main/java/com/bumptech/glide/GlideBuilder.java
@@ -500,6 +500,17 @@ public final class GlideBuilder {
   }
 
   /**
+   * Fixes decoding of hardware gainmaps from Ultra HDR images on Android U.
+   *
+   * <p>Without this flag on, gainmaps may be dropped when decoding Ultra HDR on Android U devices
+   * using skiagl for hwui as described in https://github.com/bumptech/glide/issues/5362.
+   */
+  public GlideBuilder setEnableHardwareGainmapFixOnU(boolean isEnabled) {
+    glideExperimentsBuilder.update(new EnableHardwareGainmapFixOnU(), isEnabled);
+    return this;
+  }
+
+  /**
    * @deprecated This method does nothing. It will be hard coded and removed in a future release
    *     without further warning.
    */
@@ -617,6 +628,9 @@ public final class GlideBuilder {
    * {@link com.bumptech.glide.load.resource.bitmap.TransformationUtils}.
    */
   static final class PreserveGainmapAndColorSpaceForTransformations implements Experiment {}
+
+  /** Fixes decoding of hardware gainmaps from Ultra HDR images on Android U. */
+  static final class EnableHardwareGainmapFixOnU implements Experiment {}
 
   static final class EnableImageDecoderForBitmaps implements Experiment {}
 
--- a/library/src/main/java/com/bumptech/glide/load/resource/bitmap/ImageReader.java
+++ b/library/src/main/java/com/bumptech/glide/load/resource/bitmap/ImageReader.java
@@ -3,10 +3,8 @@ package com.bumptech.glide.load.resource.bitmap;
 import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
 import android.graphics.BitmapFactory.Options;
-import android.os.Build;
 import android.os.ParcelFileDescriptor;
 import androidx.annotation.Nullable;
-import androidx.annotation.RequiresApi;
 import com.bumptech.glide.load.ImageHeaderParser;
 import com.bumptech.glide.load.ImageHeaderParser.ImageType;
 import com.bumptech.glide.load.ImageHeaderParserUtils;
@@ -17,6 +15,7 @@ import com.bumptech.glide.load.engine.bitmap_recycle.ArrayPool;
 import com.bumptech.glide.util.ByteBufferUtil;
 import com.bumptech.glide.util.Preconditions;
 import java.io.File;
+import java.io.FileDescriptor;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
@@ -29,6 +28,7 @@ import java.util.List;
  * type wrapped into a {@link DataRewinder}.
  */
 interface ImageReader {
+
   @Nullable
   Bitmap decodeBitmap(BitmapFactory.Options options) throws IOException;
 
@@ -43,17 +43,25 @@ interface ImageReader {
     private final byte[] bytes;
     private final List<ImageHeaderParser> parsers;
     private final ArrayPool byteArrayPool;
+    private final boolean enableHardwareGainmapFixOnU;
 
-    ByteArrayReader(byte[] bytes, List<ImageHeaderParser> parsers, ArrayPool byteArrayPool) {
+    ByteArrayReader(
+        byte[] bytes,
+        List<ImageHeaderParser> parsers,
+        ArrayPool byteArrayPool,
+        boolean enableHardwareGainmapFixOnU) {
       this.bytes = bytes;
       this.parsers = parsers;
       this.byteArrayPool = byteArrayPool;
+      this.enableHardwareGainmapFixOnU = enableHardwareGainmapFixOnU;
     }
 
     @Nullable
     @Override
     public Bitmap decodeBitmap(Options options) {
-      return BitmapFactory.decodeByteArray(bytes, /* offset= */ 0, bytes.length, options);
+      return enableHardwareGainmapFixOnU
+          ? GlideBitmapFactory.decodeByteArray(bytes, options)
+          : BitmapFactory.decodeByteArray(bytes, /* offset= */ 0, bytes.length, options);
     }
 
     @Override
@@ -68,6 +76,7 @@ interface ImageReader {
 
     @Override
     public void stopGrowingBuffers() {}
+
   }
 
   final class FileReader implements ImageReader {
@@ -75,11 +84,17 @@ interface ImageReader {
     private final File file;
     private final List<ImageHeaderParser> parsers;
     private final ArrayPool byteArrayPool;
+    private final boolean enableHardwareGainmapFixOnU;
 
-    FileReader(File file, List<ImageHeaderParser> parsers, ArrayPool byteArrayPool) {
+    FileReader(
+        File file,
+        List<ImageHeaderParser> parsers,
+        ArrayPool byteArrayPool,
+        boolean enableHardwareGainmapFixOnU) {
       this.file = file;
       this.parsers = parsers;
       this.byteArrayPool = byteArrayPool;
+      this.enableHardwareGainmapFixOnU = enableHardwareGainmapFixOnU;
     }
 
     @Nullable
@@ -88,7 +103,9 @@ interface ImageReader {
       InputStream is = null;
       try {
         is = new RecyclableBufferedInputStream(new FileInputStream(file), byteArrayPool);
-        return BitmapFactory.decodeStream(is, /* outPadding= */ null, options);
+        return enableHardwareGainmapFixOnU
+            ? GlideBitmapFactory.decodeStream(is, options)
+            : BitmapFactory.decodeStream(is, /* outPadding= */ null, options);
       } finally {
         if (is != null) {
           try {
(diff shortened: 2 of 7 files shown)


falhassen:
I can try adding unit tests, but I wanted to check on overall feedback with this approach before adding tests I might have to erase depending on reviewer feedback.

falhassen:
Similarly, I can also add a flag guard, but I wanted your take on this approach first.

sjudd:
What if you turn off hardware Bitmaps for U until the fix goes out?

falhassen:
I'm worried about the memory implications of doing that. If I understand
correctly, we would double our bitmap memory allocation.

falhassen:
Also, we will have to live with some devices being permanently in this
state and I would prefer a workaround. As devidd upgrade to V, this will be
a non-issue.

falhassen:
Finally, I have a client workaround (using BitmapRegionDecoder, which
doesn't use hardware bitmaps), but that fix has some UI drawbacks in our
particular application.

falhassen:
Final finally, I don't think we can expect all affected apps to turn off
hardware bitmaps. It would be nice if Glide could just have the fix
built-in.

library/src/main/java/com/bumptech/glide/load/resource/bitmap/ImageReader.java
```diff
@@ -68,6 +80,44 @@ public int getImageOrientation() throws IOException {
...
+    @RequiresApi(VERSION_CODES.UPSIDE_DOWN_CAKE)
+    @Nullable
+    private static Bitmap safeDecodeHardwareBitmapWithGainmap(byte[] bytes, Options options) {
+      Bitmap softwareBitmap = null;
+      try {
+        options.inPreferredConfig = Bitmap.Config.ARGB_8888;
+        softwareBitmap =
+            BitmapFactory.decodeByteArray(bytes, /* offset= */ 0, bytes.length, options);
+        if (softwareBitmap == null) {
+          return null;
```
jreck, line 103: This block of code is repeated a bunch, I think you could do something like this instead:

```
    private static Bitmap safeDecodeHardwareBitmapWithGainmap(byte[] bytes, Options options) {
        try {
            options.inPreferredConfig = Bitmap.Config.ARGB_8888;
            Bitmap softwareBitmap =
                    BitmapFactory.decodeByteArray(bytes, /* offset= */ 0, bytes.length, options);
            return safeDecodeBitmapWithGainmapCommon(softwareBitmap);
        } finally {
            options.inPreferredConfig = Config.HARDWARE;
        }
    }

    private static Bitmap safeDecodeBitmapWithGainmapCommon(Bitmap softwareBitmap) {
        if (softwareBitmap == null) {
            return null;
        }
        try {
            Gainmap gainmap = softwareBitmap.getGainmap();
            if (gainmap != null) {
                Bitmap gainmapContents = gainmap.getGainmapContents();
                if (gainmapContents.getConfig() == Bitmap.Config.ALPHA_8) {
                    softwareBitmap.setGainmap(
                            GainmapCopier.convertSingleChannelGainmapToTripleChannelGainmap(gainmap));
                }
            }
            return softwareBitmap.copy(Bitmap.Config.HARDWARE, /* isMutable= */ false);
        } finally {
            softwareBitmap.recycle();
        }
    }
```

jreck (approved):

library/src/main/java/com/bumptech/glide/load/resource/bitmap/ImageReader.java
falhassen: Thanks for the tip. DOne.

jreck (approved):

falhassen:
I added emulator tests to DownsamplerEmulatorTest. Unfortunately, my emulated doesn't seem to reproduce this framework issue, but at least they can guard against regressions.

library/src/main/java/com/bumptech/glide/load/resource/bitmap/ImageReader.java
```diff
@@ -245,8 +341,199 @@ public int getImageOrientation() throws IOException {
...
+                parcelFileDescriptor.getFileDescriptor(), /* outPadding= */ null, options);
+        if (softwareBitmap == null) {
+          return null;
+        }
+        return AndroidUHardwareBitmapFactoryDecoder.safeDecodeBitmapWithGainmap(softwareBitmap);
+      } finally {
+        if (softwareBitmap != null) {
+          softwareBitmap.recycle();
+        }
+        options.inPreferredConfig = Config.HARDWARE;
```
kanelbulle, line 374: nit: It took me a second to convince myself that you're just resetting it to what it was before, but I think it would be easier to reason about this method if you add an assertion at the start of it that the options are indeed set to `HARDWARE`.

library/src/main/java/com/bumptech/glide/load/resource/bitmap/ImageReader.java
```diff
@@ -245,8 +341,199 @@ public int getImageOrientation() throws IOException {
...
+     * Returns a decoded bitmap for the input software bitmap, ensuring that any associated gainmap
+     * is decoded without errors on Android U if it is a valid gainamp.
+     *
+     * <p>Callers should use the method to decode bitmaps with {@link
+     * BitmapFactory#decodeStream(InputStream, Rect, Options)} on Android U.
+     *
+     * @param softwareBitmap The bitmap to be decoded. Must not be a hardware bitmap..
+     */
+    public static Bitmap safeDecodeBitmapWithGainmap(Bitmap softwareBitmap) {
+      if (softwareBitmap == null) {
```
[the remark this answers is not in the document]
falhassen: You're right. Removed.

library/src/main/java/com/bumptech/glide/load/resource/bitmap/ImageReader.java
falhassen: Done.

library/src/main/java/com/bumptech/glide/load/resource/bitmap/ImageReader.java
```diff
@@ -245,8 +341,199 @@ public int getImageOrientation() throws IOException {
...
+     */
+    public static boolean needsGainmapDecodeWorkaround(Options options) {
+      if (VERSION.SDK_INT != VERSION_CODES.UPSIDE_DOWN_CAKE) {
+        return false;
+      }
+      if (options.inPreferredConfig != Config.HARDWARE) {
+        return false;
+      }
+      Boolean requiresGainmapDecodeWorkaround = REQUIRES_GAIN_MAP_FIX.get();
+      if (requiresGainmapDecodeWorkaround == null) {
```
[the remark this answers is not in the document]
falhassen: Good catch. This is a bug. I fixed it (and added a @GuardedBy)

library/src/main/java/com/bumptech/glide/load/resource/bitmap/ImageReader.java
```diff
@@ -43,16 +51,27 @@ final class ByteArrayReader implements ImageReader {
...
       this.bytes = bytes;
       this.parsers = parsers;
       this.byteArrayPool = byteArrayPool;
+      this.enableHardwareGainmapFixOnU = enableHardwareGainmapFixOnU;
     }
 
     @Nullable
     @Override
     public Bitmap decodeBitmap(Options options) {
+      if (enableHardwareGainmapFixOnU
```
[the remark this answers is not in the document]
falhassen: Done.

library/src/main/java/com/bumptech/glide/GlideBuilder.java
```diff
@@ -499,6 +499,17 @@ public GlideBuilder setPreserveGainmapAndColorSpaceForTransformations(boolean is
     return this;
   }
 
+  /**
+   * Fixes decoding of hardware gainmaps from Ultra HDR images on Android U.
+   *
+   * <p>Without this flag on, gainmaps may be dropped when decoding Ultra HDR on Android U devices
+   * using skiagl for hwui..
```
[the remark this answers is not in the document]
falhassen: Done

falhassen:
> I added emulator tests to DownsamplerEmulatorTest. Unfortunately, my emulated device doesn't seem to reproduce this framework issue, but at least they can guard against regressions. Certainly, the emulated device can reproduce the non-default exif orientation bug when I disable the fix.

Actually, I found a bug in my commit where I configured the gainmap to use `ARGB_8888`. To reproduce the bug, I need to configure the gainmap to use `ALPHA_8`. Once I do that, the emulator test will fail without the workaround.

library/src/main/java/com/bumptech/glide/load/resource/bitmap/GlideBitmapFactoryWrapper.java
```diff
@@ -0,0 +1,291 @@
...
+ * <ul>
+ *   <li>Ultra HDR image single-channel gainmaps not being decoded on Android U when hardware
+ *       bitmaps are enabled. This issue is further described in
+ *       https://github.com/bumptech/glide/issues/5362.
+ * </ul>
+ *
+ * <p>New usages of {@link BitmapFactory} APIs within Glide should be added here rather than called
+ * directly.
+ */
+final class GlideBitmapFactoryWrapper {
```
kanelbulle, line 35: nit: drop the Wrapper? it's cleaner

library/src/main/java/com/bumptech/glide/load/resource/bitmap/GlideBitmapFactoryWrapper.java
```diff
@@ -0,0 +1,291 @@
...
+   *
+   * @param inputStream for the bitmap to be decoded.
+   * @param options to be applied in the {@link BitmapFactory#decodeStream} call.
+   */
+  @RequiresApi(VERSION_CODES.UPSIDE_DOWN_CAKE)
+  @Nullable
+  private static Bitmap safeDecodeHardwareBitmapWithGainmap(
+      InputStream inputStream, Options options) {
+    Preconditions.checkArgument(options.inPreferredConfig == Config.HARDWARE);
+    options.inPreferredConfig = Bitmap.Config.ARGB_8888;
```
kanelbulle, line 85: you're setting options twice (once here and once inside the try). It should be the last statement before the `try {`

Applies to this and all the methods below

library/src/main/java/com/bumptech/glide/load/resource/bitmap/GlideBitmapFactoryWrapper.java
```diff
@@ -0,0 +1,291 @@
...
+    /**
+     * Returns true if a gainmap decoding workaround is required to mitigate an Android U bug. This
+     * method tests for the presence of the bug, which only affects hardware bitmaps, and caches the
+     * result in memory.
+     *
+     * <p>This method is thread-safe.
+     *
+     * @param options which will be used to decode the gainmap.
+     */
+    static synchronized boolean needsGainmapDecodeWorkaround(Options options) {
```
kanelbulle, line 265: I think you can be a bit more efficient with the locking, some of the preamble here doesn't have to be synchronized, and they don't need a shared lock until REQUIRES_GAIN_MAP_FIX is found to be null?

Not sure if it matters that much, but we do know these methods are invoked from 4 threads concurrently.

falhassen: Done. Sorry this had violated the Effective Java rule of avoid excessive synchronization. 

I switched over to using GlideSupplier, which is the Glide way of memoizing expensive compuatiohns.

library/src/main/java/com/bumptech/glide/load/resource/bitmap/GlideBitmapFactoryWrapper.java
falhassen: Whoops thanks for catching this.

library/src/main/java/com/bumptech/glide/load/resource/bitmap/GlideBitmapFactoryWrapper.java
falhassen: Done

falhassen:
Sorry, I think there is a bug in my last push. Will update once I fix...

falhassen:
> Sorry, I think there is a bug in my last push. Will update once I fix...

Looks like it's working as intended (a gainmap is decoded), but I just discovered a different bug in my code

library/src/main/java/com/bumptech/glide/load/resource/bitmap/ImageReader.java
```diff
@@ -43,17 +43,25 @@ final class ByteArrayReader implements ImageReader {
...
       this.parsers = parsers;
       this.byteArrayPool = byteArrayPool;
+      this.enableHardwareGainmapFixOnU = enableHardwareGainmapFixOnU;
     }
 
     @Nullable
     @Override
     public Bitmap decodeBitmap(Options options) {
-      return BitmapFactory.decodeByteArray(bytes, /* offset= */ 0, bytes.length, options);
+      return enableHardwareGainmapFixOnU
```
kanelbulle, line 62: The goal here and below is that once enableHardwareGainmapFixOnU is permanently shipped, we remove enableHardwareGainmapFixOnU and only call into GlideBitmapFactory, right?

falhassen: Yes, exactly. We will clean it up as we would any Glide flag.

falhassen:
I fixed the issue. The bitmap.copy operation was setting all alpha values to zero. I copied the gainmap with a filter forcing the 100% opacity (which should be working-as-intended, since ALPHA_8 images only have a single channel).

Issue #5333, opened 2023-11-16
Glide transformation drop gainmap and colorspace


falhassen:
This is because LruBitmapPool can recycle Bitmaps with no gainmaps and a different colorspace.

Since gainmaps can't be reused between bitmaps, and must be set manually, either we have to do a BitmapPool API change to handle that, or all callers of BitmapPool have to handle it explicitly, or we start migrating from BitmapPool, since it is arguably not needed in Android M+ since the Android framework takes care of Bitmap reuse anyway.

falhassen:
Note that this fix is opt-in now via a Glide flag, [GlideBuilder#setPreserveGainmapAndColorSpaceForTransformations](https://github.com/bumptech/glide/blob/e77562a432352da19741d93b8c1f6599b558676a/library/src/main/java/com/bumptech/glide/GlideBuilder.java#L497).

Once we have confidence in the fix, we can clean-up the flag and default the behavior to true.

PR #5366, opened 2024-02-03
Clean up preserveGainmapAndColorSpaceForTransformations flag

falhassen:
This is almost a pure clean-up, with only one notable difference.

The new code path changed the behavior of rotateImageExif, which in the old path forced a non-null config of ARGB_8888 for bitmaps with null configs. In order to reuse bitmaps using `Bitmap.reconfigure`, a non-null config is required (see https://github.com/bumptech/glide/issues/194). A null config can be due to a non-public bitmap type as documented in that issue.

The new flow removes the need to reconfigure a bitmap, and thus should do a better job of preserving the config and not forcing it to ARGB_8888.

Thus, I had to change one unit test that tested this behavior being exercised in BitmapPool.


falhassen:
Clean up for https://github.com/bumptech/glide/issues/5333

kanelbulle (approved):

PR #5437, opened 2024-08-14
Update incorrect import in migration guide

NinoDLC:
## Description
The import for the new API was incorrect there : https://bumptech.github.io/glide/doc/generatedapi.html


kanelbulle:
Thanks!

Issue #5404, opened 2024-05-02
error(R.drawable.xxxx) and placeholder(R.drawable.xxxx) do not work in glide


dunkindonass:
**Glide Version**: 4.11.0

**Integration libraries**:

**Device/Android Version**: any devices

**Issue details / Repro steps / Use case background**: 
I am using Glide in dfm and save the image in the drawble folder so that an error image is displayed when the image load fails as shown below.
```
 Glide.with(binding.root.context)
  .load("url")
  .error(R.drawable.xxxxx)
  .into(binding.icon)
```

If the resource id value of R.drawable.xxxxx is a negative integer, it will not operate. If it is a positive integer, it will operate normally.

```
 public T error(@DrawableRes int resourceId)
```
Are you separately checking the validity of the errorId value saved in the above function?

eyeahs:
In the Glide library version 4.14.2, there is an issue with the method getErrorDrawable() located in the com.bumptech.glide.request.SingleRequest<R> class, specifically at line 386. The method is defined as follows:
```
@GuardedBy("requestLock")
private Drawable getErrorDrawable() {
  if (errorDrawable == null) {
    errorDrawable = requestOptions.getErrorPlaceholder();
    if (errorDrawable == null && requestOptions.getErrorId() > 0) {
      errorDrawable = loadDrawable(requestOptions.getErrorId());
    }
  }
  return errorDrawable;
}
```

The conditional check requestOptions.getErrorId() > 0 is problematic because Android resource IDs can be negative values. Thus, this code may lead to incorrect behavior by not properly handling cases where the error ID is negative. This oversight could prevent the loading of valid negative resource IDs as error drawables.

injae-kim:
Fix PR: https://github.com/bumptech/glide/pull/5447!

PR #5447, opened 2024-09-09
Fix ResourceID validation check expression

hyuns66:
## Description

related issue : #5404
 I changed three methods which have wrong expression
- getFallbackDrawable()
- getErrorDrawable()
- getPlaceholderDrawable()
Resource IDs can be negative, but the functions were treating such IDs as invalid. Therefore, I improved the if statement logic to allow valid resource IDs that have negative values to pass through.

## Motivation and Context
According to the [referenced documentation](https://android.googlesource.com/platform/frameworks/base/+/HEAD/core/java/android/content/res/ResourceId.java), resource IDs can be negative, and while some legacy code assumes -1 to be an invalid resource ID, this is incorrect. The issue #5404 involves a method malfunctioning when using an image with a specific negative value as its resource ID, which was identified as a validation issue related to resourceID. The original code incorrectly assumed any negative value to be an invalid resource ID. However, I have refactored it using the correct validation logic as outlined in the android.content.res package.

### DIFF
--- a/library/src/main/java/com/bumptech/glide/request/SingleRequest.java
+++ b/library/src/main/java/com/bumptech/glide/request/SingleRequest.java
@@ -387,7 +387,7 @@ public final class SingleRequest<R> implements Request, SizeReadyCallback, Resou
   private Drawable getErrorDrawable() {
     if (errorDrawable == null) {
       errorDrawable = requestOptions.getErrorPlaceholder();
-      if (errorDrawable == null && requestOptions.getErrorId() > 0) {
+      if (errorDrawable == null && requestOptions.hasErrorId()) {
         errorDrawable = loadDrawable(requestOptions.getErrorId());
       }
     }
@@ -398,7 +398,7 @@ public final class SingleRequest<R> implements Request, SizeReadyCallback, Resou
   private Drawable getPlaceholderDrawable() {
     if (placeholderDrawable == null) {
       placeholderDrawable = requestOptions.getPlaceholderDrawable();
-      if (placeholderDrawable == null && requestOptions.getPlaceholderId() > 0) {
+      if (placeholderDrawable == null && requestOptions.hasPlaceholderId()) {
         placeholderDrawable = loadDrawable(requestOptions.getPlaceholderId());
       }
     }
@@ -409,7 +409,7 @@ public final class SingleRequest<R> implements Request, SizeReadyCallback, Resou
   private Drawable getFallbackDrawable() {
     if (fallbackDrawable == null) {
       fallbackDrawable = requestOptions.getFallbackDrawable();
-      if (fallbackDrawable == null && requestOptions.getFallbackId() > 0) {
+      if (fallbackDrawable == null && requestOptions.hasFallbackId()) {
         fallbackDrawable = loadDrawable(requestOptions.getFallbackId());
       }
     }
--- a/library/src/main/java/com/bumptech/glide/request/BaseRequestOptions.java
+++ b/library/src/main/java/com/bumptech/glide/request/BaseRequestOptions.java
@@ -1364,8 +1364,18 @@ public abstract class BaseRequestOptions<T extends BaseRequestOptions<T>> implem
   }
 
   @SuppressWarnings("WeakerAccess")
+  public final boolean hasErrorId() {
+    return isSet(ERROR_ID);
+  }
+
+  @SuppressWarnings("WeakerAccess")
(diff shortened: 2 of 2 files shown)


library/src/main/java/com/bumptech/glide/request/SingleRequest.java
```diff
@@ -387,7 +387,8 @@ public boolean isAnyResourceSet() {
   private Drawable getErrorDrawable() {
     if (errorDrawable == null) {
       errorDrawable = requestOptions.getErrorPlaceholder();
-      if (errorDrawable == null && requestOptions.getErrorId() > 0) {
+      if (errorDrawable == null && requestOptions.getErrorId() != -1 &&
+              (requestOptions.getErrorId() & 0xff000000) != 0 && (requestOptions.getErrorId() & 0x00ff0000) != 0) {
```
injae-kim, line 391:
```java
/**
 * @see android.content.res.ResourceId.isValid
 */
private static boolean isValidId(int id) {
  return id != -1 && id & 0xff000000 != 0 && id  & 0x00ff0000 != 0;
}

if (isValidId(requestOptions.getErrorId()))
if (isValidId(requestOptions.getPlaceholderId()))
if (isValidId(requestOptions.getFallbackId()))
```
Can we reduce duplicated code like above?

hyuns66: Thanks! I have made the changes accordingly. 😊

library/src/main/java/com/bumptech/glide/request/SingleRequest.java
```diff
@@ -409,13 +411,20 @@ private Drawable getPlaceholderDrawable() {
...
+      if (fallbackDrawable == null && isValidId(requestOptions.getFallbackId())) {
         fallbackDrawable = loadDrawable(requestOptions.getFallbackId());
       }
     }
     return fallbackDrawable;
   }
 
+  /**
+   * @see android.content.res.ResourceId.isValid
+   */
```
kanelbulle, line 423: I'd prefer not to rely on this hidden framework method to identify the validity of this ID - I'm guessing it's not exposed for a reason. This may not be consistent across SDKs and it relies on the behavior of aapt.

I think it would be better to expose `hasErrorId`, `hasPlaceholderId`, `hasFallbackId` on requestoptions that checks the underlying `fields` field instead, and make the getters return `null`.

hyuns66: I agree with your point about not relying on this hidden framework method. 

However, I’d like to ask about the intention behind the code `requestOptions.getErrorId() > 0`. 
I thought this code was intended to verify whether the resource ID is valid or not.

However, based on my understanding of your suggestion, the code should be modified as follows. 

```java
if (errorDrawable == null && requestOptions.getErrorId() != null) {
        errorDrawable = loadDrawable(requestOptions.getErrorId());
      }
```

And it seems that it is not verifying the validity of the resourceID, but simply checking for its existence.
Could you clarify the purpose of this code `requestOptions.getErrorId() > 0`?

hyuns66: I searched for a definitive method to validate the resourceId, but I couldn't find a clear and publicly documented way to do so. Therefore, I believe that the responsibility for validating resourceId should lie with the users themselves, as shown in the code below, rather than relying solely on the Glide library.

```kotlin
try {
    Drawable drawable = getResources().getDrawable(R.drawable.my_image);
    // Valid resource
} catch (Resources.NotFoundException e) {
    // Invalid resource
}
```

Therefore, I refactored my commit to check the existance of `errorId`, `placeholderId`, `fallbackId` as you suggested.
Please, verify it @kanelbulle

kanelbulle: Thanks! I looked back in the history and this check was added in https://github.com/bumptech/glide/commit/33b6412185eb0d6860c2eebbc330d5c31c4429da#diff-ee7b45dc91d122ca9ab77bb479e9daaebdf68d3705a77a09a0a5e71cba590528 combined with a number of other changes. I suspect it was an easy way to check if an id was set or not, which I think your change now better accomplishes.

hyuns66: @kanelbulle
It seems that auto-merge has been disabled after I rebased my commits. 
Could you kindly take a look when you have a moment?

PR #5471, opened 2025-01-16
Reverts Clean up preserveGainmapAndColorSpaceForTransformations flag #5386

falhassen:
## Description
Reverts Clean up setEnableHardwareGainmapFixOnU flag #5386

## Motivation and Context
The flow executed when setEnableHardwareGainmapFixOnU is set to true excessively uses software bitmaps, and without pooling, significantly increasing memory usage in apps.

A future PR will attempt fix the memory issues, but at least the setEnableHardwareGainmapFixOnU flow will now become opt-in.


falhassen:
Might as well fix-forward

Issue #5474, opened 2025-02-03
Android U Gainmap decoding workaround uses too much memory


falhassen:
We should not be allocating software bitmaps unless absolutely necessary. We should probably pool them as well.

PR #5475, opened 2025-02-03
Only do the Android U gainmap workaround on bitmaps that have a gainmap

falhassen:
## Description
Before attempting to software decode bitmaps on Android U devices that are susceptible to the Android framework gainmap bug, we parse the image to detect the presence of multi-picture format data. If we detect such data, it is very likely that there is a gainmap there, and thus we proceed with the software bitmap workaround. This prevents having to add any extra decoding for images without gainmaps.

## Motivation and Context
We should only decode software bitmaps as a last resort. See https://github.com/bumptech/glide/issues/5474.


library/src/main/java/com/bumptech/glide/load/resource/bitmap/GlideBitmapFactory.java
```diff
@@ -85,8 +92,40 @@ public static Bitmap decodeFileDescriptor(
...
+    // Check to make sure that the inputStream has an associated bitmap using BitmapRegionDecoder,
+    // which does not require the workaround.
+    Bitmap onePixelBitmap = null;
+    BitmapRegionDecoder bitmapRegionDecoder = null;
+    try {
+      bitmapRegionDecoder = BitmapRegionDecoder.newInstance(inputStream);
+      if (bitmapRegionDecoder == null) {
+        return null;
+      }
+      Rect onePercentRect =
```
kanelbulle, line 104: Why one percent instead of a fixed 1x1 size?

library/src/main/java/com/bumptech/glide/load/resource/bitmap/GlideBitmapFactory.java
```diff
@@ -72,7 +76,10 @@ public static Bitmap decodeFileDescriptor(
 
   /**
    * Returns a decoded bitmap for the input stream, ensuring that any associated gainmap is decoded
-   * without errors on Android U if it is a valid gainamp.
+   * without errors on Android U if it is a valid gainmap.
```
kanelbulle, line 79: Could you document what happens if there are errors with the gainmap - what does this method return then?

library/src/main/java/com/bumptech/glide/load/resource/bitmap/GlideBitmapFactory.java
```diff
@@ -85,8 +92,40 @@ public static Bitmap decodeFileDescriptor(
...
+    Bitmap onePixelBitmap = null;
+    BitmapRegionDecoder bitmapRegionDecoder = null;
+    try {
+      bitmapRegionDecoder = BitmapRegionDecoder.newInstance(inputStream);
+      if (bitmapRegionDecoder == null) {
+        return null;
+      }
+      Rect onePercentRect =
+          getOnePercentRect(bitmapRegionDecoder.getWidth(), bitmapRegionDecoder.getHeight());
+      onePixelBitmap = bitmapRegionDecoder.decodeRegion(onePercentRect, options);
```
kanelbulle, line 106: This is "onePercentBitmap" right now, but maybe this will be onePixelBitmap again in response to my above comment..

kanelbulle (commented):
Are there any tests that would be useful here?

library/src/main/java/com/bumptech/glide/load/resource/bitmap/GlideBitmapFactory.java
falhassen: Because a fixed 1x1 size seems to drop the gainmap. I'm not sure if the issue is in Android framework or in the SKIA codec that it uses to decode the image.

falhassen: I added a comment clarifying the need to do this. I would have really preferred the 1x1 pixel solution. If we find a way to make 1x1 work, I can send a separate PR to go back to 1x1 along with the workaround (how many layers of workarounds will we have?).

library/src/main/java/com/bumptech/glide/load/resource/bitmap/GlideBitmapFactory.java
```diff
@@ -85,8 +92,40 @@ public static Bitmap decodeFileDescriptor(
   private static Bitmap safeDecodeHardwareBitmapWithGainmap(
       InputStream inputStream, Options options) {
     Preconditions.checkArgument(options.inPreferredConfig == Config.HARDWARE);
-    Bitmap softwareBitmap = null;
+    // Check to make sure that the inputStream has an associated bitmap using BitmapRegionDecoder,
+    // which does not require the workaround.
+    Bitmap onePixelBitmap = null;
```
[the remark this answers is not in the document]
falhassen: Thanks, fixed.

library/src/main/java/com/bumptech/glide/load/resource/bitmap/GlideBitmapFactory.java
falhassen: Ack

falhassen:
> Are there any tests that would be useful here?

This behavior is already being tested, with the test added.
https://github.com/bumptech/glide/pull/5357/files#diff-2f35cfb70148884b2a151428db8b770eb84c23765ce119578b945a59c26cc832

This change simply is more memory efficient. There should be no API changes. I could add more test files that I know fail the 1x1 pixel test but pass the 1% test, but unfortunately this bug only manifests in hardware and we would need Glide instrumentation tests actually running on real Samsung S24+ devices that have this bug, which is beyond the scope of what I can do (instrumentation tests are not running this Glide repo, I don't know if S24 devices are available anyway).

library/src/main/java/com/bumptech/glide/load/resource/bitmap/GlideBitmapFactory.java
kanelbulle: Is there a minimum pixel size that is guaranteed to work regardless of image size? Any guidance from the framework team?

falhassen: Since this 1x1 pixel decoding (or any reasonable-sounding minimum pixel decoding) is inherently broken in Android U and on earlier versions of Android V, I decided to go with an approach where we bypass decoding altogether and just parse images to see if they contain MPF, a strong sign that there is a gainmap.

I could have also checked definitively for gainmaps, but given that Ultra HDR images only have a clear signal in XMP that there is a gainmap (see https://developer.android.com/media/platform/hdr-image-format), while only ISO 21496-1 gainmaps (used by iOS) have a clear signal in MPF, but both use MPF, it makes sense just to check for the presence of MPF.

library/src/main/java/com/bumptech/glide/load/ImageHeaderParserUtils.java
```diff
@@ -235,11 +235,130 @@ private static int getOrientationInternal(
     return ImageHeaderParser.UNKNOWN_ORIENTATION;
   }
 
+  /**
+   * Returns the result from the first of {@code parsers} that returns true when MPF is detected, if
+   * any..
+   *
+   * <p>If {@code buffer} is null, the parers list is empty, or none of the parsers returns a valid
```
kanelbulle, line 242: typo: parers

library/src/main/java/com/bumptech/glide/load/resource/bitmap/DefaultImageHeaderParser.java
```diff
@@ -301,6 +351,37 @@ private boolean hasJpegExifPreamble(byte[] exifData, int exifSegmentLength) {
...
    */
   private int moveToExifSegmentAndGetLength(Reader reader) throws IOException {
+    return moveToSegmentAndGetLength(reader, EXIF_SEGMENT_TYPE);
+  }
+
+  private boolean hasJpegMpfPreamble(Reader reader, byte[] tempArray, int mpfSegmentLength)
+      throws IOException {
+    int read = reader.read(tempArray, mpfSegmentLength);
+    if (read != mpfSegmentLength) {
+      if (Log.isLoggable(TAG, Log.DEBUG)) {
```
kanelbulle, line 361: is this really debug, isn't this expected that images do not have this data?

library/src/main/java/com/bumptech/glide/load/resource/bitmap/ImageReader.java
```diff
@@ -134,6 +141,23 @@ public int getImageOrientation() throws IOException {
       }
     }
 
+    @Override
+    public boolean hasJpegMpf() throws IOException {
+      InputStream is = null;
+      try {
+        is = new RecyclableBufferedInputStream(new FileInputStream(file), byteArrayPool);
```
kanelbulle, line 148: is there a point in creating a recyclable one here, the hasJpegMpf does this internally if needed?

kanelbulle (approved):

library/src/main/java/com/bumptech/glide/load/resource/bitmap/DefaultImageHeaderParser.java
falhassen: Sorry, I forgot to rename the param. This error is WAI, as basically what is happening is we log an error if the APP2 segment you are passing is invalid. I'll rename the param and add javadoc to clarify the contract.

library/src/main/java/com/bumptech/glide/load/resource/bitmap/ImageReader.java
falhassen: That's true - I'll remove this.

library/src/main/java/com/bumptech/glide/load/ImageHeaderParserUtils.java
falhassen: Done in subsequent PR

Issue #5356, opened 2024-01-23
Glide compose crash when using GlideImage inside a Row(Modifier.height(IntrinsicSize.Min))


nordfalk:
**Glide Version**: 5.0.0-rc01
```
    implementation "com.github.bumptech.glide:glide:5.0.0-rc01"
    implementation "com.github.bumptech.glide:okhttp3-integration:5.0.0-rc01"
    implementation platform("androidx.compose:compose-bom:2023.10.01")
    implementation "androidx.compose.ui:ui:"
    implementation "androidx.compose.material:material"
    implementation 'androidx.compose.material3:material3-android:1.2.0-beta02'
    implementation "com.github.bumptech.glide:compose:1.0.0-beta01"
```

**Device/Android Version**: I only tried in the emulator

**If I use GlideImage inside Row(Modifier.height(IntrinsicSize.Min)) it crashes**:
```
            Row(Modifier.height(IntrinsicSize.Min)) {
                GlideImage(
                    model = "https://javabog.dk/JSP-lille.jpg",
                    contentDescription = "Text",
                )
            }
```

**Stack trace / LogCat**:
```
FATAL EXCEPTION: main
             Process: dk.dinero.android.debug, PID: 30072
             java.lang.IllegalArgumentException: Cannot round NaN value.
             	at kotlin.math.MathKt__MathJVMKt.roundToInt(MathJVM.kt:1165)
             	at com.bumptech.glide.integration.compose.GlideNode.modifyConstraints-ZezNO4M(GlideModifier.kt:571)
             	at com.bumptech.glide.integration.compose.GlideNode.measure-3p2s80s(GlideModifier.kt:523)
             	at androidx.compose.ui.node.LayoutModifierNode$minIntrinsicHeight$1.measure-3p2s80s(LayoutModifierNode.kt:92)
             	at androidx.compose.ui.node.NodeMeasuringIntrinsics.minHeight$ui_release(LayoutModifierNode.kt:201)
             	at androidx.compose.ui.node.LayoutModifierNode.minIntrinsicHeight(LayoutModifierNode.kt:90)
             	at androidx.compose.ui.node.LayoutModifierNodeCoordinator.minIntrinsicHeight(LayoutModifierNodeCoordinator.kt:144)
             	at androidx.compose.ui.node.LayoutNodeLayoutDelegate$MeasurePassDelegate.minIntrinsicHeight(LayoutNodeLayoutDelegate.kt:785)
[...]

Issue #5256, opened 2023-08-23
Crash in glide compose alpha.5, when computing scale factor


joerogers:
**Glide Version**: 1.0.0-alpha.5/4.16.0

**Integration libraries**: okhttp4-integration (although issue is not here)

**Device/Android Version**: Emulator/Pixel 6a

**Issue details / Repro steps / Use case background**: 

Looks like glide triggers a `ScaleFactor is unspecified` exception, if either the width or height is implicit and the other is computed from a ratio if the width or height resolves to 0.dp.

I found this, because I had gotten 1.0.0-alpha.3 working with a compose MotionLayout where the image is "Gone" in the initial constraints and appears in end constraints. It took time to determine the root cause to see if it was the MotionLayout code or Glide, but I was able to reproduce using a standard GlideImage without MotionLayout involved.

Here is the abbreviated stack:

```
FATAL EXCEPTION: main
                              Process: com.example.app, PID: 5135
                              java.lang.IllegalStateException: ScaleFactor is unspecified
                              at androidx.compose.ui.layout.ScaleFactor.getScaleX-impl(ScaleFactor.kt:48)
                              at androidx.compose.ui.layout.ScaleFactorKt.times-UQTWf7w(ScaleFactor.kt:154)
                              at com.bumptech.glide.integration.compose.GlideNode.modifyConstraints-ZezNO4M(GlideModifier.kt:481)
                              at com.bumptech.glide.integration.compose.GlideNode.measure-3p2s80s(GlideModifier.kt:440)
                              at androidx.compose.ui.node.LayoutModifierNodeCoordinator.measure-BRTryo0(LayoutModifierNodeCoordinator.kt:116)
                              at androidx.compose.foundation.layout.AspectRatioNode.measure-3p2s80s(AspectRatio.kt:119)
```

It seems like the bug is the intrinsicWidth/intrinsicHeight is 0.dp and/or the constrainedWidth/Height is 0.dp.  Essentially this should be checked for 0.dp as division by 0 results in NaN and ends up triggering the exception.  I suspect if 0.dp is determined as the width or height there is no need to issue the network download or draw the image as it can not be seen.  The easy fix is to not crash, but I couldn't quite follow to determine if there is an optimization to avoid the network/cache fetch  as well.

The code in question is:

```
   val constrainedWidth = constraints.constrainWidth(intrinsicWidth)
   val constrainedHeight = constraints.constrainHeight(intrinsicHeight)

[...]

PR #5482, opened 2025-02-20
Fix NaN exception when rounding dimensions

dmytroKarataiev:
## Description

Fixes this issue: https://github.com/bumptech/glide/issues/5356

This change:
1. Checks if the scaled width/height values are finite (not NaN or infinite) using isFinite()
2. If the value is finite, rounds it to an integer as before
3. If the value is not finite (NaN or infinite), falls back to using the constraints' minimum dimensions
This prevents any potential IllegalArgumentException that could be thrown when trying to round NaN or infinite values, making the code more robust.


falhassen:
Can you add tests?

donaldchai:
I didn't notice this (or the several bugs filed starting with https://github.com/bumptech/glide/issues/5256), but I think https://github.com/bumptech/glide/pull/5499 subsumes this?

PR #5499, opened 2025-04-11
Avoid NaN when Constraints are zero (but non-fixed size).

donaldchai:
If e.g. constraints.width=0 the logic ends up computing:

```
 intrinsicWidth = constraints.width (i.e. 0)
 constrainedWidth = 0
 srcSize.width = 0
 scaleFactor.width = computeFillWidth(0,0) --> NaN
```

### DIFF
--- a/integration/compose/src/main/java/com/bumptech/glide/integration/compose/GlideModifier.kt
+++ b/integration/compose/src/main/java/com/bumptech/glide/integration/compose/GlideModifier.kt
@@ -540,11 +540,8 @@ internal class GlideNode : DrawModifierNode, LayoutModifierNode, SemanticsModifi
   private fun Constraints.hasFixedSize() = hasFixedWidth && hasFixedHeight
 
   private fun modifyConstraints(constraints: Constraints): Constraints {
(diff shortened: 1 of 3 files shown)


integration/compose/src/main/java/com/bumptech/glide/integration/compose/GlideModifier.kt
```diff
@@ -541,10 +541,10 @@ internal class GlideNode : DrawModifierNode, LayoutModifierNode, SemanticsModifi
 
   private fun modifyConstraints(constraints: Constraints): Constraints {
     if (constraints.hasFixedSize()) {
```
falhassen, line 543: Why not

`if (constraints.hasFixedSize() || constraints.isZero) {`

donaldchai: Because they're two separate reasons?

Done.

falhassen: Sorry I should have just accepted it as-is.

PR #5504, opened 2025-04-28
Integrate with the newly introduced MediaStore.openAssetFileDescriptor API

Omaaarz:
## Description

Update FileDescriptorLocalUriFetcher, StreamLocalUriFetcher and AssetFileDescriptorLocalUriFetcher to make use of the newly created MediaStore.openAssetFileDescriptor when opening Media content URI.

## Motivation and Context

New API (MediaStore.openAssetFileDescriptor) has been added in SDK extension 16 that helps with keeping MediaProvider stable when many cloud picker items are being opened at once.


falhassen:
Can you flag-guard this change?

khma93 (commented):
k

PR #5523, opened 2025-05-28
Fix: Prevent crash in LruBitmapPool.put() on recycled bitmap

yogeshkumawat:
### What this fixes

Prevents crash in `LruBitmapPool.put()` when a recycled bitmap is passed to the pool.

### Why it matters

In rare cases, Glide attempts to pool a bitmap that has already been recycled — often due to race conditions during decode, transformation, or request cancellation. Since `isRecycled()` throws `IllegalStateException`, this leads to a fatal crash.

### Fix

Add a defensive check and early return if the bitmap is already recycled. This mirrors similar fixes in other parts of Glide (e.g., `GifDrawable.draw()` crash protection).

### Impact

- ✅ Prevents crash without affecting cache logic
- ✅ Improves robustness
- ✅ Compatible with current logic (only avoids pooling invalid bitmaps)

### DIFF
--- a/library/src/main/java/com/bumptech/glide/load/engine/bitmap_recycle/LruBitmapPool.java
+++ b/library/src/main/java/com/bumptech/glide/load/engine/bitmap_recycle/LruBitmapPool.java
@@ -101,11 +101,11 @@ public class LruBitmapPool implements BitmapPool {
 
   @Override
   public synchronized void put(Bitmap bitmap) {
-    if (bitmap == null) {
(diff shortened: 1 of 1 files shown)


terancek72-dot (approved):

falhassen:
The pull request description does not discuss the behavior to change to make `bitmapp == null` no longer throw an NPE. Are you sure weant to change this?

falhassen:
Also, this change needs tests.

PR #5626, opened 2025-10-22
Limit the size of LruBitmapPool/MemoryCache when app is in background

ericmiao:
When an app is moved to background, onTrimMemory() will be called, and LruBitmapPool/MemoryCache will be trimmed. However, the app could keep running and bitmaps will be retired to MemoryCache and LruBitmapPool. In many cases, the amount of memory due to the retired bitmaps in cache is non-trivial.

This change allows an app to limit of the sizes of LruBitmapPool/MemoryCache (by MemoryCategory) when the app is in background, and restore the sizes to normal when it's moved out of background.


ericmiao:
+@sjudd , note a previous pull request was here https://github.com/bumptech/glide/pull/5546, and I had some time to address your feedback)

PR #5654, opened 2026-01-09
Fix NPE in DrawableTransformation when custom transform returns null

### LINKED ISSUE #5651: Crash on DrawableTransformation if the custom transform returns null
When implementing a custom `DrawableTransformation`, if for any reason, the custom `transform()` method returns `null`, then a check [here](https://github.com/bumptech/glide/blob/5e158121a6cac01503f1e0f6c312aa241f045552/library/src/main/java/com/bumptech/glide/load/resource/bitmap/DrawableTransformation.java#L63) fails with a null object exception because it tests if `transformed.equals(original)`.

On the other hand, on `BitmapTransformation`, the check is the opposite [here](https://github.com/bumptech/glide/blob/5e158121a6cac01503f1e0f6c312aa241f045552/library/src/main/java/com/bumptech/glide/load/resource/bitmap/BitmapTransformation.java#L86) : it tests if `original.equals(transformed)` which lets the code handle null correctly.

As a result, if the transformation can't be done, this forces to throw an Exception that won't be caught. Inverting the test as done on the BitmapTransformation seems like an easy fix (and even though the custom transform() should probably avoid returning null, this would at least provide better consistency).

jayyaj12:
## Description
This PR inverts the equality check in `DrawableTransformation`'s `transform` method.
Instead of calling `.equals()` on the potentially null `transformedBitmapResource`, it now calls it on `bitmapResourceToTransform`, which is guaranteed to be non-null by the preceding null check.

## Motivation and Context
If a custom `Transformation` implementation returns `null`, the current implementation throws a `NullPointerException` when attempting to check for equality.
By swapping the order of comparison to `original.equals(transformed)`, we prevent this crash. This change also aligns the behavior with `BitmapTransformation`.

Fixes #5651

### DIFF
--- a/library/src/main/java/com/bumptech/glide/Glide.java
+++ b/library/src/main/java/com/bumptech/glide/Glide.java
@@ -1,10 +1,12 @@
 package com.bumptech.glide;
 
 import android.app.Activity;
+import android.app.Application;
 import android.content.ComponentCallbacks2;
 import android.content.Context;
 import android.content.res.Configuration;
 import android.graphics.Bitmap;
+import android.os.Bundle;
 import android.os.MessageQueue.IdleHandler;
 import android.util.Log;
 import android.view.View;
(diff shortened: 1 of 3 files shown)


jayyaj12:
Branch updated and all checks passing. Ready for review when you have time. Thanks!

Issue #440, opened 2015-04-24
Potential issues around HttpUrlFetcher


TWiStErRob:
**Glide Version/Integration library (if any)**: 3.5+
**Device/Android Version**: mostly since Gingerbread
**Issue details/Repro steps**: _No effort was made to reproduce any of the following issues, it is solely based on documentation and static code analysis._
It started by noticing the `AndroidHttpClient` deprecation warning and reading the [blog post](http://android-developers.blogspot.hu/2011/09/androids-http-clients.html) it mentions. Reading the blog post and looking at `HttpUrlFetcher` sources I found the following things to review:

#### Compression and Content-Length

> In Gingerbread, we added transparent response compression. [...] HTTP’s Content-Length header returns the compressed size

I think either the compression should be disabled as stated in `HttpURLConnection` docs (Performance section) or the `Content-Length` header ignored if the response has a [`Content-Encoding` header](http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.11)

#### Caching
If a library user enables the cache by `HttpResponseCache.install` (see class javadoc) it may affect Glide's behavior. Hopefully the already used `setUseCaches(false)` mitigates this, however I didn't find any evidence in public SDK sources that it counts.

#### Parsing Content-Length
This one is just strange to my eye: there's an `int getContentLength()` method which can be used to retrieve the length or -1, exactly what `obtain` does behind the scenes. A few methods and constants could be removed, I like writing negative code :)

**Glide load line**:
```
Glide.with(context).load("http://...");
```

Issue #239, opened 2014-11-07
Make DiskCacheStrategy.SOURCE the default for remote loads


sjudd:
I'm pretty sure doing so will require a change to either ModelLoaders or DataFetchers or both, so this will have to wait until 4.0.

sjudd:
Fixed here: https://github.com/bumptech/glide/commit/0e91adf65591fd5dc58b2834c50dd453b0114818

PR #5657, opened 2026-02-04
Fix ksp2 reproducibility

Hamza417:
## Description

This PR modifies `IndexerGenerator` to enforce deterministic sorting of library modules by their qualified class names. It also adds a unit test (`ModuleSortingLogicTest`) to verify that the sorting logic remains consistent regardless of the input order provided by the compiler environment.

## Motivation and Context

The PR intends to solve the problem of determinism. Since KSP2 the module indexer is not processing the modules in deterministic order leading to unpredictable outcomes over each build, breaking the reproducibility. The problem can be circumvented temporarily by falling back to KSP1 or KAPT, but both of these processors are obsolete now, and Google recently [removed KSP1 entirely](https://github.com/google/ksp/pull/2657). The issue was discussed in [Inure #440](https://github.com/Hamza417/Inure/issues/440) and [Peristyle #239](https://github.com/Hamza417/Peristyle/issues/239) and the `GlideIndexer` broke the reproducibility each time the difference is tested.

So far using the fallback worked, but since Gradle 9.0 most of the older methods stopped working and forcing dependency updates now that triggers this non-determinism making it difficult to publish the app on F-Droid repos.

### DIFF
--- a/annotation/compiler/src/main/java/com/bumptech/glide/annotation/compiler/IndexerGenerator.java
+++ b/annotation/compiler/src/main/java/com/bumptech/glide/annotation/compiler/IndexerGenerator.java
@@ -7,6 +7,7 @@ import com.squareup.javapoet.ClassName;
 import com.squareup.javapoet.TypeSpec;
 import java.lang.annotation.Annotation;
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.List;
 import java.util.UUID;
 import javax.lang.model.element.Modifier;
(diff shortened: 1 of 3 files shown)


Hamza417:
Hey, can someone look at this PR? This one is urgently needed to be merged as it holding multiple dependency updates in the project and updating it breaks reproducibility.

falhassen:
Sorry I missed this. Let's see if validation passes and then we can merge. Thanks for the fix!

